PR #1, opened 2020-02-04
chore: add split repo templates

plamut:
This is "step 6 - Add Repo Templates" of the repo split procedure.

~~**Warning** - no `.kokoro`, `.github/*`, etc. files generated by the synthtool. There might be something missing~~

~~**Update:** BigQuery's custom `noxfile.py` did not include templated files, had to add that manually. Mind that `noxfile.py` is still autogenerated and would get overridden every time synthtool is run - we probably want to exclude it?~~

~~Alternatively, we could go back to the autogenerated noxfile. @tswast do yo maybe remember why BigQuery uses a custom noxfile in the first place?~~

**Update 2:** Managed to hack around overriding the custom noxfile in this library.


synth.py
```diff
@@ -57,4 +59,12 @@
 # Format quoted strings as plain text.
 s.replace("google/cloud/bigquery_v2/proto/*.py", "[“”]", '``')
 
+# ----------------------------------------------------------------------------
+# Add templated files
+# ----------------------------------------------------------------------------
+templated_files = common.py_library(cov_level=100)
+# we do not want override the custom noxfile with the generated one
+os.remove(os.path.join(templated_files, "noxfile.py"))
```
plamut: Looking at [CommonTemplates.py_library()](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/common.py#L56-L71), I suppose there is no better way of excluding a particular template file? Cannot manipulate the `excludes` list through `kwargs`...

synth.py
```diff
@@ -57,4 +59,12 @@
 # Format quoted strings as plain text.
 s.replace("google/cloud/bigquery_v2/proto/*.py", "[“”]", '``')
 
+# ----------------------------------------------------------------------------
+# Add templated files
+# ----------------------------------------------------------------------------
+templated_files = common.py_library(cov_level=100)
+# we do not want to override the custom noxfile with the generated one
+os.remove(os.path.join(templated_files, "noxfile.py"))
+s.move(templated_files)
```
busunkim96: You can add an excludes list to `s.move`.

```python
s.move(templated_files, excludes=["noxfile.py"])
```

plamut: Thanks!

Since it's a minor thing, I'll do it separately to not block the main PR of the repo migration.

PR #26, opened 2020-02-06
perf: remove redundant array deepcopy

### LINKED ISSUE #6: BigQuery: complex repeated schema fields are slow
When converting rows to JSON, repeated schema fields are deepcopied causing noticeable slowdowns in complex schemas.
Since conversion is performed recursively, complex schemas containing multiple levels of repeated fields (e.g. repeated list of structs each containing multiple fields of repeated structs) are copied multiple times when converting a single row - causing even further slowdown due to redundant copying. 

I've created a proposal for removing the deepcopy operation. Will reference relevant PR.

grooveygr:
Deepcopy can be a very costly operation when considering large arrays with complex nested objects.
Refactor helpers to allow recursive conversion without copying arrays.

Fixes #6 🦕


tseaver:
@plamut I'm less familiar than you with this bit of the code.  PTAL.

tests/unit/test__helpers.py
```diff
@@ -806,6 +806,35 @@ def test_w_known_field_type(self):
...
+
+    def test_w_none(self):
+        field = _make_field("INT64")
+        original = None
+        converted = self._call_fut(field, original)
+        self.assertIsNone(converted)
+
+    def test_w_record(self):
+        subfields = [
+            _make_field("INT64", name="one"),
```
tswast, line 823: The main point of the `_single_field_to_json` function is that it ignores the `REPEATED` property, right? We should have tests that validate this behavior (that is: set mode = REPEATED in some fields)

tswast (changes requested):

tswast (approved):

google/cloud/bigquery/_helpers.py
```diff
@@ -458,7 +458,11 @@ def _record_field_to_json(fields, row_value):
 
 
 def _single_field_to_json(field, row_value):
-    """Convert a single (non-repeating) field into JSON-serializable values.
+    """Convert a single field into JSON-serializable values.
+
+    Ignores mode so that this can function for ARRAY / REPEATING fiels
```
busunkim96, line 463:
```suggestion
    Ignores mode so that this can function for ARRAY / REPEATING fields
```

tswast: thank you Bu Sun

PR #32, opened 2020-02-07
feat(bigquery): add create job method

### LINKED ISSUE #14: BigQuery: add `create_job` method that takes any kind of job config (and public configuration property to job classes)

Some job failures are just due to current conditions and the job will succeed if started from the beginning. The problem is that this isn't a simple "retry" because the job configuration is mutable.

For example, BigQuery automatically populates a destination table if not set. In this case, the destination table should not be part of the retried request. In my opinion, this kind of logic is outside the scope of the client libraries, as it's not clear when destination table needs to be cleared just from the job resource.

- [ ] Provide a `create_job` method that takes any JobConfig object.
- [ ] Add a `configuration` property to job classes.
- [ ] Add an example of retrying a job that has failed due to `403 rateLimitExceeded` (possibly with resetting the destination table if it was a query job)

Add `.retry()` method to job classes. This is problematic, primarily because the configuration may have changed since the job was initially created, leading to unintended consequences or hard-to-debug failures.

 **Additional context**

See customer request for a method to retry any job at https://github.com/googleapis/google-cloud-python/issues/5555

HemangChothani:
Fixes #14

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -1370,11 +1370,12 @@ class Client(ClientWithProject):
                 source, destination_uris, job_config=extract_job_config, retry=retry
             )
         elif "query" in job_config:
-            _del_sub_prop(job_config, ["query", "destinationTable"])
+            copy_config = copy.deepcopy(job_config)
+            _del_sub_prop(copy_config, ["query", "destinationTable"])
             query_job_config = google.cloud.bigquery.job.QueryJobConfig.from_api_repr(
-                job_config
+                copy_config
             )
-            query = _get_sub_prop(job_config, ["query", "query"])
+            query = _get_sub_prop(copy_config, ["query", "query"])
             return self.query(query, job_config=query_job_config, retry=retry)
         else:
             raise TypeError("Invalid job configuration received.")
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -53,6 +53,7 @@ import google.cloud._helpers
 from google.cloud import exceptions
 from google.cloud.client import ClientWithProject
 
+from google.cloud.bigquery._helpers import _get_sub_prop
 from google.cloud.bigquery._helpers import _record_field_to_json
 from google.cloud.bigquery._helpers import _str_or_none
 from google.cloud.bigquery._helpers import _verify_job_config_type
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/client.py
```diff
@@ -1314,6 +1315,73 @@ def job_from_resource(self, resource):
...
+            ]:
+                A new job instance.
+        """
+
+        if "load" in job_config:
+            load_job_config = google.cloud.bigquery.job.LoadJobConfig.from_api_repr(
+                job_config
+            )
+            destination = TableReference.from_api_repr(
+                job_config["load"]["destinationTable"]
```
plamut: Just a thought - what if the input dict is incorrectly structured? The user might then face rather uninformative `KeyError`s.

It seems that it would make sense to use `_get_sub_prop()` in all potentially risky cases?

google/cloud/bigquery/client.py
```diff
@@ -1314,6 +1315,73 @@ def job_from_resource(self, resource):
...
+            extract_job_config = google.cloud.bigquery.job.ExtractJobConfig.from_api_repr(
+                job_config
+            )
+            source = TableReference.from_api_repr(job_config["extract"]["sourceTable"])
+            destination_uris = _get_sub_prop(job_config, ["extract", "destinationUris"])
+            return self.extract_table(
+                source, destination_uris, job_config=extract_job_config, retry=retry
+            )
+        elif "query" in job_config:
+            del job_config["query"]["destinationTable"]
```
plamut: This will modify the `dict` passed in, which is probably undesired? IIRC there was an issue back then that addressed a very similar case by operating on a copy of the input argument.

HemangChothani:
@googlebot I consent

HemangChothani:
@googlebot I fixed it.

google/cloud/bigquery/client.py
```diff
@@ -1367,13 +1364,13 @@ def create_job(self, job_config, retry=DEFAULT_RETRY):
...
             )
-            source = TableReference.from_api_repr(job_config["extract"]["sourceTable"])
+            source = _get_sub_prop(job_config, ["extract", "sourceTable"])
             destination_uris = _get_sub_prop(job_config, ["extract", "destinationUris"])
             return self.extract_table(
                 source, destination_uris, job_config=extract_job_config, retry=retry
             )
         elif "query" in job_config:
-            del job_config["query"]["destinationTable"]
+            _del_sub_prop(job_config, ["query", "destinationTable"])
```
plamut: This will still potentially modify the input argument in-place, do we mind? (cc: @tswast )

tswast: Let's not do that if at all possible. Maybe make a copy before calling `_del_sub_prop`?

HemangChothani: @plamut , @tswast  Ok, may i do `copy_config = copy.deepcopy(job_config)` before calling `_del_sub_prop` ?

If  i consider `clear destination_table if it was a query job`  statement which mentioned in issue's description, need to modify the input argument.

plamut: @HemangChothani Creating a copy and operating on the latter sounds good, yes.

> If i consider `clear destination_table if it was a query job` statement which mentioned in issue's description, need to modify the input argument.

If I understood the ticket description correctly, it is not clear when a client library should clear the destination table property. But if we do clear it (as is the case with `query` jobs), it is safer to do it in a config copy, because users normally don't expect that their input parameter could be modified.

PR #36, opened 2020-02-12
feat: add BigQuery storage client support to DB API

### LINKED ISSUE #16: [BigQuery, BigQuery Storage]: Add option to use BigQuery Storage API to download results in BigQuery DB-API

In projects like superset, that use the SQL Alchemy connector, it can be quite slow to download large query results. The BigQuery Storage API speeds this up for `to_dataframe` / pandas, but not when used via the DB-API / SQL Alchemy. See: https://github.com/mxmzdlv/pybigquery/issues/41

When creating a DB-API Connection, provide a way to supply a BQ Storage client, in addition to a BQ client. Use this client to download results for the relevant methods in the Cursor object.

Could have a `use_bqstorage_api` option, but this would be inconsistent with the current constructor, which expects a client.

/cc @yiga2

plamut:
Closes #16.

The title says it all. Details in the ticket description.

PR checklist:

### DIFF
--- a/google/cloud/bigquery/dbapi/connection.py
+++ b/google/cloud/bigquery/dbapi/connection.py
@@ -23,10 +23,24 @@ class Connection(object):
 
     Args:
         client (google.cloud.bigquery.Client): A client used to connect to BigQuery.
+        bqstorage_client(\
+            Optional[google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient] \
+        ):
+            [Beta] An alternative client that uses the faster BigQuery Storage
+            API to fetch rows from BigQuery. If both clients are given,
+            ``bqstorage_client`` is used first to fetch query results,
+            with a fallback on ``client``, if necessary.
+
+            .. note::
+                There is a known issue with the BigQuery Storage API with small
+                anonymous result sets, which results in such fallback.
+
+                https://github.com/googleapis/python-bigquery-storage/issues/2
     """
 
-    def __init__(self, client):
+    def __init__(self, client, bqstorage_client=None):
         self._client = client
+        self._bqstorage_client = bqstorage_client
 
     def close(self):
         """No-op."""
@@ -43,17 +57,30 @@ class Connection(object):
         return cursor.Cursor(self)
 
 
-def connect(client=None):
+def connect(client=None, bqstorage_client=None):
     """Construct a DB-API connection to Google BigQuery.
 
     Args:
-        client (google.cloud.bigquery.Client):
-            (Optional) A client used to connect to BigQuery. If not passed, a
-            client is created using default options inferred from the environment.
+        client (Optional[google.cloud.bigquery.Client]):
+            A client used to connect to BigQuery. If not passed, a client is
+            created using default options inferred from the environment.
+        bqstorage_client(\
+            Optional[google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient] \
+        ):
+            [Beta] An alternative client that uses the faster BigQuery Storage
+            API to fetch rows from BigQuery. If both clients are given,
+            ``bqstorage_client`` is used first to fetch query results,
+            with a fallback on ``client``, if necessary.
+
+            .. note::
+                There is a known issue with the BigQuery Storage API with small
+                anonymous result sets, which results in such fallback.
+
+                https://github.com/googleapis/python-bigquery-storage/issues/2
 
     Returns:
         google.cloud.bigquery.dbapi.Connection: A new DB-API connection to BigQuery.
     """
     if client is None:
         client = bigquery.Client()
-    return Connection(client)
+    return Connection(client, bqstorage_client)
--- a/google/cloud/bigquery/dbapi/cursor.py
+++ b/google/cloud/bigquery/dbapi/cursor.py
@@ -21,12 +21,17 @@ try:
 except ImportError:  # Python 2.7
     import collections as collections_abc
 
+import logging
+
 import six
 
 from google.cloud.bigquery import job
 from google.cloud.bigquery.dbapi import _helpers
 from google.cloud.bigquery.dbapi import exceptions
 import google.cloud.exceptions
+
+
+_LOGGER = logging.getLogger(__name__)
 
 # Per PEP 249: A 7-item sequence containing information describing one result
 # column. The first two items (name and type_code) are mandatory, the other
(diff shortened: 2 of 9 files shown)


google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -192,11 +198,55 @@ def executemany(self, operation, seq_of_parameters):
...
+        destination = self._query_job.destination
+        table_reference = bigquery.table.TableReference(
+            bigquery.dataset.DatasetReference(
+                destination.project, destination.dataset_id,
+            ),
+            destination.table_id,
+        )
+
+        # TODO: same checks as in _pandas_helpers._download_table_bqstorage()?
+        # (rading from partitions and snapshots)
```
tswast: DB-API is only used to run queries and download results, so these cases shouldn't happen.

I suppose they could happen if someone writes to a specific partition by manually setting a destination table, but I don't believe we support this in the DB-API.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -192,11 +198,55 @@ def executemany(self, operation, seq_of_parameters):
...
         """Try to start fetching data, if not yet started.
 
         Mutates self to indicate that iteration has started.
         """
+        # TODO: fetch results... using bq_storage_client...
+        #  else: fallback to default client
+        #  (mention in docstring higher levels up?)
+
+        # TODO: add this known issue to docstring? just as in to_dataframe*()
+        # (about small result sets)
```
tswast: We should transparently fallback to tabledata.list with small results, just as we do in `to_dataframe`. Might be worth a mention in the docstring for the `bqstorage_client` in `connect`.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -212,6 +262,33 @@ def _try_fetch(self, size=None):
 
         if self._query_data is None:
             client = self.connection._client
+            bqstorage_client = self.connection._bqstorage_client
+
+            if bqstorage_client:
+
+                # TODO: Are we supposed to read somehting here in order to
```
tswast: I recall that the error with small results happens at `create_read_session` time, so I don't think reading anything is necessary.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -192,11 +198,55 @@ def executemany(self, operation, seq_of_parameters):
         for parameters in seq_of_parameters:
             self.execute(operation, parameters)
 
+    def _bqstorage_fetch(self, bqstorage_client, size):
```
tswast: We aren't actually using `size`, right? I don't believe that's a supported [read option](https://googleapis.dev/python/bigquerystorage/latest/gapic/v1beta1/types.html#google.cloud.bigquery_storage_v1beta1.types.TableReadOptions).

Any reason we have the parameter here? If so, let's document why it's present despite it not being used.

google/cloud/bigquery/dbapi/connection.py
```diff
@@ -23,10 +23,16 @@ class Connection(object):
 
     Args:
         client (google.cloud.bigquery.Client): A client used to connect to BigQuery.
+        bqstorage_client(\
+            Optional[google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient] \
+        ):
+            An alternative client to connect to BigQuery. If provided, it takes
```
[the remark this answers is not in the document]
plamut: Yeah, I should have reworded. One might believe that BQ storage client is a replacement for the BQ client, which is indeed not true.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -192,11 +198,55 @@ def executemany(self, operation, seq_of_parameters):
...
+        """TODO: docstring
+        """
+        # TODO: add assumption that query job exists check has been performed
+
+        # Since a BQ storage client instance is passed in, it means that
+        # bigquery_storage_v1beta1 library is available (no ImportError)
+        from google.cloud import bigquery_storage_v1beta1
+
+        destination = self._query_job.destination
+        table_reference = bigquery.table.TableReference(
```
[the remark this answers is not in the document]
plamut: I will re-check, might be a refactoring relic (I initially used plain dicts, but hit that unexpected "missing header" issue with them).

google/cloud/bigquery/dbapi/cursor.py
plamut: Indeed, that error already already happens before streaming the rows, that comment was already dated. :)

google/cloud/bigquery/dbapi/cursor.py
plamut: No reason, just passed in the arguments to `_try_fetch()` to this method initially. It actually turns out that even `_try_fetch()` ignores it completely.

plamut:
The coverage failure is due to Python 2.7 unit tests are not run on Kokoro (it's fine locally):
```
Session unit-2.7 skipped: Python interpreter 2.7 not found.
```

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -212,13 +217,74 @@ def _try_fetch(self, size=None):
 
         if self._query_data is None:
             client = self.connection._client
+            bqstorage_client = self.connection._bqstorage_client
+
+            if bqstorage_client:
```
tseaver:
```suggestion
            if bqstorage_client is not None:
```

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -212,13 +217,74 @@ def _try_fetch(self, size=None):
...
+                try:
+                    rows_iterable = self._bqstorage_fetch(bqstorage_client)
+                    self._query_data = _helpers.to_bq_table_rows(rows_iterable)
+                    return
+                except google.api_core.exceptions.Forbidden:
+                    # Don't hide errors such as insufficient permissions to create
+                    # a read session, or the API is not enabled. Both of those are
+                    # clearly problems if the developer has explicitly asked for
+                    # BigQuery Storage API support.
+                    raise
```
tseaver: Wgt catch an exception only to re-raise it without any further processing here? It seems likely to create a confusing traceback.

noxfile.py
```diff
@@ -48,7 +48,7 @@ def default(session):
         # Since many tests are skipped due to missing dependencies, test
         # coverage is much lower in Python 3.8. Remove once we can test with
         # pyarrow.
-        coverage_fail_under = "--cov-fail-under=92"
+        coverage_fail_under = "--cov-fail-under=91"
```
tseaver: Grrr.  We need to get rid of these majyk numbers altogether.  Individual unit test sessions don't need to be enforcing coverage, since we explicitly aggregate them, and require 100% on the aggregate.

tests/unit/test_dbapi_connection.py
```diff
@@ -27,19 +32,32 @@ def _get_target_class():
...
 
         mock_client = self._mock_client()
-        connection = self._make_one(client=mock_client)
+        mock_bqstorage_client = self._mock_bqstorage_client()
+        connection = self._make_one(
+            client=mock_client, bqstorage_client=mock_bqstorage_client,
+        )
         self.assertIsInstance(connection, Connection)
         self.assertIs(connection._client, mock_client)
+        self.assertIs(connection._bqstorage_client, mock_bqstorage_client)
```
tseaver: Seems like we need an explicit test for the constructor without the `bqstorage_client` argument passed?

google/cloud/bigquery/dbapi/cursor.py
plamut: This was modeled after a [similar logic](https://github.com/googleapis/google-cloud-python/blob/b387134827dbc3be0e1b431201e0875798002fda/bigquery/google/cloud/bigquery/table.py#L1407-L1436) in `table.RowItrator`.

I _think_ the intent was to explain why a particular subset of exceptions (Forbidden) does not result in a fallback to the default client, but a reraise instead, although that catch-reraise block is indeed redundant. I'll remove it, and only keep the comment.

noxfile.py
plamut: I'll put that to the agenda of the next client libs meeting.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -212,13 +217,76 @@ def _try_fetch(self, size=None):
...
+        # NOTE: Given that BQ storage client instance is passed in, it means
+        # that bigquery_storage_v1beta1 library is available (no ImportError).
+        from google.cloud import bigquery_storage_v1beta1
+
+        table_reference = self._query_job.destination
+
+        read_session = bqstorage_client.create_read_session(
+            table_reference.to_bqstorage(),
+            "projects/{}".format(table_reference.project),
+            # only a single stream only, as DB API is not well-suited for multithreading
```
plamut:
```suggestion
            # a single stream only, as DB API is not well-suited for multithreading
```

plamut:
@tseaver As a co-author of one of the changes, please just appease the CLA bot, thanks!

plamut:
> We'll want to update this from v1beta1->v1 in the near future, not sure how you'd like to track that.

@shollyman  Track in the sense of the tasks that need to be done for the promotion to GA? A GitHub issue for it would be just fine.

plamut:
@tseaver Friendly ping, the CLA bot still needs confirmation from you. Thanks!

tseaver:
@googlebot I signed it!

yiga2:
Can usage for this be updated at https://github.com/mxmzdlv/pybigquery/blob/master/README.rst#usage ?
Not sure how to enable storage API when loading in superset.

yiga2:
To be closed as per https://github.com/googleapis/python-bigquery-sqlalchemy/pull/61#issuecomment-756187394 ?

plamut:
@yiga2 Sorry, I'm not following? What should be closed, this PR has been merged?

yiga2:
My bad - meant to close https://github.com/googleapis/python-bigquery-sqlalchemy/issues/41

PR #47, opened 2020-02-26
chore: declare full Python 3.8 support

plamut:
Fixes #12.

This PR declares full support for Python 3.8 and uses/includes the latter to run various nox sessions.

### PR checklist
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

PR #48, opened 2020-02-26
chore: deprecate pandas code paths that don't use pyarrow

### LINKED ISSUE #3: BigQuery: deprecate pandas code paths that do not use pyarrow
In the pandas-related BigQuery code, there is a lot of branching on whether `pyarrow` is available or not, and significant portions of business logic dealing with both cases.

The `pyarrow` side is often more efficient, more concise, and less prone to weird edge cases, thus it is preferred that it eventually becomes the only code path.

The goal of this issue is to emit deprecation warnings whenever a code path is hit that deals with `pandas`, but without the `pyarrow` dependency available.

plamut:
Fixes #3.

This PR adds warnings when a pandas code path is taken without the `pyarrow` dependency available.

### PR checklist

### DIFF
--- a/google/cloud/bigquery/__init__.py
+++ b/google/cloud/bigquery/__init__.py
@@ -38,6 +38,7 @@ from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetReference
 from google.cloud.bigquery import enums
 from google.cloud.bigquery.enums import StandardSqlDataTypes
+from google.cloud.bigquery.exceptions import PyarrowMissingWarning
 from google.cloud.bigquery.external_config import ExternalConfig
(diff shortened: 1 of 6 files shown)


plamut:
I'll check tomorrow why Python 3.8 unit tests fail on Kokoro, we might need to merge #47 first, or maybe Kokoro does not yet have full support for Python 3.8?
(the tests pass locally just fine)

plamut:
Figured it out, it's indeed #47 that has to be merged first, because currently `pyarrow` is [not installed](https://github.com/googleapis/python-bigquery/pull/47/files#diff-37409d3afed5d01396a659e1b20bdf08L46-L53) when running unit tests with Python 3.8, resulting in different code paths taken.

PR #52, opened 2020-03-06
docs: use timeout in create dataset sample

### LINKED ISSUE #8: BigQuery: Document the use of the timeout parameter in samples
After adding the new `timeout` parameter to various public methods (#9987), we should demonstrate its usage in the code samples.

Users should be aware of this new feature, and should probably use it by default to avoid sporadic weird issues related to a method "getting stuck" at the transport layer.

plamut:
Fixes #8.

### PR checklist

### DIFF
--- a/samples/create_dataset.py
+++ b/samples/create_dataset.py
@@ -30,7 +30,7 @@ def create_dataset(dataset_id):
     # TODO(developer): Specify the geographic location where the dataset should reside.
     dataset.location = "US"
 
-    # Send the dataset to the API for creation, with a timeout.
+    # Send the dataset to the API for creation, with an explicit timeout.
     # Raises google.api_core.exceptions.Conflict if the Dataset already
     # exists within the project.
     dataset = client.create_dataset(dataset, timeout=30)  # Make an API request.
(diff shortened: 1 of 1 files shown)


plamut:
From the [issue thread](https://github.com/googleapis/python-bigquery/issues/8#issuecomment-566785126):
> Makes sense to me. I think that once we have a timeout parameter on all relevant methods, we can surface an example in an API best practices doc such as https://cloud.google.com/bigquery/docs/api-performance

@tswast What's the repo for changing that, is it public? If not, somebody else should update those docs and mention timeouts among recommended good practices.

tswast:
> @tswast What's the repo for changing that, is it public? If not, somebody else should update those docs and mention timeouts among recommended good practices.

It's in `google3` (private). @shollyman or one of the other BQ DevRel folks should be able to take care of that.

samples/create_dataset.py
```diff
@@ -30,9 +30,9 @@ def create_dataset(dataset_id):
     # TODO(developer): Specify the geographic location where the dataset should reside.
     dataset.location = "US"
 
-    # Send the dataset to the API for creation.
+    # Send the dataset to the API for creation, with a timeout.
```
tswast: Let's say "explicit timeout", since we have default values for timeouts now.

plamut:
```suggestion
    # Send the dataset to the API for creation, with an explicit timeout.
```

PR #55, opened 2020-03-09
feat: use BigQuery Storage client by default

plamut:
No issue yet, just a POC preview.

### DIFF
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -3220,7 +3220,7 @@ class QueryJob(_AsyncJob):
         self,
         progress_bar_type=None,
         bqstorage_client=None,
-        create_bqstorage_client=False,
+        create_bqstorage_client=True,
     ):
         """[Beta] Create a class:`pyarrow.Table` by loading all pages of a
         table or query.
@@ -3255,11 +3255,10 @@ class QueryJob(_AsyncJob):
                 Reading from a specific partition or snapshot is not
                 currently supported by this method.
             create_bqstorage_client (bool):
-                **Beta Feature** Optional. If ``True``, create a BigQuery
-                Storage API client using the default API settings. The
-                BigQuery Storage API is a faster way to fetch rows from
-                BigQuery. See the ``bqstorage_client`` parameter for more
-                information.
+                Optional. If ``True`` (default), create a BigQuery Storage API
+                client using the default API settings. The BigQuery Storage API
+                is a faster way to fetch rows from BigQuery. See the
+                ``bqstorage_client`` parameter for more information.
 
                 This argument does nothing if ``bqstorage_client`` is supplied.
 
@@ -3290,15 +3289,15 @@ class QueryJob(_AsyncJob):
         bqstorage_client=None,
         dtypes=None,
         progress_bar_type=None,
-        create_bqstorage_client=False,
+        create_bqstorage_client=True,
     ):
         """Return a pandas DataFrame from a QueryJob
 
         Args:
             bqstorage_client (google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient):
-                **Alpha Feature** Optional. A BigQuery Storage API client. If
-                supplied, use the faster BigQuery Storage API to fetch rows
-                from BigQuery. This API is a billable API.
+                Optional. A BigQuery Storage API client. If supplied, use the
+                faster BigQuery Storage API to fetch rows from BigQuery. This
+                API is a billable API.
 
                 This method requires the ``fastavro`` and
                 ``google-cloud-bigquery-storage`` libraries.
@@ -3325,11 +3324,10 @@ class QueryJob(_AsyncJob):
 
                 ..versionadded:: 1.11.0
             create_bqstorage_client (bool):
-                **Beta Feature** Optional. If ``True``, create a BigQuery
-                Storage API client using the default API settings. The
-                BigQuery Storage API is a faster way to fetch rows from
-                BigQuery. See the ``bqstorage_client`` parameter for more
-                information.
+                Optional. If ``True`` (default), create a BigQuery Storage API
+                client using the default API settings. The BigQuery Storage API
+                is a faster way to fetch rows from BigQuery. See the
+                ``bqstorage_client`` parameter for more information.
 
                 This argument does nothing if ``bqstorage_client`` is supplied.
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1467,7 +1467,7 @@ class RowIterator(HTTPIterator):
         self,
         progress_bar_type=None,
         bqstorage_client=None,
-        create_bqstorage_client=False,
+        create_bqstorage_client=True,
     ):
         """[Beta] Create a class:`pyarrow.Table` by loading all pages of a
         table or query.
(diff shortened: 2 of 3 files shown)


plamut:
I did some timings to compare fetching data with and without the BQ Storage client, results below. Spoiler: When using the BQ Storage API, performance improvements are enormous.

#### Query
```
SELECT * FROM `project.dataset.table`
LIMIT 200000
```
#### Source tables
- A synthetic table with two FLOAT columns containing 10M rows of random numbers
- `bigquery-public-data-cms_medicare.inpatient_charges_2011` -  multiple numeric and string columns, contains ~163k rows.

#### Timings
|  | REST API (tabledata.list) | BQ Storage API |
|---|---|---|
| Synthetic table | 10 - 11.5 s| 4.5 - 5 s |
| Inpatient charges table | 23 - 25s | 10.5 - 12.5 s|

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -577,8 +577,7 @@ def _bqstorage_page_to_dataframe(column_names, dtypes, page):
 def _download_table_bqstorage_stream(
     download_state, bqstorage_client, session, stream, worker_queue, page_to_item
 ):
-    position = bigquery_storage_v1beta1.types.StreamPosition(stream=stream)
-    rowstream = bqstorage_client.read_rows(position).rows(session)
+    rowstream = bqstorage_client.read_rows(stream.name).rows(session)
```
tswast, line 580: Can we detect when a v1beta1 client is passed in and still use `StreamPosition`? Otherwise, this is a breaking change for `pandas-gbq` and customer code that used the BQ Storage API before this change.

plamut: We can check the client type, yes, and construct the read session and its parameters differently. It should only be necessary to have this check at a few limited places, thus the extra noise should not be too bad, especially if we prevent a severe breaking change.

shollyman: Supporting a transition period where we support either storage client makes sense.  I'd love to emit deprecation warnings for the v1beta1 client case so people can identify they should migrate sooner rather than later.  We should pick a reasonable drop dead date for this as well so we're not carrying it indefinitely if we can avoid it.

A quick scan of api core blunted my hopes of trying to do something clever like using a v1beta1 client to build a v1 client, it looks like there's too much we could get wrong with transports and such.

plamut: Done, and seems to work (except for the deprecation warning that I still need to add).

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -256,34 +260,63 @@ def _bqstorage_fetch(self, bqstorage_client):
...
-        read_session = bqstorage_client.create_read_session(
-            table_reference.to_bqstorage(),
-            "projects/{}".format(table_reference.project),
-            # a single stream only, as DB API is not well-suited for multithreading
-            requested_streams=1,
+        is_v1beta1_client = isinstance(
+            bqstorage_client, bigquery_storage_v1beta1.BigQueryStorageClient
         )
 
+        # We want to preserve comaptibility with the v1beta1 BQ Storage clients,
```
shollyman, line 282: nit: compatibility

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -256,34 +260,63 @@ def _bqstorage_fetch(self, bqstorage_client):
...
+            read_session = bqstorage_client.create_read_session(
+                table_reference.to_bqstorage(v1beta1=True),
+                "projects/{}".format(table_reference.project),
+                # a single stream only, as DB API is not well-suited for multithreading
+                requested_streams=1,
+            )
+        else:
+            requested_session = bigquery_storage_v1.types.ReadSession(
+                table=table_reference.to_bqstorage(),
+                data_format=bigquery_storage_v1.enums.DataFormat.AVRO,
```
shollyman, line 299: I thought our fastpath in python was definitely arrow.  Is there something we're getting from passing through avro?  Or are we in a path where we have the storage import but not arrow?

google/cloud/bigquery/job.py
```diff
@@ -3247,7 +3247,7 @@ def to_arrow(
                 ``'tqdm_gui'``
                   Use the :func:`tqdm.tqdm_gui` function to display a
                   progress bar as a graphical dialog box.
-            bqstorage_client (google.cloud.bigquery_storage_v1beta1.BigQueryStorageClient):
+            bqstorage_client (google.cloud.bigquery_storage_v1.BigQueryReadClient):
                 **Beta Feature** Optional. A BigQuery Storage API client. If
```
shollyman, line 3251: nit: Still beta with this change?

setup.py
```diff
@@ -39,7 +39,7 @@
 ]
 extras = {
     "bqstorage": [
-        "google-cloud-bigquery-storage >= 0.6.0, <2.0.0dev",
+        "google-cloud-bigquery-storage >= 0.8.0, <2.0.0dev",
```
shollyman, line 42: Independent of this change, but we should consider bumping this to a 1.x now that we have a GA v1.

tests/system.py
```diff
@@ -1599,10 +1601,60 @@ def test_dbapi_fetch_w_bqstorage_client_small_result_set(self):
...
+        bigquery_storage_v1 is None, "Requires `google-cloud-bigquery-storage`"
     )
     @unittest.skipIf(fastavro is None, "Requires `fastavro`")
     def test_dbapi_fetch_w_bqstorage_client_large_result_set(self):
+        bqstorage_client = bigquery_storage_v1.BigQueryReadClient(
+            credentials=Config.CLIENT._credentials
+        )
+        cursor = dbapi.connect(Config.CLIENT, bqstorage_client).cursor()
+
+        # Pick a large enouhg LIMIT value to assure that the fallback to the
```
shollyman, line 1613: nit: s/enouhg/enough/

tests/system.py
```diff
@@ -1599,10 +1601,60 @@ def test_dbapi_fetch_w_bqstorage_client_small_result_set(self):
...
+        # default client is not needed due to the result set being too small
+        # (a known issue that causes problems when reding such result sets with
```
shollyman, line 1615: So, the fix for small result sets is rolling out at we speak, and we shouldn't need to fallback in this case. We can leave this in here for now and clean it up when it's rolled out.

google/cloud/bigquery/job.py
plamut: Shouldn't be, since BQ Storage is now stable. Forgot to remove the `beta` mark here.

tests/system.py
plamut: Noted. Will create a ticket to clean it up when ready.

setup.py
plamut: I don't see this published on [PyPI](https://pypi.org/project/google-cloud-bigquery-storage/#history) yet? I used `0.8.0` because it was the most recent version at that time.

google/cloud/bigquery/dbapi/cursor.py
plamut: It is ARROW, indeed. This seems like a copy-pasta error, I think I mistakenly took some of the lines from one of the samples that happened to use the AVRO format.

plamut: Hmm, using ARROW breaks a few system tests, I'll have to check whether this is legitimate or is it just tests that need to be adjusted.

plamut: Now I remember - I initially chose AVRO, because that's the default the BQ Storage `v1beta1` client uses, and I wanted to match that with the v1 stable client.

When switching to ARROW it turned out that DB API actually never worked with ARROW format, but fortunately system tests caught that. The issue is now fixed and DB API uses ARROW now (both with v1beta1 and v1 stable BQ Storage clients).

setup.py
shollyman: Sorry, I wasn't clear.  I meant we should do a release of storage to bump the major version to indicate the stability level more clearly now that v1 is GA.  I'll see about getting this done tomorrow.

plamut:
Just a thought - since this is a pretty significant change, how about releasing all the other smaller changes and fixes that have accumulated since the last release separately, and release this one separately for an easier rollback, should that be necessary?

shollyman:
Isolating this change to its own release seems prudent.

plamut:
Putting this on hold until the next release is made, we will ship this separately.

PR #58, opened 2020-03-10
fix: improve cell magic error message on missing query

### LINKED ISSUE #57: Uninformative Jupyter cell magic error if the query is missing
### Environment details

BigQuery client v1.24.0, all Python versions.

### Steps to reproduce

1. Fire up a Jupyter notebook or an IPython session and load the `bigquery` extension:
    ```
     %load_ext google.cloud.bigquery
    ```

2. Use the `bigquery` cell magic without specifying the query:
    ```
    %%bigquery --project my-project --destination_table my_dataset.my_table
    ```
**Actual result:**
An error message is displayed, but is kind of misleading:
```
ERROR:
 table_id must be a fully-qualified ID in standard SQL format, e.g., "project.dataset.table_id", got 
```
It might make the user wonder why the table ID is incorrect even if `--destination` is correctly specified.

**Expected result:**
The error message should explain that the query is missing in the cell body.

### Proposed solution
The [current implementation](https://github.com/googleapis/python-bigquery/blob/v1.24.0/google/cloud/bigquery/magics.py#L507-L511) assumes that any query without whitespace represents a table ID, but it fails to check if the query is non-empty in the first place. The latter check should be added.

plamut:
Fixes #57.

This PR fixes the Jupyter cell magic error message if query is missing. The steps to reproduce/verify the fix are in the issue description.

### PR checklist

### DIFF
--- a/google/cloud/bigquery/magics.py
+++ b/google/cloud/bigquery/magics.py
@@ -68,8 +68,7 @@
     * ``<query>`` (required, cell argument):
         SQL query to run. If the query does not contain any whitespace (aside
         from leading and trailing whitespace), it is assumed to represent a
-        fully-qualified table ID, and the latter's data will be fetched and
-        returned as ``pandas.DataFrame``.
+        fully-qualified table ID, and the latter's data will be fetched.
 
     Returns:
         A :class:`pandas.DataFrame` with the query results.
--- a/google/cloud/bigquery/magics.py
+++ b/google/cloud/bigquery/magics.py
@@ -66,7 +66,10 @@
           the variable name (ex. ``$my_dict_var``). See ``In[6]`` and ``In[7]``
           in the Examples section below.
     * ``<query>`` (required, cell argument):
-        SQL query to run.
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/magics.py
```diff
@@ -66,7 +66,10 @@
           the variable name (ex. ``$my_dict_var``). See ``In[6]`` and ``In[7]``
           in the Examples section below.
     * ``<query>`` (required, cell argument):
-        SQL query to run.
+        SQL query to run. If the query does not contain any whitespace (aside
+        from leading and trailing whitespace), it is assumed to represent a
+        fully-qualified table ID, and the latter's data will be fetched and
```
shollyman: nit:  possibly remove the extra reference in the docstring to returning dataframe here?  As written I expected this to mean that this special case returned something different.

plamut: Sure, why not, let's remove this possible source of confusion then.

Issue #61, opened 2020-03-17
A system test for load_table_from_datarame() consistently fails on master branch


plamut:
A system test `test_load_table_from_dataframe_w_explicit_schema()` consistently fails on the latest `master` branch, both under Python 2.7 and Python 3.8 (example [Kokoro run](https://source.cloud.google.com/results/invocations/0db81847-def8-4d23-be0e-446b4855ad7f/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fpresubmit/log)). It is also consistently reproducible locally.

```
...
google.api_core.exceptions.BadRequest: 400 Error while reading data, error message: Invalid datetime value -62135596800000000 for field 'dt_col' of type 'INT64' (logical type 'TIMESTAMP_MICROS'): generic::out_of_range: Cannot return an invalid datetime value of -62135596800000000 microseconds relative to the Unix epoch. The range of valid datetime values is [0001-01-1 00:00:00, 9999-12-31 23:59:59.999999]
```

Ticket in the Google issue tracker: https://issuetracker.google.com/issues/151765076

plamut:
@shollyman Have there been any changes on the backend in the last week or so? The system tests passed fine when #58 was initially submitted, but now fail on the latest `master`.

Specifically, have there been any changes to handling the TIMESTAMP_MICROS logical type?

shollyman:
Let's grab the avro file that's created indirectly from the test, use that to file an issue with the backend team to aid them reproducing it: https://issuetracker.google.com/issues/new?component=187149&template=0.

The kokoro log:
BadRequest: 400 Error while reading data, error message: Invalid datetime value -62135596800000000 for field 'dt_col' of type 'INT64' (logical type 'TIMESTAMP_MICROS'): generic::out_of_range: Cannot return an invalid datetime value of -62135596800000000 microseconds relative to the Unix epoch. The range of valid datetime values is [0001-01-1 00:00:00, 9999-12-31 23:59:59.999999]

However, this is the value for the min allowed datetime, the query engine agrees:
SELECT TIMESTAMP_MICROS(-62135596800000000) ==> 0001-01-01 00:00:00 UTC

Peter, can you take care of this?

plamut:
OK, I'll intercept the file the test generates and submits to the backend, and open a ticket in the issuetracker.

plamut:
[...]

PR #64, opened 2020-03-23
docs: include details of inherited class members

### LINKED ISSUE #5: BigQuery google.cloud.bigquery.job.LoadJob methods return types documentation
Good day,
I am using the Python client libraries for BigQuery. I noticed in the documentation that the google.cloud.bigquery.job.LoadJob methods do not indicate return types. They only seem to give a brief explanation of the method. 

In particular, I am looking for what the return type of the result() method is. I noticed that in the source code, queryJob.result() returns an EmptyRowIterator, but I am not able to find out whether this is the case for loadJob's as there doesn't seem to be a  result() method defined for the LoadJob class. Is the result  used in this context that from google.api_core.future.polling.PollingFuture?

My goal is simply to determine whether the load job was successful or not.

Would it be possible to add some information on the return types of the methods, as is the case with other classes in the library?

plamut:
Fixes #5.

This PR fixes the docs for inherited class members, their details are no longer omitted. To test, generate the docs locally and verify that e.g. `LoadJob.result()` details are now actually present.

### PR checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,7 +42,7 @@ extensions = [
 
 # autodoc/autosummary flags
 autoclass_content = "both"
-autodoc_default_flags = ["members"]
+autodoc_default_flags = ["members", "inherited-members"]
 autosummary_generate = True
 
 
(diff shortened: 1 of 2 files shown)


plamut:
FWIW, the failed test has nothing to do with the changes here, the issue is tracked [here](https://github.com/googleapis/python-bigquery/issues/61).

PR #76, opened 2020-04-09
docs: fix generation warnings in new Sphinx version

### LINKED ISSUE #75: Building the docs breaks (duplicate object description)
I noticed that building the docs locally on `master` fails due to the following error:
> sphinx.errors.SphinxWarning: <workspace_dir>/python-bigquery/google/cloud/bigquery_v2/types.py:docstring of google.cloud.bigquery_v2.types.DeleteModelRequest.dataset_id:1:duplicate object description of google.cloud.bigquery_v2.types.DeleteModelRequest.dataset_id, other instance in gapic/v2/types, use :noindex: for one of them

Checking what gets injected into the `types` module does not reveal anything, the `DeleteModelRequest` type is **not** duplicated.

It turns out that the issue does not occur with the older version of Sphinx (< 3.x), ~~thus it appears as a bug in the Sphinx package itself.~~ A workaround is to pin that dependency to an older version.

Preferably, though, the new warnings emitted in Sphinx 3 should be addressed at their root causes.

plamut:
Fixes #75.

This PR fixes adds a workaround to the docs generation error that started occurring recently by restricting Sphinx to older versions.

To test, try generating the docs with and without the change. Without the change, generating the docs failed due to the warning mentioned in the issue description.

### PR checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,6 +38,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.todo",
     "sphinx.ext.viewcode",
+    "recommonmark",
 ]
 
 # autodoc/autosummary flags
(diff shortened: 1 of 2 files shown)


plamut:
@HemangChothani The PR now actually resolves all the warnings, meaning that the docs build should pass again even with Sphinx 3.x

HemangChothani:
@plamut Just asking for my reference,  `.rst` files in  docs/gapic folder are generated files or handwritten.

plamut:
@HemangChothani It appears they have been added manually ([commit](https://github.com/googleapis/python-bigquery/commit/9678ec972d2ef31c092a274a2fd2f733a036ef0b)) when Models API support was added.

PR #80, opened 2020-04-21
chore: Migrate python-bigquery synth.py from artman to bazel

### LINKED ISSUE #81: chore: replace Artman with bazel for synthesizing code
The synthtool should start using bazel instead of Artman.

vam-google:
Closes #81.

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -17,16 +17,14 @@
 import synthtool as s
 from synthtool import gcp
 
-gapic = gcp.GAPICGenerator()
+gapic = gcp.GAPICBazel()
 common = gcp.CommonTemplates()
 version = 'v2'
 
 library = gapic.py_library(
-    'bigquery',
-    version,
-    config_path='/google/cloud/bigquery/'
-                'artman_bigquery_v2.yaml',
-    artman_output_name='bigquery-v2',
+    service='bigquery',
+    version=version,
+    bazel_target=f"//google/cloud/bigquery/{version}:bigquery-{version}-py",
     include_protos=True,
 )
 


synth.py
```diff
@@ -17,16 +17,14 @@
...
 
 library = gapic.py_library(
-    'bigquery',
-    version,
-    config_path='/google/cloud/bigquery/'
-                'artman_bigquery_v2.yaml',
-    artman_output_name='bigquery-v2',
+    service='bigquery',
+    version=version,
+    bazel_target=f"//google/cloud/bigquery/{version}:bigquery-{version}-py",
```
plamut, line 27: The library still needs to support Python 2.7 and 3.5, thus f-strings cannot be used, unfortunately.

plamut:
@vam-google BTW, are there any specific steps to test this change, or is bazel simply a drop-in replacement with no strings attached?

vam-google:
@plamut It is 99% drop-in replacement. It is not 100% because of the following differences, all of which are supposed to be ok:

1) Slightly different version (minor update) of protoc/grpc may be used from what what was used during last release. This results in different python syntax used in the generated stubs (`dict()` in docs is replaced with `{}`, `b""` string syntax is used)
2) Slightly different formatting of documentation (line breaks for some of the longer lines can be put earlier now)
3) In rare cases, there can be difference in _GAPIC_LIBRARY_VERSION value (which should be fixed in synth.py if it is not a correct one), and/or files may be in a slightly different directory (which should also be fixed in synth.py if that is the case).

Those are basically the things to look for, but overall it should be generating completely backward-compatible client with no visible/breaking changes for anybody.

You can try generating it (works same way as before - just run synthtool).

synth.py
vam-google: Synth.py just calls bazel, and for what bazel is doing, it is ok with any version of python. And yes, synth.py is not what end users call, it is used just to generate the library itself.

plamut:
@vam-google Yeah, just got the same info through another channel, thanks!

Will try it out tomorrow, and if it works (it should), I'll happily merge this.

PR #89, opened 2020-04-29
chore: remove local test_utils directory

### LINKED ISSUE #88: Remove test_utils directory
Now that test utils are available in a GitHub repo, a local `test_utils` copy is not needed anymore and should be removed.

plamut:
Closes #88. 🦕

Test utils are now available in their own repo, meaning that we can remove the local copy of it.

### PR checklist


plamut:
I noticed that system test failure on another branch, too, but thought it has something to do with changes I was making to how BQ Storage client is used. I now see that it can also happen on `master` with a few unrelated modifications.

I cannot reproduce it locally, however, I only observed it on Kokoro. I restarted the checks, and if they fail again for the same reason, I will submit [the workaround](https://github.com/googleapis/python-bigquery/pull/55/files#diff-af76fe2460d060c4e74d44c75d35074dR2639-R2642) from another branch in a new PR.

PR #93, opened 2020-05-06
feat: add a __hash__ implementation to AccessEntry

### LINKED ISSUE #94: Make AccessEntry hashable
I would like to make AccessEntry hashable. I ran into it being unhashable and could not see a reason why.

worlvlhole:
added `__hash__` to AccessEntry 

Closes #94.

### DIFF
--- a/google/cloud/bigquery/dataset.py
+++ b/google/cloud/bigquery/dataset.py
@@ -186,11 +186,7 @@ class AccessEntry(object):
         Returns:
             Tuple: The contents of this :class:`~google.cloud.bigquery.dataset.AccessEntry`.
         """
-        return (
-            self._role,
-            self._entity_type,
-            self._entity_id
-        )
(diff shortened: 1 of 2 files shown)


worlvlhole:
https://github.com/googleapis/python-bigquery/issues/94

worlvlhole:
added two test

shollyman:
The pattern I can think of where mutable ACE mutations be leveraged is for code that amends an ACL via dataset update.  There are cases where walking the existing ACL to amend the entries might be happening via direct mutation rather than replacing via a newly constructed ACE.

I tried to search for code examples that mutate ACEs in place, but I didn't find anything with a coarse pass.  The much more common idiom is appending an entry to an existing ACL.  I can't think of any existing integrations that would do this, it feels like a very specific action.

This feels relatively low risk, and not sufficient to necessitate a major rev in release to me.  Peter, let's make sure we call this out clearly in the release notes however.

PR #110, opened 2020-05-18
fix: make TimePartitioning repr evaluable

### LINKED ISSUE #109: TimePartitioning printable representation is not evaluable
#### Environment details

  - OS type and version: `All`
  - Python version: `All`
  - pip version: `All`
  - google-cloud-bigquery version: `1.24.0`

#### Code example

```python
from google.cloud.bigquery import *
eval(TimePartitioning(field='myfield').__repr__())
```

#### Stack trace
```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-2705370d2a8c> in <module>
      1 from google.cloud.bigquery import *
----> 2 eval(TimePartitioning(field='myfield').__repr__())
<string> in <module>
NameError: name 'myfield' is not defined
```

#### Code example

```python
from google.cloud.bigquery import *
eval(TimePartitioning(field='myfield',type='DAY').__repr__())
```

#### Stack trace
```

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-4733c3100f2b> in <module>
      1 from google.cloud.bigquery import *
----> 2 eval(TimePartitioning(field='myfield',type='DAY').__repr__())
TypeError: __init__() got an unexpected keyword argument 'type'
```

francois-baptiste:
Fixes #109  🦕


francois-baptiste:
@googlebot I signed it!

google/cloud/bigquery/table.py
```diff
@@ -2005,13 +2005,13 @@ class TimePartitioning(object):
     """
 
     def __init__(
-        self, type_=None, field=None, expiration_ms=None, require_partition_filter=None
+        self, type=None, field=None, expiration_ms=None, require_partition_filter=None
```
plamut, line 2008: Modifying a signature of a public method can break user code, unfortunately, and I don't know if the gain (evaluability of repr) is worth it. The linter might also complain, because the new parameter name shadows the `type` builtin.

google/cloud/bigquery/table.py
```diff
@@ -2142,7 +2142,7 @@ def __hash__(self):
         return hash(self._key())
 
     def __repr__(self):
-        key_vals = ["{}={}".format(key, val) for key, val in self._key()]
+        key_vals = ["{}={}".format(key, val.__repr__()) for key, val in self._key()]
```
plamut, line 2145:
```suggestion
        key_vals = ["{}={}".format(key, repr(val)) for key, val in self._key()]
```
(nit)
A bit more readable this way, IMO.

google/cloud/bigquery/table.py
francois-baptiste: Yes, I agree, I will revert this and propose a change to output repr with `type_ ` parameter instead.

google/cloud/bigquery/table.py
francois-baptiste: Yes I agree!

francois-baptiste:
Thank you for your help @plamut.
I reworked the pull request from scratch. The properties from `TimePartitioning `object are now converted from the REST API to the python API in the `repr` method.

google/cloud/bigquery/table.py
```diff
@@ -2128,7 +2128,14 @@ def to_api_repr(self):
         return self._properties
 
     def _key(self):
-        return tuple(sorted(self._properties.items()))
+        properties = self._properties.copy()
```
plamut, line 2131: (nit)
It might be a good idea to explicitly explain in a short comment why shallow copy is sufficient here (i.e. we are only "renaming" top level keys).

tests/unit/test_table.py
```diff
@@ -3752,7 +3752,7 @@ def test___repr___explicit(self):
         time_partitioning = self._make_one(
             type_=TimePartitioningType.DAY, field="name", expiration_ms=10000
         )
-        expected = "TimePartitioning(" "expirationMs=10000," "field=name," "type=DAY)"
+        expected = "TimePartitioning(" "expiration_ms=10000," "field=name," "type_='DAY')"
```
plamut, line 3755: (nit)
Three implicitly concatenated strings are a leftover from a past blacken session and we can get rid of it while already modifying this line.
```suggestion
        expected = "TimePartitioning(expiration_ms=10000,field=name,type_='DAY')"
```

google/cloud/bigquery/table.py
```diff
@@ -2142,7 +2149,7 @@ def __hash__(self):
         return hash(self._key())
 
     def __repr__(self):
-        key_vals = ["{}={}".format(key, val) for key, val in self._key()]
+        key_vals = ["{}={}".format(key, repr(val)) for key, val in self._key()]
```
plamut, line 2152: The side effect of the `repr()` is that numeric values are now presented as strings, which breaks one of the unit tests:
```py
>>> TimePartitioning(expiration_ms=42)                                                                                                                                                                                                                                                                                                                                                                
TimePartitioning(expiration_ms='42',type_='DAY')
```
Even though the property still works as expected, because it's converted from/to string on get/set, we should really fix that here in the `__repr__()`, as those internal conversions should be considered an implementation detail.

google/cloud/bigquery/table.py
francois-baptiste: You are right, better be explicit here.

google/cloud/bigquery/table.py
francois-baptiste: Yes I eventually call `repr` for non built-in type objects only in the `_key` method.

tests/unit/test_table.py
```diff
@@ -3752,7 +3752,7 @@ def test___repr___explicit(self):
         time_partitioning = self._make_one(
             type_=TimePartitioningType.DAY, field="name", expiration_ms=10000
         )
-        expected = "TimePartitioning(" "expirationMs=10000," "field=name," "type=DAY)"
+        expected = "TimePartitioning(expiration_ms=10000,field=name,type_='DAY')"
```
plamut, line 3755: This test still fails, the new expectation of a string value should be quoted now that has been made evaluable:
```suggestion
        expected = "TimePartitioning(expiration_ms=10000,field='name',type_='DAY')"
```

francois-baptiste:
@plamut I have fix lint and coverage. Because of the coverage, I had to test over `TimePartitioning.require_partition_filter` field even if it is deprecated according to docs.

francois-baptiste:
@plamut Can you please help me understand why code coverage fails?

```
Name                                                Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------
google/cloud/bigquery/table.py                        770      0    252      1    99%   2133->2136
```
On my local machine, code coverage pass...
Thank you.

plamut:
@francois-baptiste Perhaps the coverage stats include previous runs when the coverage was fine? (the stats are accumulating)

Can you reproduce by removing the `.coverage` file and run the tests again? (both 2.7 and at least one Python 3 version).

If not, I will take a look as it indeed seems that the missing part should be covered by the updated test, at least at first sight.

google/cloud/bigquery/table.py
```diff
@@ -2128,7 +2128,21 @@ def to_api_repr(self):
         return self._properties
 
     def _key(self):
-        return tuple(sorted(self._properties.items()))
+        # because we are only "renaming" top level keys shallow copy is sufficient here.
+        properties = self._properties.copy()
+        if "type" in properties.keys():
+            # calling repr for non built-in type objects.
+            properties["type_"] = repr(properties.pop("type"))
```
plamut, line 2135: @francois-baptiste Had a look and the reason for the coverage complaint is the if statement above. It is unnecessary for "type", because the latter always exists - if it's not given on instantiation, a default value is for it is stored.

The coverage check complained about the jump from line 2133 to 2136, i.e. the "negative" branching condition that was never hit.

(unrelated, minor) Dict containment can be directly checked, e.g. `if "foo" in dict`, it is not necessary to invoke the `.keys()` method first.

francois-baptiste: Thank you for your help @plamut and the detailed explanation! I have make the change accordingly and revert my unfortunate fix.

shollyman:
Sorry, getting back to this one.  There's an internal issue (155198865 for any googlers following along) related to null behavior being undefined.  I've been trying to get that finalized.  We declare that time partitioning type is required, but in practice it's not enforced and defaults to DAY.  Before we commit on that, I want to ensure we're documenting this and treat it as defined.

unset field and field==none should be equivalent.  It's simply toggling between column partition and ingestion time partitioning.

re: requirePartitionFilter in the backend, that's deprecated in the existing definition, and is actually a shadow of the parent value (the same name, just outside the TimePartitioningMessage).  The ambiguities come in during conflicting dual mutations to both fields, but shouldn't be in scope for this change.

plamut:
@shollyman Friendly ping. :) Any updates on those internal questions to date?

shollyman:
Documentation for behavior is getting updated.   Lack of value will be treated as DAY interval.

tswast (approved):

PR #121, opened 2020-06-04
feat(bigquery): expose start index parameter for query result

### LINKED ISSUE #120: Why QueryJob.result() function doesn't support 'start_index' parameter?
Why QueryJob.result function doesn't support a 'start_index' parameter for reading data by page?
I know the JAVA version support this feature, and I read the code, found easy to add this parameter because list_rows() function support this, and result() function call list_row() at last.

I need like this:
        query_job = self.client.query(sql)
        return query_job.result(start_index=100, page_size=20)

Thanks.

HemangChothani:
Fixes #120

### DIFF
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -3160,7 +3160,12 @@ class QueryJob(_AsyncJob):
             raise
 
     def result(
-        self, page_size=None, max_results=None, retry=DEFAULT_RETRY, timeout=None
+        self,
+        start_index=None,
+        page_size=None,
+        max_results=None,
+        retry=DEFAULT_RETRY,
+        timeout=None,
     ):
         """Start the job and wait for it to complete and get the result.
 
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/job.py
```diff
@@ -3170,6 +3175,8 @@ def result(
                 request. Non-positive values are ignored.
             max_results (Optional[int]):
                 The maximum total number of rows from this request.
+            start_index (int):
+                (Optional) The zero-based index of the starting row to read.
```
plamut, line 3179: Let's move the word "optional" to the type itself, i.e. `Optional[int]`, to be consistent with other parameters.

google/cloud/bigquery/job.py
```diff
@@ -3160,7 +3160,12 @@ def _begin(self, client=None, retry=DEFAULT_RETRY, timeout=None):
             raise
 
     def result(
-        self, page_size=None, max_results=None, retry=DEFAULT_RETRY, timeout=None
+        self,
+        start_index=None,
+        page_size=None,
+        max_results=None,
+        retry=DEFAULT_RETRY,
+        timeout=None,
```
plamut, line 3168: Even though the new parameter's position seems logical and appealing, it can break any existing user code that passes the values as positional arguments. It would thus be more safe to add `start_index` at the end.

PR #125, opened 2020-06-09
feat: added typing hints to query function in client.py

### LINKED ISSUE #124: Improve code typing for a better coding experience
I began working with Google's BigQuery product a week ago and since I use python, I installed this client library. Having used it to manage my data, run queries etc, I have realised that the library lacks proper typing for it's functions to allow IDEs to properly display the returned object's properties and methods.

I feel that adding typing hints to all the functions would help newbies like myself to find what they need and access the properties mentioned in the documentation for objects like QueryJob.

I'm raising this feature request to get feedback on whether this is a good idea.

Before:
Screenshot from 2020-06-09 18-40-12

After (Adding typing to query() in client.py):
Screenshot from 2020-06-09 18-40-29

Edit:
Let me know if this issue is only affecting me as I'm using the PyCharm IDE or whether this is affecting others.
I know the doc strings are present in all the functions, but I feel signature typing hints would help a great deal.

shenoy-anurag:
python typing for query function in client.py to allow IDEs to correctly show the properties of the query job object returned and save them time. I propose that the same typing hints should be added to all the functions to help developers and improve their experience of using google bigquery.

Please check the issue mentioned below and please comment there if you feel this is a bad idea.

Fixes #124

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2391,7 +2391,7 @@ class Client(ClientWithProject):
         project=None,
         retry=DEFAULT_RETRY,
         timeout=None,
-    ):
+    ) -> "job.QueryJob":
         """Run a SQL query.
 
         See


plamut:
@shenoy-anurag Thanks for the pull request, but unfortunately compatibility with Python 2.7 has not yet been dropped (despite what the docs said), thus this change would cause a `SyntaxError` under Python 2, breaking the tests.

PR #127, opened 2020-06-09
doc: update wording in rowiterator docstrings to reduce confusion

shollyman:
Talking about partition filtration can be problematic, so refer
to Storage API capabilities more obliquely.  Particularly, partition
filtration is possible via a query or a direct storage API read, so
don't use as severe of language when describing helper mechanisms such
as to_dataframe() on row iterators.

Fixes: https://github.com/googleapis/python-bigquery-storage/issues/22


shollyman:
I anticipate this may cause minor conflicts with https://github.com/googleapis/python-bigquery/pull/55 so let's get that in, then we can followup with smaller doc changes like this and cleanup for the small anonymous result issues (I didn't excise those adjacent language blocks).

PR #135, opened 2020-06-12
chore: restrict `rsa` version to <=4.0 for Python 2.7 compatibility

### LINKED ISSUE #134: A new release of rsa dependency breaks Python 2.7 tests
Recent `rsa` releases are not compatible with Python 2.7 anymore, the last compatible version is 4.0. We need to bound its version in order to preserve Python 2.7 compatibility.

> Major changes in 4.1
Version 4.0 was the last version to support Python 2 and 3.4. Version 4.1 is compatible with Python 3.5+ only.

plamut:
Fixes #134.

This PR assures that too new a version of the `rsa` dependency does not break the library's Python 2.7 compatibility.

That version is already two years old, though, and it represents a yet another reason to start dropping Python 2.7 (and 3.5) support, too.

### PR checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ dependencies = [
     "google-resumable-media >= 0.5.0, < 0.6dev",
     "protobuf >= 3.6.0",
     "six >=1.13.0,< 2.0.0dev",
+    "rsa <= 4.0",  # rsa 4.0 is the last release compatible with Python 2.7
 ]
 extras = {
     "bqstorage": [
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,10 @@ dependencies = [
     "google-resumable-media >= 0.5.0, < 0.6dev",
     "protobuf >= 3.6.0",
     "six >=1.13.0,< 2.0.0dev",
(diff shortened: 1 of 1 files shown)


plamut:
> AttributeError: module 'virtualenv.activation.python' has no attribute 'PestythonActivator'

I'm not sure what that is, happens when commands in virtualenvs are run, hopefully a restart will make the CI checks green again.

shollyman:
bad day for python releases.  looks like the new breakage here is in virtualenv

HemangChothani:
@plamut This issue has been handled in `google-auth` library as you can see [here](https://github.com/googleapis/google-auth-library-python/blob/master/setup.py#L24-L26), is this change still needed.

plamut:
@HemangChothani Thanks for the update!

In theory the change should not be needed, just a version bump of the `auth` library, but I'm not 100% sure `pip` resolves always determines the matching versions correctly (the resolver is currently being rewritten AFAIK).

I would thus play it safe and be explicit about this, but I will borrow the fix from `auth`, which only restricts the RSA version for Python 2, while keeping it up to date in Python 3.

PR #139, opened 2020-06-18
docs(bigquery): fix the broken docs

### LINKED ISSUE #138: [docs][1.25] Documentation on https://googleapis.dev/python/bigquery/1.25.0/ is broken
Hi!
The docs for 1.25 version is broken. At least for `Client` class, there is no methods documentation any more.
Just compare:
* [1.24](https://googleapis.dev/python/bigquery/1.24.0/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.query)
* [1.25](https://googleapis.dev/python/bigquery/1.25.0/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.query)

HemangChothani:
Fixes #138 

-> I know conf.py file generated by synthtool ,this is temporary patch for bigquery.

-> new release of sphinx 3.1.1 Consider duplicate docstring of  `Attributes:` docstring and `property` docstring and raised an error of duplicate description so change the docstring from  `Attributes:` to `Args` for more specific info refer this issue : https://github.com/sphinx-doc/sphinx/issues/7817

### DIFF
--- a/synth.py
+++ b/synth.py
@@ -61,4 +61,10 @@ s.replace("google/cloud/bigquery_v2/proto/*.py", "[“”]", '``')
 templated_files = common.py_library(cov_level=100)
 s.move(templated_files, excludes=["noxfile.py"])
 
+s.replace(
+    "docs/conf.py",
+    r'\{"members": True\}',
+    '{"members": True, "inherited-members": True}'
+)
+
 s.shell.run(["nox", "-s", "blacken"], hide_output=False)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -43,7 +43,7 @@ extensions = [
 
 # autodoc/autosummary flags
 autoclass_content = "both"
-autodoc_default_flags = ["members", "inherited-members"]
+autodoc_default_options = {"members": True, "inherited-members": True}
 autosummary_generate = True
 
 
--- a/google/cloud/bigquery/dataset.py
+++ b/google/cloud/bigquery/dataset.py
@@ -84,7 +84,7 @@ class AccessEntry(object):
 
     See https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets.
 
-    Attributes:
+    Args:
         role (str):
             Role granted to the entity. The following string values are
             supported: `'READER'`, `'WRITER'`, `'OWNER'`. It may also be


HemangChothani:
Opened a PR in `synthtool` to update `conf.py` file template https://github.com/googleapis/synthtool/pull/629

HemangChothani:
@plamut The PR of synthtool https://github.com/googleapis/synthtool/pull/629 has been merged so i will remove the doc/conf.py change from here, but i don't have an idea about change in synth.py file.

plamut:
@HemangChothani I presume we still need to add the `"inherited-members": True` option to the docs config, because the synth template contains just the `"members": True` option? Re-generating the files with `synthtool` erases that latter part, thus we need to add it manually (through a replacement configured in `synth.py`)

Or do you think we could actually make that change in the `docs/conf.py` template itself?
```
autodoc_default_options = {"members": True, "inherited-members": True}
```

HemangChothani:
@plamut  
In synth.py file can we do that? I am not sure right now as i am struggling with synth config to local machine. 

s.replace(
    "docs/conf.py",
    {"members": True},
    {"members": True, "inherited-members": True}
)

I am not sure that it will work for all clients or not , i will test it.

plamut:
@HemangChothani Yes, each library can add custom modifications to the generated files in its `synth.py`.

The snippet above is the right direction, just mind that the second argument is treated as a regular expression, thus regex special characters need to be escaped.

If you have trouble setting up synth locally, I can post a patch here myself to save time, as I already happen to have everything working locally.

HemangChothani:
@plamut Thank you for the guidance and support, i am able to run locally an test it and push the code, could you please verify it?

synth.py
```diff
@@ -61,4 +61,10 @@
 templated_files = common.py_library(cov_level=100)
 s.move(templated_files, excludes=["noxfile.py"])
 
+s.replace(
+    "docs/conf.py",
+    '{"members": True}',
```
plamut, line 66:
```suggestion
    r'\{"members": True\}',
```
The second argument is a regex, special characters need escaping.

noxfile.py
```diff
@@ -129,7 +129,7 @@ def cover(session):
     session.run("coverage", "erase")
 
 
-@nox.session(python="3.8")
+@nox.session(python="3.7")
```
plamut, line 132: Just curious, is this necessary?

HemangChothani: Nop, sorry that's my bad.

synth.py
plamut: Hmm... it appears that it works as is and the escape might not be needed, interesting.

Or I might have messed up something locally when testing the PR again...

HemangChothani: Yup, both works fine for me so i have changed to regex as you recommend.

plamut: Sounds safer, yes.

I can confirm that the replacement works, please just commit the changes in the `docs/conf.py` file, too, otherwise the PR on its own will have no effect on the docs. :wink:

PR #150, opened 2020-06-26
feat(bigquery): expose date_as_object parameter to users

### LINKED ISSUE #11: BigQuery: to_dataframe does not respect date fields
#### Environment details

1. google-cloud-bigquery: version 1.22
2. OS type and version: Centos 7
3. Python version: 3.6.9

#### Steps to reproduce

  1. Download a table from BigQuery to a pandas Dataframe

#### Code example

```python
df = bq_client.list_rows(table).to_dataframe()
df.types

contractId                       int64
contractTypeId                   int64
affiliateId                      int64
invoicePeriodId                  int64
startDate                       **object**
endDate                         **object**
contractName                    object
details                         object
updated            datetime64[ns, UTC]
created            datetime64[ns, UTC]
dtype: object
```
Timestamps seem to work fine, but **date** is just being treated as a string.

Table schema is defined as:
```python
'contract' : [
            bigquery.SchemaField("contractId", "INTEGER", mode="REQUIRED"),
            bigquery.SchemaField("contractTypeId", "INTEGER", mode="REQUIRED"),
            bigquery.SchemaField("affiliateId", "INTEGER", mode="REQUIRED"),
            bigquery.SchemaField("invoicePeriodId", "INTEGER", mode="REQUIRED"),
            bigquery.SchemaField("startDate", "DATE", mode="REQUIRED"),
            bigquery.SchemaField("endDate", "DATE", mode="NULLABLE"),
            bigquery.SchemaField("contractName", "STRING", mode="NULLABLE"),
            bigquery.SchemaField("details", "STRING", mode="NULLABLE"),
            bigquery.SchemaField("created", "TIMESTAMP", mode="REQUIRED"),
            bigquery.SchemaField("updated", "TIMESTAMP", mode="REQUIRED"),
        ]
```
This should have anything to do with pagination, because the _startDate_ field is required; also the table only has 20 records.

HemangChothani:
Fixes #11

### DIFF
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -3351,7 +3351,7 @@ class QueryJob(_AsyncJob):
                 for details.
 
                 ..versionadded:: 1.11.0
-            create_bqstorage_client Optional[bool]:
+            create_bqstorage_client (Optional[bool]):
                 If ``True`` (default), create a BigQuery Storage APIclient
                 using the default API settings. The BigQuery Storage API
                 is a faster way to fetch rows from BigQuery. See the
@@ -3361,7 +3361,7 @@ class QueryJob(_AsyncJob):
 
                 ..versionadded:: 1.24.0
 
-            date_as_object Optional[bool]:
+            date_as_object (Optional[bool]):
                 If ``True`` (default), cast dates to objects. If ``False``, convert
                 to datetime64[ns] dtype.
 
@@ -3437,7 +3437,7 @@ class QueryPlanEntry(object):
             resource(Dict[str: object]):
                 ExplainQueryStage representation returned from API.
 
-        Returns:
+        Returns:bigquery_issue_11
             google.cloud.bigquery.QueryPlanEntry:
                 Query plan entry parsed from ``resource``.
         """
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -3351,9 +3351,9 @@ class QueryJob(_AsyncJob):
                 for details.
 
                 ..versionadded:: 1.11.0
-            create_bqstorage_client (bool):
-                Optional. If ``True`` (default), create a BigQuery Storage API
-                client using the default API settings. The BigQuery Storage API
+            create_bqstorage_client Optional[bool]:
+                If ``True`` (default), create a BigQuery Storage APIclient
+                using the default API settings. The BigQuery Storage API
                 is a faster way to fetch rows from BigQuery. See the
                 ``bqstorage_client`` parameter for more information.
 
@@ -3361,9 +3361,9 @@ class QueryJob(_AsyncJob):
 
                 ..versionadded:: 1.24.0
 
-            date_as_object (bool):
-                Optional. If ``True`` (default), cast dates to objects.
-                If False, convert to datetime64[ns] dtype.
+            date_as_object Optional[bool]:
+                If ``True`` (default), cast dates to objects. If ``False``, convert
+                to datetime64[ns] dtype.
 
                 ..versionadded:: 1.26.0
 
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -3320,6 +3320,7 @@ class QueryJob(_AsyncJob):
         dtypes=None,
         progress_bar_type=None,
         create_bqstorage_client=True,
+        date_as_object=True,
     ):
         """Return a pandas DataFrame from a QueryJob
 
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/job.py
```diff
@@ -3360,6 +3361,12 @@ def to_dataframe(
 
                 ..versionadded:: 1.24.0
 
+            date_as_object (bool):
```
plamut, line 3364: (nit) Should be `Optional[bool]` for consistency. And while at it, we can also fix the same for the `create_bqstorage_client` parameter.

The same for the parameter added in `table.py`.

google/cloud/bigquery/job.py
```diff
@@ -3360,6 +3361,12 @@ def to_dataframe(
...
+                Optional. If ``True`` (default), cast dates to objects.
+                If False, convert to datetime64[ns] dtype.
```
plamut, line 3366: (nit) missing backticks around `False` (in both comments)

tests/unit/test_job.py
```diff
@@ -5533,6 +5541,37 @@ def test_to_dataframe_column_dtypes(self):
...
+        job = self._make_one(self.JOB_ID, self.QUERY, client)
+
+        df = job.to_dataframe(date_as_object=False, create_bqstorage_client=False)
+
+        self.assertIsInstance(df, pandas.DataFrame)
+        self.assertEqual(len(df), 1)  # verify the number of rows
+        exp_columns = [field["name"] for field in query_resource["schema"]["fields"]]
+        self.assertEqual(list(df), exp_columns)  # verify the column names
+
+        self.assertEqual(df.date.dtype.name, "datetime64[ns]")
```
plamut, line 5573: If I read the code correctly, the new `date_as_object` parameter is only applied if `pyarrow` is installed?

I would thus skip this test if `pyarrow` is not installed, and also add another one checking that `date_as_object` works as expected even if `pyarrow` is _not_ installed.

google/cloud/bigquery/table.py
```diff
@@ -1799,6 +1806,7 @@ def to_dataframe(
         dtypes=None,
         progress_bar_type=None,
         create_bqstorage_client=True,
+        date_as_object=True,
```
plamut, line 1809: Ah, sorry for initially missing that - let's add a docstring for the new parameter here, too.

google/cloud/bigquery/job.py
HemangChothani: Ok, I think i need to take a separate cleanup task to make use of `optional` keyword consistent as the same method has different ways to use of  `optional` keyword.

google/cloud/bigquery/table.py
HemangChothani: I think it's there https://github.com/googleapis/python-bigquery/pull/150/files/7abb6e0ec64c55023df6fbb0809abe1f8cf6ec3e#diff-f0febbf80214497b41d3dba7fe3b8d37R1687-R1692

google/cloud/bigquery/job.py
plamut: If you want to do this holistically across the class/module/library, that's totally fine, we don't have to bloat this PR.

google/cloud/bigquery/table.py
plamut: That's a different one. :)

This particular `to_dataframe()` method is in the `_EmptyRowIterator` class and the new parameter does not seem to be documented a few liens below.

tests/unit/test_job.py
HemangChothani: Yes, it's only applied if `pyarrow` installed, without pyarrow it gives date as object.

google/cloud/bigquery/table.py
```diff
@@ -1673,16 +1674,22 @@ def to_dataframe(
...
+                If ``True`` (default), create a BigQuery Storage API client
+                using the default API settings. The BigQuery Storage API
                 is a faster way to fetch rows from BigQuery. See the
                 ``bqstorage_client`` parameter for more information.
 
                 This argument does nothing if ``bqstorage_client`` is supplied.
 
                 ..versionadded:: 1.24.0
 
+            date_as_object Optional[bool]:
```
plamut, line 1687: Parameter type should be enclosed in parentheses, otherwise it confuses the parser and it ends up looking weird in the docs:

Screenshot from 2020-06-29 08-33-40


Applies to multiple places, please check.

google/cloud/bigquery/job.py
```diff
@@ -3437,7 +3437,7 @@ def from_api_repr(cls, resource):
             resource(Dict[str: object]):
                 ExplainQueryStage representation returned from API.
 
-        Returns:
+        Returns:bigquery_issue_11
```
plamut, line 3440: A leftover? :)

google/cloud/bigquery/job.py
```diff
@@ -3350,16 +3351,22 @@ def to_dataframe(
                 for details.
 
                 ..versionadded:: 1.11.0
-            create_bqstorage_client (bool):
-                Optional. If ``True`` (default), create a BigQuery Storage API
-                client using the default API settings. The BigQuery Storage API
+            create_bqstorage_client (Optional[bool]):
+                If ``True`` (default), create a BigQuery Storage APIclient
```
plamut, line 3355:
```suggestion
                If ``True`` (default), create a BigQuery Storage API client
```

PR #162, opened 2020-07-10
feat(bigquery): add __eq__ method for class PartitionRange and RangePartitioning

### LINKED ISSUE #159: Comparaison between RangePartitioning objects seems to be incorrect.
Comparaison between RangePartitioning objects seems to be incorrect.

#### Environment details

  - `google-cloud-bigquery` version: `1.25.0`

#### Code example

```python
from google.cloud.bigquery import RangePartitioning

print(RangePartitioning(field='something') ==
      RangePartitioning(field='something'))
```
Code above returns `False`. I was expecting `True`.

HemangChothani:
Fixes #159


google/cloud/bigquery/table.py
```diff
@@ -1891,10 +1891,21 @@ def interval(self, value):
...
+
+    def __ne__(self, other):
+        return not self == other
+
     def __repr__(self):
         key_vals = ["{}={}".format(key, val) for key, val in self._key()]
         return "PartitionRange({})".format(", ".join(key_vals))
 
+    def __hash__(self):
+        return hash(self._key())
```
plamut, line 1907: A hash should never change during the object's lifetime, as it can lead to unexpected results. Mutable objects should thus not be hashable:
```py
>>> from google.cloud.bigquery import PartitionRange
>>> pr = PartitionRange(start=42)
>>> s = {pr}
>>> pr in s
True
>>> pr.start = 43
>>> s
{PartitionRange(start=43)}
>>> pr in s
False   # huh??
```
Until Python 2 compatibility needs to be preserved, we need to manually mark the object as unhashable:
```py
__hash__ = None
```

google/cloud/bigquery/table.py
```diff
@@ -1961,10 +1972,21 @@ def field(self, value):
...
+
+    def __ne__(self, other):
+        return not self == other
+
     def __repr__(self):
         key_vals = ["{}={}".format(key, repr(val)) for key, val in self._key()]
         return "RangePartitioning({})".format(", ".join(key_vals))
 
+    def __hash__(self):
+        return hash(self._key())
```
plamut, line 1988:
```suggestion
    __hash__ = None
```
Same as above.

google/cloud/bigquery/table.py
HemangChothani: I have tried same with the `table.TimePartitioning` and `dataset.EncryptionConfiguration` classes and got the output same as you had mentioned above, any suggestion on that?

plamut: The "same output" meaning the same issue when using the instances in a hashable collection?

If an object with a custom `eq` is mutable, it really shouldn't be hashable, unless the custom equality and hash are based on some immutable subset of the attributes. If we want hashability, we need to make the instances immutable. If we nevertheless still want to "edit" such instances, we can add methods that make a _copy_ of the instance containing the changes (IIRC the Retry objects from API core are an example of that).

It might not be worth it, though, depending on how much we really need hashability.

tests/unit/test_table.py
```diff
@@ -3550,19 +3550,10 @@ def test__eq___type_mismatch(self):
...
-        set_two = {object_under_test1, object_under_test2}
-        self.assertEqual(set_one, set_two)
 
-    def test___hash__not_equals(self):
-        object_under_test1 = self._make_one(start=1, end=10, interval=2)
-        object_under_test2 = self._make_one(start=1, end=10, interval=3)
-        set_one = {object_under_test1}
-        set_two = {object_under_test2}
-        self.assertNotEqual(set_one, set_two)
+        self.assertIsNone(object_under_test1.__hash__)
```
plamut, line 3556: I would make this assertion more explicit, i.e. checking that hashing an object produces an error. Something like the following:
```py
with six.assertRaisesRegex(self, TypeError, r".*unhashable type.*"):
    hash(object_under_test1)
```

francois-baptiste:
Great! Thank you @HemangChothani

PR #163, opened 2020-07-11
fix: raise error if inserting rows with unknown fields

### LINKED ISSUE #151: client.insert_rows should fail when inserting non-existing fields
Version
```
google-cloud-bigquery==1.25.0
```

The [`client.insert_rows()`](https://github.com/googleapis/python-bigquery/blob/v1.25.0/google/cloud/bigquery/client.py#L2471) function doesn't fail when inserting non-existing fields whereas the BigQuery API does fail with a message like
```
{
  "kind": "bigquery#tableDataInsertAllResponse",
  "insertErrors": [
    {
      "index": 0,
      "errors": [
        {
          "reason": "invalid",
          "location": "zap",
          "debugInfo": "",
          "message": "no such field."
        }
      ]
    }
  ]
}
```
`insert_rows()` silently drops the additional columns instead.

This happens because `insert_rows()` uses [`_record_field_to_json`](https://github.com/googleapis/python-bigquery/blob/3869e34f3eff6fabca21c397e3f1cbc368ec880e/google/cloud/bigquery/_helpers.py#L409) which only iterates over the list of fields that are provided ignoring all the other fields that are part of the data and `insert_rows()` passes the table's schema as list of fields to `_record_field_to_json`.

This behavior is opposite to the BigQuery API and means we cannot reliably insert data because we're not made aware of changes to the incoming data because there's no failure.
IMHO this behavior is not correct, I think it would be OK if `selected_fields` was provided but it should not silently use the schema to limit which of the fields of the input data are processed and ignore the rest.
I can image there might be cases where one wants to be lenient/ignore all fields that are not part of table so this behavior might have to be an option, possibly combined with `selected_fields`.

P.S. By extension this also applies to the [`client.insert_rows_from_dataframe()`](https://github.com/googleapis/python-bigquery/blob/v1.25.0/google/cloud/bigquery/client.py#L2533) function which uses `client.insert_rows()`.

P.P.S We initially ran into this when using `insert_rows_from_dataframe()` and it was a bit of a search to find where this was going wrong because it's a somewhat indirect chain of `insert_rows_from_dataframe -> insert_rows -> insert_rows_json`.
Why was this long way chosen/added instead of simply using `insert_rows_json(table, df.to_dict(orient="records"))`? It seems a lot simpler and will probably be the workaround we'll implement for now.

plamut:
Fixes #151.

This PR makes sure that unknown fields are not silently ignored when inserting rows, but instead such errors are detected and raised loudly.

### PR checklist


google/cloud/bigquery/_helpers.py
```diff
@@ -419,9 +419,23 @@ def _record_field_to_json(fields, row_value):
     Returns:
         Mapping[str, Any]: A JSON-serializable dictionary.
     """
-    record = {}
     isdict = isinstance(row_value, dict)
```
tswast, line 422: If it's already a dictionary, is there any need for additional processing?

google/cloud/bigquery/_helpers.py
```diff
@@ -430,6 +444,19 @@ def _record_field_to_json(fields, row_value):
...
             record[subname] = _field_to_json(subfield, subvalue)
 
+        if isdict:
+            processed_fields.add(subname)
+
+    # Unknown fields should not be silently dropped.
+    if isdict:
+        not_processed = set(row_value.keys()) - processed_fields
+        if not_processed:
+            raise ValueError(
```
tswast, line 454: Seems to me as though we should prefer keeping any unknown fields in the dictionary and let the API return an error.

google/cloud/bigquery/_helpers.py
plamut: We do, because we need to explicitly exclude any `None` values from the subsequent request ([commit](https://github.com/googleapis/python-bigquery/commit/60926a2e954b4bc2f5796b5ade339342c574b481)).

google/cloud/bigquery/_helpers.py
plamut: I was actually considering this, but then opted for more consistency with `tuple` rows, which also raise the error immediately if a length mismatch is detected.

On the other hand, though, we cannot detect all bad cases, and we definitely don't want to replicate the server-side logic in the client. I'll change it to include the unknown fields and let the backend complain.

plamut:
@shollyman Hmm, good point. I'll check how the helper can support the `ignore_unknown_values` parameter.

plamut:
Changed the helper to include missing fields and let the backend handle that.

The `ignore_unknown_values` argument to user-facing methods in the client is orthogonal to that, thus the helper signature didn't need a change (FWIW, the backend honors the argument, checked it).

plamut:
Just the coverage failure it seems, will cover the missed code path.

tseaver:
Assuming that  the [Python3 test failures in `test_download_arrow_tabledata_list_unknown_field_type`](https://source.cloud.google.com/results/invocations/7a942794-e7a4-4513-97ac-06a281367307/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fpresubmit/log) due to a pyarrow release, I've kicked off a merge.

plamut:
Indeed, and those tests were updated in https://github.com/googleapis/python-bigquery/pull/192/files

PR #164, opened 2020-07-13
chore(deps): remove redundant dependencies

### LINKED ISSUE #158: Remove redundant dependencies
I think we can remove the following from the dependencies list:

https://github.com/googleapis/python-bigquery/blob/dbaf3bdc35656fdfef2d4380befdcc2392c1ca83/setup.py#L31-L43

* `google-auth` is already required via [google-api-core](https://github.com/googleapis/python-api-core/blob/master/setup.py#L34)
* `protobuf` is already required via [google-api-core](https://github.com/googleapis/python-api-core/blob/master/setup.py#L33)
* `rsa` [google-auth](https://github.com/googleapis/google-auth-library-python/blob/master/setup.py#L25-L26) is pinning compatible versions now

HemangChothani:
Fixes #158

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ version = "1.25.0"
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
     'enum34; python_version < "3.4"',
-    "google-api-core >= 1.15.0, < 2.0dev",
+    "google-api-core >= 1.21.0, < 2.0dev",
     "google-cloud-core >= 1.1.0, < 2.0dev",
     "google-resumable-media >= 0.5.0, < 0.6dev",
     "six >=1.13.0,< 2.0.0dev",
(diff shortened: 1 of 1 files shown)


plamut:
I believe we also need to bump the `api-core` version to benefit from the most recent version pins in it?

For example, the `rsa` pins have been [tightened](https://github.com/googleapis/google-auth-library-python/commit/46677a0cb3bde6622be10061bc61daaff7a0aaca#diff-2eeaed663bd0d25b7e608891384b7298) in `google-auth` in version `1.18.0`, but `google-auth` was [bumped](https://github.com/googleapis/python-api-core/commit/ded92d0acdcde4295d0e5df05fda0d83783a3991#diff-2eeaed663bd0d25b7e608891384b7298R34) only in `api-core` version `1.21.0`.

(we currently pin `api-core` to  `>= 1.15.0`)

tswast:
Ummm, protobuf was not redundant. It's needed for the BQML files in the `google.cloud.bigquery_v2` package.

PR #148, opened 2020-06-25
docs: add samples from bigquery/cloud-client

plamut:
Samples from https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigquery/cloud-client

---
> **IMPORTANT:** When merging, use the **rebase and merge** option to preserve the samples commit history!
---

**NOTE:** There are also other [bigquery-related samples](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigquery) in the samples repo. If any others besides `cloud-client` need to be moved here, please let me know and I'll open additional PRs.

Closes #149.
Towards #147.

Issue #147, opened 2020-06-25
Transition the library to the new microgenerator


plamut:
With the new code generator ready to be rolled out, we can make the transition here in PubSub. **This implies dropping support for Pythom 2.7 and 3.5!**

plamut:
Closed by #278

PR #174, opened 2020-07-21
docs: add samples from bigquery/cloud-client

### LINKED ISSUE #149: Add bigquery/cloud-client samples from the samples repo
A transition to the new code generator (#147) requires the code samples to be moved from the common samples repo to here.

plamut:
Samples from https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigquery/cloud-client

---
> **IMPORTANT:** When merging, use the **rebase and merge** option to preserve the samples commit history!
---

**NOTE:** There are also other [bigquery-related samples](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigquery) in the samples repo. If any others besides `cloud-client` need to be moved here, please let me know and I'll open additional PRs.

Closes #149.
Towards #147.
Supersedes #148 (regenerates latest samples from scratch).


plamut:
I would need some CI insight here, I can't see the samples log to see why those checks failed. Are the failures the same as in the "samples" session of the main noxfile?

Importing `IPython` fails,  even though it was added to samples dependencies and samples tests pass locally (using the samples/snippets noxfile). Ditto for the missing `GOOGLE_CLOUD_PROJECT` env variable.

Also, should the `snippets` session new be removed from the main noxfile?

busunkim96:
CI seems to be falling over during the setup stage. This usually happens when the path to the build script is misconfigured. 

```
[08:05:42] FATAL: Build script missing in checked out sources
[08:05:42] Failed to start execution
[08:05:42] hudson.AbortException: Build script does not exist: /tmp/workspace/workspace/cloud-devrel/client-libraries/python/googleapis/python-bigquery/samples/python3.6/presubmit/src/github/python-bigquery/.kokoro/trampoline.sh
```

Will submit a PR to fix this, once I've figured out which config it is 😄

busunkim96:
@plamut Once all the samples are being tested via a samples noxfile you can delete the extra session in the main noxfile.

That said, I see there are existing samples in the base [`samples`](https://github.com/googleapis/python-bigquery/tree/master/samples) directory. Is it okay to move those into the `samples/snippets` directory so they get tested? Should they be grouped into some other directory? @shollyman

shollyman:
Yeah, we'll need to move the existing samples.  Happy to do this in a subsequent PR, since it also means we need to update all the include paths in the documentation site.

plamut:
It turned out that tests in `samples/snippets` were run by both the `snippets` session in the main noxfile and the `py` session in the local noxfile placed in that directory. Some of those tests failed, because the main session config sets up a slightly different environment.

A clean solution was to simply exclude `samples/snippets` from the main snippets run and moving the existing samples turned out to not even be necessary (at least for this particular PR).

noxfile.py
```diff
@@ -113,8 +113,10 @@ def snippets(session):
     session.install("-e", ".[all]")
 
     # Run py.test against the snippets tests.
+    # Skip tests in samples/snippets, as those are run in a different session
+    # using the nox config in that directory.
```
plamut, line 117:
```suggestion
    # using the nox config from that directory.
```

plamut:
Meh, rebase and merge is not allowed due to conflicts. I will have to do it locally and re-push.

plamut:
@busunkim96 @shollyman Please just flip the CLA label again, and I think we can merge this, yay. Thanks!

PR #190, opened 2020-07-24
chore: constrain upper bound on pyarrow dependency

shollyman:
It appears pyarrow just released 1.0.0 and we'll need to do more work to
address changes.  This PR sets an upper bound on this exclude the new
1.0 series for now.


plamut:
This might not be needed anymore, as all tests seem to pass in #192.

shollyman:
Dropping this.

Issue #191, opened 2020-07-24
improve test latency


shollyman:
recent runs of presubmit kokoro show we're taking something on the order of 37-48 minutes for typical runs, with a couple invocations significantly higher than that.  It's a significant impediment to actually getting things done.

We can definitely test much faster than this.

Some things to look at:

* Decompose the monolilthic kokoro job that runs all the nox sessions serially into parallel invocations, the same way we decompose the samples invocations.  Fan out to get better wall time.

* Look for obvious slow tests and improve them, reducing the tail of the (to be) parallelized test runs.  Based on logging, system tests are unsuprisingly the longer session typically. 

* Look at coverage overlap between samples and library; we may be able to reduce redundancy in testing.

Here's some quick info from running locally on 3.8:

```
$ nox -s system-3.8 -- --durations=0

44.84s call     tests/system.py::TestBigQuery::test_dbapi_w_standard_sql_types
29.28s call     tests/system.py::TestBigQuery::test_dbapi_w_query_parameters
27.39s call     tests/system.py::TestBigQuery::test_query_many_columns
15.91s call     tests/system.py::TestBigQuery::test_query_w_query_params
15.04s call     tests/system.py::TestBigQuery::test_load_table_from_file_w_explicit_location
13.98s call     tests/system.py::TestBigQuery::test_load_table_from_dataframe_w_automatic_schema
13.79s call     tests/system.py::TestBigQuery::test_query_w_standard_sql_types
13.19s call     tests/system.py::TestBigQuery::test_load_table_from_dataframe_w_nullable_int64_datatype
11.90s call     tests/system.py::TestBigQuery::test_load_table_from_local_avro_file_then_dump_table
11.19s call     tests/system.py::TestBigQuery::test_load_table_from_dataframe_w_required
10.83s call     tests/system.py::TestBigQuery::test_nested_table_to_dataframe
10.50s call     tests/system.py::TestBigQuery::test_load_table_from_dataframe_w_nulls
10.30s call     tests/system.py::TestBigQuery::test_load_table_from_json_schema_autodetect
10.11s call     tests/system.py::TestBigQuery::test_copy_table
9.60s call     tests/system.py::TestBigQuery::test_load_table_from_dataframe_w_explicit_schema
8.56s call     tests/system.py::TestBigQuery::test_dbapi_fetchall
8.33s call     tests/system.py::TestBigQuery::test_extract_table
[...]

PR #202, opened 2020-07-30
tests: skip snippets / samples in main Kokoro build

tseaver:
Toward #191.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -29,6 +29,8 @@ export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
 # Setup project id.
 export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 
+export SESSIONS=$(nox --list | grep "^\*" | cut -d " " -f 2 | grep -v snippets)
+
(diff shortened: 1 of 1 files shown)


.kokoro/build.sh
```diff
@@ -29,11 +29,13 @@ export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
 # Setup project id.
 export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 
+export SESSIONS=$(nox --list | grep "^\*" | cut -d " " -f 2 | grep -v snippets)
```
plamut, line 32: Neat!

plamut:
@tseaver The changes here will [conflict](https://github.com/googleapis/python-bigquery/pull/207/files#diff-ce0d2ea7b6085e6c4bd286d0a15a3971) with the PR that will split system tests. Maybe coordinate with @tmatsuo to include skipping the snippets tests in that other PR or to unify the env variable names?

tseaver:
@plamut I think my work here is premature, because we need to get the various snippet files in `samples/` (e.g., `samples/add_empty_column.py`) tested by the external `snippets-x.y` job before dropping them in the main build.

PR #207, opened 2020-07-31
testing: split system tests

tmatsuo:
I manually run synthtool with the changes in https://github.com/googleapis/synthtool/pull/712

Motivation: #191 

Note that noxfile.py is skipped, so I manually updated it (probably ok)


tmatsuo:
@tseaver @shollyman @busunkim96 I can not assign a reviewer or add a label. Can you review?

tmatsuo:
I checked the test logs and they work as expected:

* main presubmit build is skipping system tests
* system-x.x builds are only running a system test for a single python version
* presubmit is the slowest. It took 24min 40secs.

Another observation is that the snippets tests are taking long time. Once we move them out, the build will be much faster.

noxfile.py
```diff
@@ -190,3 +194,36 @@ def docs(session):
...
+            "sphinx.ext.todo,"
+            "sphinx.ext.viewcode,"
+            "recommonmark"
+        ),
+        "-b",
+        "html",
+        "-d",
+        os.path.join("docs", "_build", "doctrees", ""),
+        os.path.join("docs", ""),
+        os.path.join("docs", "_build", "html", ""),
```
tseaver, line 228: Is this supposed to build into the same output directory as the main docs build?

synth.py
```diff
@@ -59,7 +59,7 @@
 # ----------------------------------------------------------------------------
 # Add templated files
 # ----------------------------------------------------------------------------
-templated_files = common.py_library(cov_level=100, samples=True)
+templated_files = common.py_library(cov_level=100, samples=True, split_system_tests=True)
 
 # BigQuery has a custom multiprocessing note
 s.move(templated_files, excludes=["noxfile.py", "docs/multiprocessing.rst"])
```
tseaver, line 65: I think we need to exclude `docs/conf.py` as well, to prevent future synth failures.

docs/conf.py
```diff
@@ -20,6 +20,9 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath(".."))
 
+# For plugins that can not read conf.py.
+sys.path.insert(0, os.path.abspath("."))
```
tseaver, line 24: Is this a workaround for a `sphinx-docfx-yaml` bug?  Should we link that issue here?

noxfile.py
tmatsuo: For docfx build, we only upload yaml files, so I don't think it matters.

synth.py
tmatsuo: I can do that, but I think we should rather fix this by modifying the template to include the excludes. WDYT?

docs/conf.py
tmatsuo: I will add it in the template.

tmatsuo:
@tseaver Thanks, I opened https://github.com/googleapis/synthtool/pull/713 which will allow us to use the template.

synth.py
tmatsuo: I updated the template in synthtool and re-generated the files. Now we don't manually modify the files.

tmatsuo:
@tseaver Tests are passing. I rebased to the master. PTAL

PR #215, opened 2020-08-09
Opentelemetry tracing

aravinsiva:
PR is to add Opentelemetry instrumentations to all api call's made by client.py and job.py modules.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -102,3 +102,51 @@ Perform a query
 
     for row in rows:
         print(row.name)
+
+Instrumenting With OpenTelemetry
+--------------------------------
+
+This application uses OpenTelmetry to output tracing data from
+api calls to BigQuery. To enable OpenTelemetry tracing in
+the BigQuery client the following PyPi packages need to be installed:
+
+.. code-block:: console
+
+    pip install opentelemetry-api
+
+.. code-block:: console
+
+    pip install opentelemetry-sdk
+
+.. code-block:: console
+
+    pip install opentelemetry-instrumentaion
+
+.. code-block:: console
+
+    pip install opentelemetry-exporter-cloud_trace
+
+After installation, OpenTelemetry can now be used in the BigQuery
+client and in BigQuery jobs. First however an exporter must be
+specified for where the trace data will be outputted to. An example of this
+can be found here:
+
+.. code-block:: python
+
+    from opentelemetry import trace
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.export import SimpleExportSpanProcessor
+    from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+    trace.set_tracer_provider(TracerProvider())
+    trace.get_tracer_provider().add_span_processor(
+        SimpleExportSpanProcessor(CloudTraceSpanExporter())
+    )
+
+In this example all tracing data will be published to the Google
+Cloud trace console. For more information on OpenTelemetry, please consult the `OpenTelemetry documentation`_.
+
+.. _OpenTelemetry documentation: https://opentelemetry-python.readthedocs.io
+
+
+
+
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -80,7 +80,6 @@ from google.cloud.bigquery.table import TableListItem
 from google.cloud.bigquery.table import TableReference
 from google.cloud.bigquery.table import RowIterator
 
-
 _DEFAULT_CHUNKSIZE = 1048576  # 1024 * 1024 B = 1 MB
 _MAX_MULTIPART_SIZE = 5 * 1024 * 1024
 _DEFAULT_NUM_RETRIES = 6
@@ -168,14 +167,14 @@ class Client(ClientWithProject):
     """The scopes required for authenticating as a BigQuery consumer."""
 
     def __init__(
-        self,
-        project=None,
-        credentials=None,
-        _http=None,
-        location=None,
-        default_query_job_config=None,
-        client_info=None,
-        client_options=None,
+            self,
+            project=None,
+            credentials=None,
+            _http=None,
+            location=None,
+            default_query_job_config=None,
+            client_info=None,
+            client_options=None,
     ):
         super(Client, self).__init__(
             project=project, credentials=credentials, _http=_http
@@ -213,7 +212,7 @@ class Client(ClientWithProject):
         self._http.close()
 
     def get_service_account_email(
-        self, project=None, retry=DEFAULT_RETRY, timeout=None
+            self, project=None, retry=DEFAULT_RETRY, timeout=None
     ):
         """Get the email address of the project's BigQuery service account
 
@@ -244,12 +243,14 @@ class Client(ClientWithProject):
         if project is None:
             project = self.project
         path = "/projects/%s/serviceAccount" % (project,)
-
-        api_response = self._call_api(retry, method="GET", path=path, timeout=timeout)
+        span_creator = SpanCreator()
+        attributes = {'path': path}
+        with span_creator.create('BigQuery.getServiceAccountEmail', attributes):
+            api_response = self._call_api(retry, method="GET", path=path, timeout=timeout)
         return api_response["email"]
 
     def list_projects(
-        self, max_results=None, page_token=None, retry=DEFAULT_RETRY, timeout=None
+            self, max_results=None, page_token=None, retry=DEFAULT_RETRY, timeout=None
     ):
         """List projects for the project associated with this client.
 
@@ -279,10 +280,15 @@ class Client(ClientWithProject):
                 Iterator of :class:`~google.cloud.bigquery.client.Project`
                 accessible to the current client.
         """
+        path = '/projects'
+        span_creator = SpanCreator()
+        attributes = {'path': path}
+        with span_creator.create('BigQuery.list_projects', attributes):
+            api_request = functools.partial(self._call_api, retry, timeout=timeout)
         return page_iterator.HTTPIterator(
             client=self,
-            api_request=functools.partial(self._call_api, retry, timeout=timeout),
-            path="/projects",
+            api_request=api_request,
+            path=path,
             item_to_value=_item_to_project,
             items_key="projects",
             page_token=page_token,
@@ -290,14 +296,14 @@ class Client(ClientWithProject):
         )
 
     def list_datasets(
-        self,
-        project=None,
-        include_all=False,
-        filter=None,
-        max_results=None,
-        page_token=None,
-        retry=DEFAULT_RETRY,
-        timeout=None,
+            self,
+            project=None,
+            include_all=False,
+            filter=None,
+            max_results=None,
+            page_token=None,
+            retry=DEFAULT_RETRY,
+            timeout=None,
     ):
         """List datasets for the project associated with this client.
 
@@ -343,9 +349,15 @@ class Client(ClientWithProject):
             # and converting it into a string here.
             extra_params["filter"] = filter
         path = "/projects/%s/datasets" % (project,)
+        span_creator = SpanCreator()
+        attributes = extra_params
+        attributes['path'] = path
+        attributes['page_token'] = page_token
+        with span_creator.create('BigQuery.list_datasets', attributes):
+            api_request = functools.partial(self._call_api, retry, timeout=timeout)
         return page_iterator.HTTPIterator(
             client=self,
-            api_request=functools.partial(self._call_api, retry, timeout=timeout),
+            api_request=api_request,
             path=path,
             item_to_value=_item_to_dataset,
             items_key="datasets",
@@ -414,7 +426,7 @@ class Client(ClientWithProject):
         return bigquery_storage_v1.BigQueryReadClient(credentials=self._credentials)
 
     def create_dataset(
-        self, dataset, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
+            self, dataset, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
     ):
         """API call: create the dataset via a POST request.
 
@@ -469,9 +481,13 @@ class Client(ClientWithProject):
             data["location"] = self.location
 
         try:
-            api_response = self._call_api(
-                retry, method="POST", path=path, data=data, timeout=timeout
-            )
+            span_creator = SpanCreator()
+            attributes = data
+            attributes['path'] = path
+            with span_creator.create('BigQuery.createDataset', attributes):
+                api_response = self._call_api(
+                    retry, method="POST", path=path, data=data, timeout=timeout
+                )
             return Dataset.from_api_repr(api_response)
         except google.api_core.exceptions.Conflict:
             if not exists_ok:
@@ -479,7 +495,7 @@ class Client(ClientWithProject):
             return self.get_dataset(dataset.reference, retry=retry)
 
     def create_routine(
-        self, routine, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
+            self, routine, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
     ):
         """[Beta] Create a routine via a POST request.
 
@@ -513,9 +529,13 @@ class Client(ClientWithProject):
         )
         resource = routine.to_api_repr()
         try:
-            api_response = self._call_api(
-                retry, method="POST", path=path, data=resource, timeout=timeout
-            )
+            span_creator = SpanCreator()
+            attributes = {'path': path,
+                          'exists_ok': exists_ok}
+            with span_creator.create('BigQuery.createRoutine', attributes):
+                api_response = self._call_api(
+                    retry, method="POST", path=path, data=resource, timeout=timeout
+                )
             return Routine.from_api_repr(api_response)
         except google.api_core.exceptions.Conflict:
             if not exists_ok:
@@ -556,13 +576,17 @@ class Client(ClientWithProject):
                 If the table already exists.
         """
         table = _table_arg_to_table(table, default_project=self.project)
-
-        path = "/projects/%s/datasets/%s/tables" % (table.project, table.dataset_id)
+        dataset_id = table.dataset_id
+        path = "/projects/%s/datasets/%s/tables" % (table.project, dataset_id)
         data = table.to_api_repr()
         try:
-            api_response = self._call_api(
-                retry, method="POST", path=path, data=data, timeout=timeout
-            )
+            span_creator = SpanCreator()
+            attributes = {'path': path,
+                          'dataset_id': dataset_id}
+            with span_creator.create('BigQuery.createTable', attributes):
+                api_response = self._call_api(
+                    retry, method="POST", path=path, data=data, timeout=timeout
+                )
             return Table.from_api_repr(api_response)
         except google.api_core.exceptions.Conflict:
             if not exists_ok:
@@ -597,18 +621,24 @@ class Client(ClientWithProject):
             google.cloud.bigquery.dataset.Dataset:
                 A ``Dataset`` instance.
         """
+
         if isinstance(dataset_ref, str):
+            ref = dataset_ref
             dataset_ref = DatasetReference.from_string(
                 dataset_ref, default_project=self.project
             )
-
-        api_response = self._call_api(
-            retry, method="GET", path=dataset_ref.path, timeout=timeout
-        )
+        span_creator = SpanCreator()
+        path = dataset_ref.path
+        attributes = {'path': path,
+                      'dataset_ref': ref}
+        with span_creator.create('BigQuery.createTable', attributes):
+            api_response = self._call_api(
+                retry, method="GET", path=path, timeout=timeout
+            )
         return Dataset.from_api_repr(api_response)
 
     def get_iam_policy(
-        self, table, requested_policy_version=1, retry=DEFAULT_RETRY, timeout=None,
+            self, table, requested_policy_version=1, retry=DEFAULT_RETRY, timeout=None,
     ):
         if not isinstance(table, (Table, TableReference)):
             raise TypeError("table must be a Table or TableReference")
@@ -619,15 +649,18 @@ class Client(ClientWithProject):
         body = {"options": {"requestedPolicyVersion": 1}}
 
         path = "{}:getIamPolicy".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'body': body}
+        with span_creator.create('BigQuery.getIAMPolicy', attributes):
+            response = self._call_api(
+                retry, method="POST", path=path, data=body, timeout=timeout,
+            )
 
         return Policy.from_api_repr(response)
 
     def set_iam_policy(
-        self, table, policy, updateMask=None, retry=DEFAULT_RETRY, timeout=None,
+            self, table, policy, updateMask=None, retry=DEFAULT_RETRY, timeout=None,
     ):
         if not isinstance(table, (Table, TableReference)):
             raise TypeError("table must be a Table or TableReference")
@@ -641,15 +674,18 @@ class Client(ClientWithProject):
             body["updateMask"] = updateMask
 
         path = "{}:setIamPolicy".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'body': body}
+        with span_creator.create('BigQuery.setIAMPolicy', attributes):
+            response = self._call_api(
+                retry, method="POST", path=path, data=body, timeout=timeout,
+            )
 
         return Policy.from_api_repr(response)
 
     def test_iam_permissions(
-        self, table, permissions, retry=DEFAULT_RETRY, timeout=None,
+            self, table, permissions, retry=DEFAULT_RETRY, timeout=None,
     ):
         if not isinstance(table, (Table, TableReference)):
             raise TypeError("table must be a Table or TableReference")
@@ -657,10 +693,13 @@ class Client(ClientWithProject):
         body = {"permissions": permissions}
 
         path = "{}:testIamPermissions".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'body': body}
+        with span_creator.create('BigQuery.testIAMPolicy', attributes):
+            response = self._call_api(
+                retry, method="POST", path=path, data=body, timeout=timeout,
+            )
 
         return response
 
@@ -686,13 +725,18 @@ class Client(ClientWithProject):
             google.cloud.bigquery.model.Model: A ``Model`` instance.
         """
         if isinstance(model_ref, str):
+            ref = model_ref
             model_ref = ModelReference.from_string(
                 model_ref, default_project=self.project
             )
-
-        api_response = self._call_api(
-            retry, method="GET", path=model_ref.path, timeout=timeout
-        )
+        span_creator = SpanCreator()
+        path = model_ref.path
+        attributes = {'path': path,
+                      'model_ref': ref}
+        with span_creator.create('BigQuery.getModel', attributes):
+            api_response = self._call_api(
+                retry, method="GET", path=path, timeout=timeout
+            )
         return Model.from_api_repr(api_response)
 
     def get_routine(self, routine_ref, retry=DEFAULT_RETRY, timeout=None):
@@ -718,14 +762,20 @@ class Client(ClientWithProject):
             google.cloud.bigquery.routine.Routine:
                 A ``Routine`` instance.
         """
+
         if isinstance(routine_ref, str):
+            ref = routine_ref
             routine_ref = RoutineReference.from_string(
                 routine_ref, default_project=self.project
             )
-
-        api_response = self._call_api(
-            retry, method="GET", path=routine_ref.path, timeout=timeout
-        )
+        path = routine_ref.path
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'routine_ref': ref}
+        with span_creator.create('BigQuery.getRoutine', attributes):
+            api_response = self._call_api(
+                retry, method="GET", path=path, timeout=timeout
+            )
         return Routine.from_api_repr(api_response)
 
     def get_table(self, table, retry=DEFAULT_RETRY, timeout=None):
@@ -752,9 +802,13 @@ class Client(ClientWithProject):
                 A ``Table`` instance.
         """
         table_ref = _table_arg_to_table_ref(table, default_project=self.project)
-        api_response = self._call_api(
-            retry, method="GET", path=table_ref.path, timeout=timeout
-        )
+        path = table_ref.path
+        span_creator = SpanCreator()
+        attributes = {'path': path}
+        with span_creator.create('BigQuery.getTable', attributes):
+            api_response = self._call_api(
+                retry, method="GET", path=path, timeout=timeout
+            )
         return Table.from_api_repr(api_response)
 
     def update_dataset(self, dataset, fields, retry=DEFAULT_RETRY, timeout=None):
@@ -791,14 +845,20 @@ class Client(ClientWithProject):
             headers = {"If-Match": dataset.etag}
         else:
             headers = None
-        api_response = self._call_api(
-            retry,
-            method="PATCH",
-            path=dataset.path,
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+        path = dataset.path
+        span_creator = SpanCreator()
+        attributes = headers
+        attributes['path'] = path
+        attributes['fields'] = fields
+        with span_creator.create('BigQuery.updateDataset', attributes):
+            api_response = self._call_api(
+                retry,
+                method="PATCH",
+                path=dataset.path,
+                data=partial,
+                headers=headers,
+                timeout=timeout,
+            )
         return Dataset.from_api_repr(api_response)
 
     def update_model(self, model, fields, retry=DEFAULT_RETRY, timeout=None):
@@ -834,14 +894,20 @@ class Client(ClientWithProject):
             headers = {"If-Match": model.etag}
         else:
             headers = None
-        api_response = self._call_api(
-            retry,
-            method="PATCH",
-            path=model.path,
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+        path = model.path
+        span_creator = SpanCreator()
+        attributes = headers
+        attributes['path'] = path
+        attributes['fields'] = fields
+        with span_creator.create('BigQuery.updateModel', attributes):
+            api_response = self._call_api(
+                retry,
+                method="PATCH",
+                path=path,
+                data=partial,
+                headers=headers,
+                timeout=timeout,
+            )
         return Model.from_api_repr(api_response)
 
     def update_routine(self, routine, fields, retry=DEFAULT_RETRY, timeout=None):
@@ -888,14 +954,20 @@ class Client(ClientWithProject):
         # TODO: remove when routines update supports partial requests.
         partial["routineReference"] = routine.reference.to_api_repr()
 
-        api_response = self._call_api(
-            retry,
-            method="PUT",
-            path=routine.path,
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+        path = routine.path
+        span_creator = SpanCreator()
+        attributes = headers
+        attributes['path'] = path
+        attributes['fields'] = fields
+        with span_creator.create('BigQuery.updateRountine', attributes):
+            api_response = self._call_api(
+                retry,
+                method="PUT",
+                path=path,
+                data=partial,
+                headers=headers,
+                timeout=timeout,
+            )
         return Routine.from_api_repr(api_response)
 
     def update_table(self, table, fields, retry=DEFAULT_RETRY, timeout=None):
@@ -931,23 +1003,30 @@ class Client(ClientWithProject):
             headers = {"If-Match": table.etag}
         else:
             headers = None
-        api_response = self._call_api(
-            retry,
-            method="PATCH",
-            path=table.path,
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+
+        path = table.path
+        span_creator = SpanCreator()
+        attributes = headers
+        attributes['path'] = path
+        attributes['fields'] = fields
+        with span_creator.create('BigQuery.updateTable', attributes):
+            api_response = self._call_api(
+                retry,
+                method="PATCH",
+                path=table.path,
+                data=partial,
+                headers=headers,
+                timeout=timeout,
+            )
         return Table.from_api_repr(api_response)
 
     def list_models(
-        self,
-        dataset,
-        max_results=None,
-        page_token=None,
-        retry=DEFAULT_RETRY,
-        timeout=None,
+            self,
+            dataset,
+            max_results=None,
+            page_token=None,
+            retry=DEFAULT_RETRY,
+            timeout=None,
     ):
         """[Beta] List models in the dataset.
 
@@ -994,9 +1073,15 @@ class Client(ClientWithProject):
             raise TypeError("dataset must be a Dataset, DatasetReference, or string")
 
         path = "%s/models" % dataset.path
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'page_token': page_token,
+                      'max_results': max_results}
+        with span_creator.create('BigQuery.listModels', attributes):
+            api_request = functools.partial(self._call_api, retry, timeout=timeout)
         result = page_iterator.HTTPIterator(
             client=self,
-            api_request=functools.partial(self._call_api, retry, timeout=timeout),
+            api_request=api_request,
             path=path,
             item_to_value=_item_to_model,
             items_key="models",
@@ -1007,12 +1092,12 @@ class Client(ClientWithProject):
         return result
 
     def list_routines(
-        self,
-        dataset,
-        max_results=None,
-        page_token=None,
-        retry=DEFAULT_RETRY,
-        timeout=None,
+            self,
+            dataset,
+            max_results=None,
+            page_token=None,
+            retry=DEFAULT_RETRY,
+            timeout=None,
     ):
         """[Beta] List routines in the dataset.
 
@@ -1059,9 +1144,15 @@ class Client(ClientWithProject):
             raise TypeError("dataset must be a Dataset, DatasetReference, or string")
 
         path = "{}/routines".format(dataset.path)
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'page_token': page_token,
+                      'max_results': max_results}
+        with span_creator.create('BigQuery.listRoutines', attributes):
+            api_requests = functools.partial(self._call_api, retry, timeout=timeout)
         result = page_iterator.HTTPIterator(
             client=self,
-            api_request=functools.partial(self._call_api, retry, timeout=timeout),
+            api_request=api_requests,
             path=path,
             item_to_value=_item_to_routine,
             items_key="routines",
@@ -1072,12 +1163,12 @@ class Client(ClientWithProject):
         return result
 
     def list_tables(
-        self,
-        dataset,
-        max_results=None,
-        page_token=None,
-        retry=DEFAULT_RETRY,
-        timeout=None,
+            self,
+            dataset,
+            max_results=None,
+            page_token=None,
+            retry=DEFAULT_RETRY,
+            timeout=None,
     ):
         """List tables in the dataset.
 
@@ -1124,9 +1215,16 @@ class Client(ClientWithProject):
             raise TypeError("dataset must be a Dataset, DatasetReference, or string")
 
         path = "%s/tables" % dataset.path
+        span_creator = SpanCreator()
+        attributes = {'path': path,
+                      'page_token': page_token,
+                      'max_results': max_results}
+        with span_creator.create('BigQuery.listTables', attributes):
+            api_requests = functools.partial(self._call_api, retry, timeout=timeout)
+
         result = page_iterator.HTTPIterator(
             client=self,
-            api_request=functools.partial(self._call_api, retry, timeout=timeout),
+            api_request=api_requests,
             path=path,
             item_to_value=_item_to_table,
             items_key="tables",
@@ -1137,12 +1235,12 @@ class Client(ClientWithProject):
         return result
 
     def delete_dataset(
-        self,
-        dataset,
-        delete_contents=False,
-        retry=DEFAULT_RETRY,
-        timeout=None,
-        not_found_ok=False,
+            self,
+            dataset,
+            delete_contents=False,
+            retry=DEFAULT_RETRY,
+            timeout=None,
+            not_found_ok=False,
     ):
         """Delete a dataset.
 
@@ -1181,23 +1279,28 @@ class Client(ClientWithProject):
             raise TypeError("dataset must be a Dataset or a DatasetReference")
 
         params = {}
+        path = dataset.path
         if delete_contents:
             params["deleteContents"] = "true"
 
         try:
-            self._call_api(
-                retry,
-                method="DELETE",
-                path=dataset.path,
-                query_params=params,
-                timeout=timeout,
-            )
+            span_creator = SpanCreator()
+            attributes = params
+            attributes["path"] = path
+            with span_creator.create('BigQuery.deleteDataset', attributes):
+                self._call_api(
+                    retry,
+                    method="DELETE",
+                    path=path,
+                    query_params=params,
+                    timeout=timeout,
+                )
         except google.api_core.exceptions.NotFound:
             if not not_found_ok:
                 raise
 
     def delete_model(
-        self, model, retry=DEFAULT_RETRY, timeout=None, not_found_ok=False
+            self, model, retry=DEFAULT_RETRY, timeout=None, not_found_ok=False
     ):
         """[Beta] Delete a model
 
@@ -1224,19 +1327,26 @@ class Client(ClientWithProject):
                 when deleting the model.
         """
         if isinstance(model, str):
+            ref = model
             model = ModelReference.from_string(model, default_project=self.project)
 
         if not isinstance(model, (Model, ModelReference)):
             raise TypeError("model must be a Model or a ModelReference")
 
+        path = model.path
         try:
-            self._call_api(retry, method="DELETE", path=model.path, timeout=timeout)
+            span_creator = SpanCreator()
+            attributes = {"model": ref,
+                          "not_found_okay": not_found_ok,
+                          "path": path}
+            with span_creator.create('BigQuery.deleteModel', attributes):
+                self._call_api(retry, method="DELETE", path=path, timeout=timeout)
         except google.api_core.exceptions.NotFound:
             if not not_found_ok:
                 raise
 
     def delete_routine(
-        self, routine, retry=DEFAULT_RETRY, timeout=None, not_found_ok=False
+            self, routine, retry=DEFAULT_RETRY, timeout=None, not_found_ok=False
     ):
         """[Beta] Delete a routine.
 
@@ -1263,21 +1373,27 @@ class Client(ClientWithProject):
                 when deleting the routine.
         """
         if isinstance(routine, str):
+            ref = routine
             routine = RoutineReference.from_string(
                 routine, default_project=self.project
             )
+        path = routine.path
 
         if not isinstance(routine, (Routine, RoutineReference)):
             raise TypeError("routine must be a Routine or a RoutineReference")
 
         try:
-            self._call_api(retry, method="DELETE", path=routine.path, timeout=timeout)
+            span_creator = SpanCreator()
+            attributes = {"path": path,
+                          "routine": ref}
+            with span_creator.create('BigQuery.deleteRoutine', attributes):
+                self._call_api(retry, method="DELETE", path=path, timeout=timeout)
         except google.api_core.exceptions.NotFound:
             if not not_found_ok:
                 raise
 
     def delete_table(
-        self, table, retry=DEFAULT_RETRY, timeout=None, not_found_ok=False
+            self, table, retry=DEFAULT_RETRY, timeout=None, not_found_ok=False
     ):
         """Delete a table
 
@@ -1308,13 +1424,17 @@ class Client(ClientWithProject):
             raise TypeError("Unable to get TableReference for table '{}'".format(table))
 
         try:
-            self._call_api(retry, method="DELETE", path=table.path, timeout=timeout)
+            path = table.path
+            span_creator = SpanCreator()
+            attributes = {"path": path}
+            with span_creator.create('BigQuery.deleteTable', attributes):
+                self._call_api(retry, method="DELETE", path=path, timeout=timeout)
         except google.api_core.exceptions.NotFound:
             if not not_found_ok:
                 raise
 
     def _get_query_results(
-        self, job_id, retry, project=None, timeout_ms=None, location=None, timeout=None
+            self, job_id, retry, project=None, timeout_ms=None, location=None, timeout=None
     ):
         """Get the query results object for a query job.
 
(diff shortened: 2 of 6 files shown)


README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
 
     for row in rows:
         print(row.name)
+
+Instrumenting With OpenTelemetry
+--------------------------------
+
+This application uses OpenTelmetry to output tracing data from
+api calls to BigQuery. To enable OpenTelemetry tracing in
```
tswast, line 110:
```suggestion
API calls to BigQuery. To enable OpenTelemetry tracing in
```

README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
...
+the BigQuery client the following PyPi packages need to be installed:
```
tswast, line 111:
```suggestion
the BigQuery client the following PyPI packages need to be installed:
```

README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
...
+
+.. code-block:: console
+
+    pip install opentelemetry-api
```
tswast, line 115: Generally, we define these as optional dependencies using `extras`. See `setup.py` for examples. Then, these instructions become `pip install google-cloud-bigquery[opentelemetry]`

README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
...
+.. code-block:: console
+
+    pip install opentelemetry-instrumentaion
+
+.. code-block:: console
+
+    pip install opentelemetry-exporter-cloud_trace
+
+After installation, OpenTelemetry can now be used in the BigQuery
+client and in BigQuery jobs. First however an exporter must be
```
tswast, line 130:
```suggestion
client and in BigQuery jobs. First, however, an exporter must be
```

README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
...
+
+.. code-block:: console
+
+    pip install opentelemetry-instrumentaion
+
+.. code-block:: console
+
+    pip install opentelemetry-exporter-cloud_trace
+
+After installation, OpenTelemetry can now be used in the BigQuery
```
tswast, line 129:
```suggestion
After installation, OpenTelemetry can be used in the BigQuery
```

README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
...
+client and in BigQuery jobs. First however an exporter must be
+specified for where the trace data will be outputted to. An example of this
+can be found here:
+
+.. code-block:: python
+
+    from opentelemetry import trace
```
tswast, line 136: Is there a runnable / tested code snippet somewhere this can be linked from? I'm worried that the code sample may get out-of-date / break.

README.rst
```diff
@@ -102,3 +102,51 @@ Perform a query
...
+    from opentelemetry.sdk.trace import TracerProvider
+    from opentelemetry.sdk.trace.export import SimpleExportSpanProcessor
+    from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+    trace.set_tracer_provider(TracerProvider())
+    trace.get_tracer_provider().add_span_processor(
+        SimpleExportSpanProcessor(CloudTraceSpanExporter())
+    )
+
+In this example all tracing data will be published to the Google
+Cloud trace console. For more information on OpenTelemetry, please consult the `OpenTelemetry documentation`_.
```
tswast, line 146:
```suggestion
[Cloud Trace](https://cloud.google.com/trace) console. For more information on OpenTelemetry, please consult the `OpenTelemetry documentation`_.
```

google/cloud/bigquery/job.py
```diff
@@ -634,9 +636,18 @@ def _begin(self, client=None, retry=DEFAULT_RETRY, timeout=None):
 
         # jobs.insert is idempotent because we ensure that every new
         # job has an ID.
-        api_response = client._call_api(
-            retry, method="POST", path=path, data=self.to_api_repr(), timeout=timeout
-        )
+        attributes = {"path": path}
+        span_creator = SpanCreator()
```
tswast, line 640: I'm curious why this has to be constructed and then also call `create`? Maybe `create` can be turned into a `classmethod` (or if there's minimal logic, moved to the constructor)?

google/cloud/bigquery/job.py
```diff
@@ -634,9 +636,18 @@ def _begin(self, client=None, retry=DEFAULT_RETRY, timeout=None):
 
         # jobs.insert is idempotent because we ensure that every new
         # job has an ID.
-        api_response = client._call_api(
-            retry, method="POST", path=path, data=self.to_api_repr(), timeout=timeout
-        )
+        attributes = {"path": path}
```
tswast, line 639: Attributes is a bit ambiguous of a name, as it could also be something related to the request we're making. How about `span_attributes` or just remove the variable entirely and construct the dictionary in the create call?

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,88 @@
...
+        "packages in order to get Big Query Tracing data"
+    )
+
+    HAS_OPENTELEMETRY = False
+
+
+class SpanCreator:
+    def __init__(self):
+        # Constructs a span creator with all default attributes
+        self.opentelemetry_enbled = HAS_OPENTELEMETRY
```
tswast, line 41: Since these are both global properties, maybe an object isn't even required? Could `create` be made into just a module-level method? https://www.youtube.com/watch?v=o9pEzgHorH0

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,88 @@
...
+            name=name, attributes=self.attributes
+        ) as span:
+            try:
+                yield span
+            except GoogleAPICallError as error:
+                if error.code is not None:
+                    span.set_status(Status(http_status_to_canonical_code(error.code)))
+                raise
+
+    def set_client_attributes(self, client):
```
tswast, line 73: These helper methods feel like they could be private module-level functions.

```
def _get_client_attributes(client):
    return {
        "db.name": client.project,
        "location": client.location,
    }

# To use the helper...
attributes = {...}
attributes.update(_get_client_attributes(client))
```

noxfile.py
```diff
@@ -33,7 +33,14 @@ def default(session):
     """
     # Install all test dependencies, then install local packages in-place.
     session.install(
-        "mock", "pytest", "google-cloud-testutils", "pytest-cov", "freezegun"
+        "mock",
+        "pytest",
+        "google-cloud-testutils",
+        "pytest-cov",
+        "freezegun",
+        "opentelemetry-api",
```
tswast, line 41: Ideally all of these extra dependencies would be included in the `[all]` extras used below.

tswast:
Build is breaking:

https://source.cloud.google.com/results/invocations/6578f68f-1263-4029-ac27-22c033fe6d86/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fdocs%2Fdocs-presubmit/log

google/cloud/bigquery/client.py
```diff
@@ -656,10 +705,14 @@ def test_iam_permissions(
...
 
         path = "{}:testIamPermissions".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_creator = SpanCreator()
+        attributes = {"path": path, "body": body}
+        with span_creator.create(
+            name="BigQuery.testIAMPolicy", attributes=attributes, client=self
```
sjbrunst, line 711: This name is `BigQuery.testIAMPolicy`, but the function is `test_iam_permissions`. The policy/permissions naming difference could be confusing; is it intentional?

google/cloud/bigquery/client.py
```diff
@@ -887,14 +971,21 @@ def update_routine(self, routine, fields, retry=DEFAULT_RETRY, timeout=None):
...
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+        path = routine.path
+        span_creator = SpanCreator()
+        attributes = {"path": path, "fields": fields}
+
+        with span_creator.create(
+            name="BigQuery.updateRountine", attributes=attributes, client=self
```
sjbrunst, line 979:
```suggestion
            name="BigQuery.updateRoutine", attributes=attributes, client=self
```

google/cloud/bigquery/client.py
```diff
@@ -1549,10 +1723,20 @@ def cancel_job(
...
         path = "/projects/{}/jobs/{}/cancel".format(project, job_id)
+        span_creator = SpanCreator()
 
-        resource = self._call_api(
-            retry, method="POST", path=path, query_params=extra_params, timeout=timeout
-        )
+        attributes = {"path": path, "job_id": job_id, "location": location}
+
+        with span_creator.create(
+            name="BigQuery.CancelJob", attributes=attributes, client=self
```
sjbrunst, line 1731: Most of the methods used lower case for the first letter; should this name be `BigQuery.cancelJob`?

google/cloud/bigquery/opentelemetry_tracing.py
aravinsiva: yeah I think a module level method might make more sense will do

README.rst
aravinsiva: Currently looking into it.

google/cloud/bigquery/job.py
aravinsiva: Created module level methods instead of using a class. As the class seemed unnecessary

noxfile.py
aravinsiva: I am not entirely sure where `[all]` is defined. I am also not completely familiar with nox.

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,88 @@
...
+from contextlib import contextmanager
+from google.api_core.exceptions import GoogleAPICallError
+
+Logger = logging.getLogger(__name__)
+try:
+    from opentelemetry import trace
+    from opentelemetry.trace.status import Status
+    from opentelemetry.instrumentation.utils import http_status_to_canonical_code
+
+    HAS_OPENTELEMETRY = True
```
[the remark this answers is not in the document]
aravinsiva: Waiting on Tim to provide more context.

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,88 @@
...
+class SpanCreator:
+    def __init__(self):
+        # Constructs a span creator with all default attributes
+        self.opentelemetry_enbled = HAS_OPENTELEMETRY
+        self.attributes = {
+            "db.system": "bigquery",
+        }
+
+    @contextmanager
+    def create(self, name, attributes=None, client=None, job_ref=None):
```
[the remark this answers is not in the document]
aravinsiva: Still need to add these type hints.

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,88 @@
...
+            self.attributes.update(attributes)
+        # yield new span value
+        with tracer.start_as_current_span(
+            name=name, attributes=self.attributes
+        ) as span:
+            try:
+                yield span
+            except GoogleAPICallError as error:
+                if error.code is not None:
+                    span.set_status(Status(http_status_to_canonical_code(error.code)))
```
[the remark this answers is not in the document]
aravinsiva: Not entirely sure what status code that entails?

aravinsiva:
> I've added a handful of comments/questions.
> 
> One open question: You have a unit test for the SpanCreator class, but the individual API calls aren't tested for instrumentation. I'm not sure what the testing strategy is for this repo, but such tests would be nice to have.

Will add these tests unless @tswast has any objections.

google/cloud/bigquery/opentelemetry_tracing.py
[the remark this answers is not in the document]
kintel: If an HTTP request is successful, we usually surface the success code (e.g. 200, 201 etc.) in the Cloud Trace UI. This comes from a span attribute. In a similar fashion we may want to write this status code even if the above exception isn't raised.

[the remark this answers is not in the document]
aravinsiva: Ah okay just wanted to mkae sure it was an HTTP request code. Sounds good

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -26,63 +26,62 @@
...
-        self.attributes["state"] = job_ref.state
+@contextmanager
+def create_span(name, attributes=None, client=None, job_ref=None):
+    if not HAS_OPENTELEMETRY:
+        yield None
+        return
+
+    default_attributes = {
+        "db.system": "BigQuery",
+    }
```
kintel, line 46: Nit: Could be a global private dict, to avoid re-building it for every invocation.

google/cloud/bigquery/opentelemetry_tracing.py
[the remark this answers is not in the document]
tswast: Yes, since this is a constant (set once, never mutated) we use uppercase

noxfile.py
tswast: Now that you've added `opentelemetry` to https://github.com/googleapis/python-bigquery/blob/bae233770463db697df63dc66f9a6f54bd10d437/setup.py#L38 these dependencies are all included in `[all]`. It's defined here:

https://github.com/googleapis/python-bigquery/blob/bae233770463db697df63dc66f9a6f54bd10d437/setup.py#L65-L75

setup.py
```diff
@@ -60,6 +60,12 @@
         "llvmlite <= 0.33.0;python_version>='3.6'",
         "llvmlite <= 0.31.0;python_version<'3.6'",
     ],
+    "opentelemetry": [
+        "opentelemetry-api",
```
tswast, line 64: Are there minimum / maximum versions we should set for these packages? If the package follows semantic versioning, we usually try to at least set a max version in case there are breaking changes in the future.

kintel: So far I think any 0.x release may have breaking changes between versions as the API only started to stabilize somewhat lately. I believe most places use exact versioning due to this. Once they hit 0.9 or 1.0, I expect semantic versioning to kick in.

aravinsiva: Will add exact versioning to avoid this.

google/cloud/bigquery/opentelemetry_tracing.py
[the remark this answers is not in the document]
aravinsiva: Added comments for documentation so more information could be included. Includes types for arguments and return types.

aravinsiva:
> > I've added a handful of comments/questions.
> > One open question: You have a unit test for the SpanCreator class, but the individual API calls aren't tested for instrumentation. I'm not sure what the testing strategy is for this repo, but such tests would be nice to have.
> 
> Will add these tests unless @tswast has any objections.

These tests will be added in a separate PR to make PR's a bit more modular.

README.rst
```diff
@@ -102,3 +102,42 @@ Perform a query
 
     for row in rows:
         print(row.name)
+
+Instrumenting With OpenTelemetry
+--------------------------------
+
+This application uses `OpenTelemetry`_
+to output tracing data from
```
sjbrunst, line 110: nit: the line wrapping here (and lines 120-122) seems inconsistent

google/cloud/bigquery/client.py
```diff
@@ -278,10 +285,16 @@ def list_projects(
                 Iterator of :class:`~google.cloud.bigquery.client.Project`
                 accessible to the current client.
         """
+        path = "/projects"
+        span_attributes = {"path": path}
+        with create_span(
+            name="BigQuery.list_projects", attributes=span_attributes, client=self
```
sjbrunst, line 291:
```suggestion
            name="BigQuery.listProjects", attributes=span_attributes, client=self
```
(Not sure which naming scheme is desired, but I'm assuming these should be camel case for consistency.)

google/cloud/bigquery/client.py
```diff
@@ -342,9 +355,15 @@ def list_datasets(
             # and converting it into a string here.
             extra_params["filter"] = filter
         path = "/projects/%s/datasets" % (project,)
+        span_attributes = {"path": path, "page_token": page_token}
+
+        with create_span(
+            name="BigQuery.list_datasets", attributes=span_attributes, client=self
```
sjbrunst, line 361:
```suggestion
            name="BigQuery.listDatasets", attributes=span_attributes, client=self
```

google/cloud/bigquery/client.py
```diff
@@ -596,14 +627,19 @@ def get_dataset(self, dataset_ref, retry=DEFAULT_RETRY, timeout=None):
...
                 dataset_ref, default_project=self.project
             )
-
-        api_response = self._call_api(
-            retry, method="GET", path=dataset_ref.path, timeout=timeout
-        )
+        path = dataset_ref.path
+        span_attributes = {"path": path}
+        with create_span(
+            name="BigQuery.createTable", attributes=span_attributes, client=self
```
sjbrunst, line 638:
```suggestion
            name="BigQuery.getDataset", attributes=span_attributes, client=self
```

google/cloud/bigquery/client.py
```diff
@@ -656,10 +698,13 @@ def test_iam_permissions(
         body = {"permissions": permissions}
 
         path = "{}:testIamPermissions".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_attributes = {"path": path, "body": body}
+        with create_span(
+            name="BigQuery.testIamPermissions", attributes=span_attributes, client=self
```
sjbrunst, line 703:
```suggestion
            name="BigQuery.testIAMPermissions", attributes=span_attributes, client=self
```
(for consistency with `getIAMPolicy` and `setIAMPolicy`)

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,104 @@
...
+
+from google.cloud.bigquery import opentelemetry_tracing, client
+
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleExportSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+
+TEST_SPAN_NAME = "bar"
+TEST_SPAN_ATTRIBUTES = {"foo": "bar"}
```
sjbrunst, line 15:
```suggestion
TEST_SPAN_ATTRIBUTES = {"foo": "baz"}
```

(to avoid confusion with the `TEST_SPAN_NAME`)

setup.py
```diff
@@ -61,10 +61,10 @@
...
     ],
     "opentelemetry": [
-        "opentelemetry-api",
-        "opentelemetry-sdk",
-        "opentelemetry-instrumentaion",
-        "opentelemetry-exporter-cloud-trace",
+        "opentelemetry-api == 0.9b0",
+        "opentelemetry-sdk == 0.9b0",
+        "opentelemetry-instrumentation == 0.9b0 ",
+        "opentelemetry-exporter-cloud-trace == 0.110",
```
kintel, line 67: typo? -> 0.11.0 ?

setup.py
```diff
@@ -60,6 +60,12 @@
         "llvmlite <= 0.33.0;python_version>='3.6'",
         "llvmlite <= 0.31.0;python_version<'3.6'",
     ],
+    "opentelemetry": [
+        "opentelemetry-api == 0.9b0",
+        "opentelemetry-sdk == 0.9b0",
+        "opentelemetry-instrumentation == 0.9b0 ",
+        "opentelemetry-exporter-cloud-trace == 0.9b0",
```
AndrewAXue, line 67: This package has been renamed to "opentelemetry-exporter-google-cloud"

aravinsiva: removed from nox since not needed and updated package name is referenced in documentation.

tswast:
Re: test failure. You're going to need to add unit tests to this PR. The Python repos enforce 100% unit test coverage. This is because there are many classes of errors the can happen at runtime due to Python's dynamic nature.

README.rst
```diff
@@ -102,3 +102,41 @@ Perform a query
...
+
+This application uses `OpenTelemetry`_ to output tracing data from
+API calls to BigQuery. To enable OpenTelemetry tracing in
+the BigQuery client the following PyPI packages need to be installed:
+
+.. _OpenTelemetry: https://opentelemetry.io
+
+.. code-block:: console
+
+    pip install opentelemetry-api opentelemetry-sdk opentelemetry-instrumentation opentelemetry-exporter-google-cloud
```
tswast, line 117:
```suggestion
    pip install google-cloud-bigquery[opentelemetry]
```

google/cloud/bigquery/client.py
```diff
@@ -278,10 +285,16 @@ def list_projects(
                 Iterator of :class:`~google.cloud.bigquery.client.Project`
                 accessible to the current client.
         """
+        path = "/projects"
+        span_attributes = {"path": path}
+        with create_span(
+            name="BigQuery.listProjects", attributes=span_attributes, client=self
+        ):
+            api_request = functools.partial(self._call_api, retry, timeout=timeout)
```
tswast, line 293: This actually doesn't make any API calls. `functools.partial` creates a function object that will call `_call_api` with the specified arguments.

If you want `HTTPIterator` to actually create spans, you'll have to create a function that does this manually. Something like:

```python
def list_projects(...):
    ...
    def api_request(*args, **kwargs):
        with create_span(
            name="BigQuery.listProjects", attributes=span_attributes, client=self
        ):
            return self._call_api(retry, *args, timeout=timeout, **kwargs)
    return page_iterator.HTTPIterator(...)
```

google/cloud/bigquery/client.py
```diff
@@ -342,9 +355,15 @@ def list_datasets(
             # and converting it into a string here.
             extra_params["filter"] = filter
         path = "/projects/%s/datasets" % (project,)
+        span_attributes = {"path": path, "page_token": page_token}
+
+        with create_span(
+            name="BigQuery.listDatasets", attributes=span_attributes, client=self
+        ):
+            api_request = functools.partial(self._call_api, retry, timeout=timeout)
```
tswast, line 363: Same idea applies here. You'll need to define an `api_request` function rather than use functools.partial if you want the span to be created when the actual API request happens.

google/cloud/bigquery/client.py
aravinsiva: Acknowledged in a separate PR as per our conversation. Draft can be found here:
https://github.com/googleapis/python-bigquery/pull/228

google/cloud/bigquery/client.py
aravinsiva: Acknowledged in a separate PR as per our conversation. Draft can be found here:
https://github.com/googleapis/python-bigquery/pull/228

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,104 @@
+import sys
+import pytest
+
+from importlib import reload
```
tswast, line 4: These tests are failing on Python 2. Need to use `six.moves.reload_module` https://six.readthedocs.io/#module-six.moves

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,104 @@
...
+import mock
+
+from google.cloud.bigquery import opentelemetry_tracing, client
+
+from opentelemetry import trace
```
tswast, line 9: I recommend using `pytest.importorskip` https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency

This will allow the tests to run when opentelemetry is not installed.

tswast: Then again, we probably need a test to check that things succeed when the library is not installed. See the https://github.com/googleapis/python-bigquery/blob/master/tests/unit/test__pandas_helpers.py tests for how to mock out imports to have a couple tests for the "no opentelemetry" path. See: https://github.com/googleapis/python-bigquery/blob/master/tests/unit/test__pandas_helpers.py#L954

google/cloud/bigquery/client.py
```diff
@@ -659,10 +701,13 @@ def test_iam_permissions(
         body = {"permissions": permissions}
 
         path = "{}:testIamPermissions".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_attributes = {"path": path, "body": body}
+        with create_span(
+            name="BigQuery.testIAMPermissions", attributes=span_attributes, client=self
```
sjbrunst, line 706:
```suggestion
            name="BigQuery.testIamPermissions", attributes=span_attributes, client=self
```

google/cloud/bigquery/client.py
```diff
@@ -621,10 +657,13 @@ def get_iam_policy(
         body = {"options": {"requestedPolicyVersion": 1}}
 
         path = "{}:getIamPolicy".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_attributes = {"path": path, "body": body}
```
sjbrunst, line 660: `body` is a dict, which is not one of the supported attribute types: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes

google/cloud/bigquery/client.py
```diff
@@ -659,10 +701,13 @@ def test_iam_permissions(
         body = {"permissions": permissions}
 
         path = "{}:testIamPermissions".format(table.path)
-
-        response = self._call_api(
-            retry, method="POST", path=path, data=body, timeout=timeout,
-        )
+        span_attributes = {"path": path, "body": body}
```
sjbrunst, line 704: Same comment here; I don't think dict attributes are supported.

google/cloud/bigquery/client.py
```diff
@@ -2875,11 +3045,22 @@ def list_rows(
...
+            "path": path,
+            "page_size": page_size,
+            "page_token": page_token,
+            "start_index": start_index,
+        }
 
+        with create_span(
+            name="BigQuery.listRows", attributes=span_attributes, client=self
+        ):
+            api_request = functools.partial(self._call_api, retry, timeout=timeout)
```
sjbrunst, line 3059: any particular reason for moving this into a variable instead of creating it inline?

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+    from opentelemetry.instrumentation.utils import http_status_to_canonical_code
+
+    HAS_OPENTELEMETRY = True
+
+except ImportError:
+    Logger.info(
+        "This service is instrumented using opentelemetry."
+        "Opentelemetry could not be imported, please"
+        "add opentelemetry-api and opentelemetry-instrumentation"
+        "packages, in order to get Big Query Tracing data"
```
sjbrunst, line 32:
```suggestion
        "packages in order to get BigQuery tracing data."
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+try:
+    from opentelemetry import trace
+    from opentelemetry.trace.status import Status
+    from opentelemetry.instrumentation.utils import http_status_to_canonical_code
+
+    HAS_OPENTELEMETRY = True
+
+except ImportError:
+    Logger.info(
+        "This service is instrumented using opentelemetry."
```
sjbrunst, line 29:
```suggestion
        "This service is instrumented using OpenTelemetry."
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+        "Opentelemetry could not be imported, please"
```
sjbrunst, line 30:
```suggestion
        "OpenTelemetry could not be imported; please"
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@contextmanager
+def create_span(name, attributes=None, client=None, job_ref=None):
+    """Creates a ContextManager for a Span to be exported to the configured exporter. If no configuration
+        exists yields None.
+
+            Args:
+                name (str): Name that will be set for the span being created
+                attributes(Optional[dict]):
```
sjbrunst, line 49:
```suggestion
                attributes (Optional[dict]):
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+                    Additional attributes that pertain to
+                    the specific API call (i.e. not a default attribute)
+                client (Optional[google.cloud.bigquery.client.Client]):
+                    Pass in a Client object to extract any attributes that may be
+                    relevant to it and add them to the created spans.
+                job_ref(Optional[google.cloud.bigquery.job._AsyncJob])
```
sjbrunst, line 55:
```suggestion
                job_ref (Optional[google.cloud.bigquery.job._AsyncJob]):
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+            if error.code is not None:
+                span.set_status(Status(http_status_to_canonical_code(error.code)))
+            raise
+
+
+def _get_final_span_attributes(attributes=None, client=None, job_ref=None):
+    if client:
+        client_attributes = _set_client_attributes(client)
+        _default_attributes.update(client_attributes)
+    elif job_ref:
```
sjbrunst, line 88: Would it be possible for someone to pass in both `client` and `job_ref`? (Should this just be `if` instead of `elif`?)

noxfile.py
```diff
@@ -48,6 +48,14 @@ def default(session):
     else:
         session.install("ipython")
 
+    # opentelemetry was not added to [all] because opentelemetry does not support Python 2
```
sjbrunst, line 51:
```suggestion
    # opentelemetry was not added to [all] because opentelemetry does not support Python 2.
```

noxfile.py
```diff
@@ -48,6 +48,14 @@ def default(session):
...
+    # exporter does not need to be in nox thus it has been added to README documentation
```
sjbrunst, line 52:
```suggestion
    # Exporter does not need to be in nox thus it has been added to README documentation.
```

noxfile.py
```diff
@@ -48,6 +48,14 @@ def default(session):
...
+    if session.python != "2.7":
```
sjbrunst, line 53: Could there be any other 2.something versions we need to avoid?

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,101 @@
...
+def setup():
+    tracer_provider = TracerProvider()
+    memory_exporter = InMemorySpanExporter()
+    span_processor = SimpleExportSpanProcessor(memory_exporter)
+    tracer_provider.add_span_processor(span_processor)
+    trace.set_tracer_provider(tracer_provider)
+    yield memory_exporter
+
+
+@pytest.mark.skipif(opentelemetry is None, reason="Require `opentelemetry`")
```
sjbrunst, line 36: Not really familiar with pytest, but is opentelemetry really required for a test that checks that things work if opentelemetry is not installed?

tests/unit/test_client.py
```diff
@@ -1477,9 +1652,16 @@ def test_create_table_w_fully_qualified_string(self):
...
-
-        got = client.create_table(
-            "{}.{}.{}".format(self.PROJECT, self.DS_ID, self.TABLE_ID)
+        with mock.patch(
+            "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+        ) as final_attributes:
+            got = client.create_table(
+                "{}.{}.{}".format(self.PROJECT, self.DS_ID, self.TABLE_ID)
+            )
+        final_attributes.assert_called_once_with(
```
sjbrunst, line 1661: style nit: most of the other tests have a blank line before `final_attributes.assert_called_once_with`. Please be consistent.

tests/unit/test_job.py
```diff
@@ -620,15 +620,16 @@ def test__begin_defaults(self):
...
         call_api = job._client._call_api = mock.Mock()
         call_api.return_value = resource
+        path = "/projects/{}/jobs".format(self.PROJECT)
+        with mock.patch(
+            "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+        ) as final_attributes:
+            job._begin()
 
-        job._begin()
+        final_attributes.assert_called_with({"path": path}, None, job)
```
sjbrunst, line 629: Can we make this test more restrictive, or are other spans generated by this call too? (Same comment on other tests.)
```suggestion
        final_attributes.assert_called_once_with({"path": path}, None, job)
```

aravinsiva: Yes multiple spans are created by this call. Which is why for these ones I just asserted a call at some point.

tests/unit/test_opentelemetry_tracing.py
aravinsiva: I believe there was an issue with this when Python 2.7 was run against all the tests in the pipeline. But I'll double check and remove it if we can run it against Python 2.7.

noxfile.py
aravinsiva: I believe BigQuery only supports Python 2.7 (In terms of 2. versioning). Similar logic is used when installing IPython.

google/cloud/bigquery/opentelemetry_tracing.py
aravinsiva: Should not be possible since I instrumented this across two different modules. The only calls being made send a job_ref or client exclusively.

noxfile.py
tswast: That's right. And there will never be a 2.8, so this should last until BigQuery drops Python 2.

google/cloud/bigquery/client.py
aravinsiva: We can type cast to a str so it can still be viewed in the UI.

google/cloud/bigquery/client.py
aravinsiva: Has to be called in the return statement. However this does not actually call the API and I am fixing that issue in another PR:
https://github.com/googleapis/python-bigquery/pull/228

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2020 Google LLC All rights reserved.
```
tswast, line 1: No "All rights reserved" in recommended header https://opensource.google/docs/releasing/preparing/#Apache-header

```suggestion
# Copyright 2020 Google LLC
```

google/cloud/bigquery/opentelemetry_tracing.py
tswast: Since the job classes do hold a client reference, I think it's reasonable to assume someone may think to set both. `create_span` could error in this case, but I think allowing it is preferable.

If both were passed in, I think preferring the `job_ref` properties is the right thing to do (that is, updating after the client attributes are set).

```suggestion
    if job_ref:
```

google/cloud/bigquery/client.py
tswast: Since these spans aren't actually wrapping the "real" request, could you remove these changes in this PR? I don't want someone to think that' telemetry is actually implemented for list calls.

google/cloud/bigquery/client.py
```diff
@@ -515,9 +538,13 @@ def create_routine(
         )
         resource = routine.to_api_repr()
         try:
-            api_response = self._call_api(
-                retry, method="POST", path=path, data=resource, timeout=timeout
-            )
+            span_attributes = {"path": path, "exists_ok": exists_ok}
```
tswast, line 541: Does it make sense to pass in `exists_ok` as an attribute if it's not passed to the actual API? It's only affecting how the client deals with certain response codes.

google/cloud/bigquery/client.py
tswast: Do you mean like the following?

```suggestion
        span_attributes = {"path": path, "body": str(body)}
```

Might make sense to use a real JSON encoder.

```suggestion
        span_attributes = {"path": path, "body": json.dumps(body)}
```

tswast: Body feels like it could be too much data to send as part of a span for most API requests (though probably not in this case). What's the way to decide what should go into an attribute and what shouldn't?

google/cloud/bigquery/client.py
```diff
@@ -890,14 +961,20 @@ def update_routine(self, routine, fields, retry=DEFAULT_RETRY, timeout=None):
...
-        api_response = self._call_api(
-            retry,
-            method="PUT",
-            path=routine.path,
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+        path = routine.path
+        span_attributes = {"path": path, "fields": fields}
```
tswast, line 965: fields is a list of strings. Serialize to JSON string?

google/cloud/bigquery/client.py
```diff
@@ -1183,17 +1295,22 @@ def delete_dataset(
...
 
         try:
-            self._call_api(
-                retry,
-                method="DELETE",
-                path=dataset.path,
-                query_params=params,
-                timeout=timeout,
-            )
+            span_attributes = {"path": path}
```
tswast, line 1303: Seems like we would want the value of `deleteContents` here, at least if it's set.

google/cloud/bigquery/client.py
aravinsiva: It was all mentioned in the design doc. I just went based off of information I felt was important. If you see I that I am missing some fields or some fields need not be added then I'd be happy to make those changes.

google/cloud/bigquery/client.py
aravinsiva: makes sense I can remove this.

google/cloud/bigquery/client.py
aravinsiva: Will do

google/cloud/bigquery/client.py
aravinsiva: will do

google/cloud/bigquery/client.py
aravinsiva: Sure

tests/unit/test_opentelemetry_tracing.py
aravinsiva: Can confirm there is an issue with python 2.7 in this test.

google/cloud/bigquery/client.py
```diff
@@ -793,14 +841,20 @@ def update_dataset(self, dataset, fields, retry=DEFAULT_RETRY, timeout=None):
...
-        api_response = self._call_api(
-            retry,
-            method="PATCH",
-            path=dataset.path,
-            data=partial,
-            headers=headers,
-            timeout=timeout,
-        )
+        path = dataset.path
+        span_attributes = {"path": path, "fields": json.dumps(fields)}
```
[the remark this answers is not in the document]
aravinsiva: This actually was a suggested change by Tim. But I suppose it makes the most sense to leave it at `fields` if it is accepted by OT.

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            yield span
+            span.set_status(Status(http_status_to_canonical_code(200)))
+        except GoogleAPICallError as error:
+            if error.code is not None:
+                span.set_status(Status(http_status_to_canonical_code(error.code)))
+            raise
+
+
+def _get_final_span_attributes(attributes=None, client=None, job_ref=None):
```
kintel, line 84: I don't quite understand this function. It's called every time a span is created, and it seems to mutate the _default_attributes dict. It feels like the intention was to either run it once per client, or to read from the _default_attributes dict from every span?

aravinsiva: This function was made to be called for every span and update the span to contain whatever client & job attributes are needed. The reason for this was for testing. Mocking a wrapped context manager function was problematic so by moving this logic it could be run for every span and mocked accordingly.

tests/unit/test_client.py
```diff
@@ -2616,8 +2971,8 @@ def test_list_routines_defaults(self):
         client = self._make_one(project=project_id, credentials=creds)
         conn = client._connection = make_connection(resource)
         dataset = DatasetReference(client.project, dataset_id)
-
         iterator = client.list_routines(dataset)
+
```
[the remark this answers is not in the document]
aravinsiva: My apologies. May have been my IDE autolinting for some reason.

setup.py
```diff
@@ -62,6 +62,7 @@
         "llvmlite<=0.34.0;python_version>='3.6'",
         "llvmlite<=0.31.0;python_version<'3.6'",
     ],
+    "frozendict": ["frozendict>=1.2"],
```
tswast, line 65: This appears to be a required dependency, not an "extras". Import of the module will fail without this installed.

It appears that its license is compatible with Apache 2.0 https://github.com/slezica/python-frozendict but is there any way we can avoid taking on this extra dependency?

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _set_client_attributes(client):
+    return {"db.name": client.project, "location": client.location}
+
+
+def _set_job_attributes(job_ref):
+    return {
+        "db.name": job_ref.project,
+        "location": job_ref.location,
+        "num_child_jobs": str(job_ref.num_child_jobs),
```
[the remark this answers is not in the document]
aravinsiva: seems to be no need to stringify. your right.

setup.py
aravinsiva: I can just initialize this inside of the create_span() method. Or just make a copy inside of these methods. However a copy of default attributes will need to be created with every span. Which shouldn't affect performance too much.

tests/unit/test_client.py
```diff
@@ -612,8 +628,12 @@ def test_get_dataset(self):
         }
         conn = client._connection = make_connection(resource)
         dataset_ref = DatasetReference(self.PROJECT, self.DS_ID)
+        with mock.patch(
+            "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+        ) as final_attributes:
+            dataset = client.get_dataset(dataset_ref, timeout=7.5)
 
-        dataset = client.get_dataset(dataset_ref, timeout=7.5)
+        final_attributes.assert_called_once_with({"path": "/%s" % path}, client, None)
```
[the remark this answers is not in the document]
aravinsiva: will address in next commit

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+except ImportError:
+    opentelemetry = None
+
+from six.moves import reload_module
+
+TEST_SPAN_NAME = "bar"
+TEST_SPAN_ATTRIBUTES = {"foo": "baz"}
+
+
```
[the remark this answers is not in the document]
aravinsiva: will address in next commit

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,114 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import sys
+import pytest
+
+import mock
+
+from google.cloud.bigquery import opentelemetry_tracing
```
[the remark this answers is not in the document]
aravinsiva: will address in next commit

setup.py
kintel: I like frozen data structures as a guard against accidentally mutating them, but I agree that adding a dependency may not be worth it.
Note: While you need a copy, you don't need to explicitly call copy(), you could clean up the dict handling into smth. a bit more readable, e.g.:
```
final_attributes = {
  **_default_attributes,
  **client_attributes,
  **job_attributes,
  **attributes,
}
```

tests/unit/test_client.py
```diff
@@ -89,7 +89,6 @@ def _make_list_partitons_meta_info(project, dataset_id, table_id, num_rows=0):
 
 
 class TestClient(unittest.TestCase):
-
```
sjbrunst, line 92: Unless the linter is complaining, don't change whitespace this far from your other changes

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,215 @@
...
+        "db.system": "BigQuery",
+        "db.name": "test_project_id",
+        "location": "test_location",
+        "num_child_jobs": "0",
+        "job_id": "test_job_id",
+        "foo": "baz",
+        "parent_job_id": "parent_job_id",
+        "timeCreated": time_created.strftime("%d-%b-%Y (%H:%M:%S.%f)"),
+        "timeStarted": started_time.strftime("%d-%b-%Y (%H:%M:%S.%f)"),
+        "timeEnded": ended_time.strftime("%d-%b-%Y (%H:%M:%S.%f)"),
```
sjbrunst, line 120: Do the attributes really use dd-mmm-yyyy formatting instead of [ISO 8601](https://xkcd.com/1179/)? 😢

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,128 @@
...
+        "job_id": job_ref.job_id,
+        "parent_job_id": job_ref.parent_job_id,
+        "state": job_ref.state,
+    }
+
+    if job_ref.error_result is not None:
+        job_attributes["hasErrors"] = True
+
+    else:
+        job_attributes["hasErrors"] = False
```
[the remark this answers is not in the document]
aravinsiva: great point will fix.

README.rst
```diff
@@ -102,3 +102,41 @@ Perform a query
...
+
+After installation, OpenTelemetry can be used in the BigQuery
+client and in BigQuery jobs. First, however, an exporter must be
+specified for where the trace data will be outputted to. An example of this
```
sjbrunst, line 121: nit: inconsistent line length breaks

google/cloud/bigquery/client.py
```diff
@@ -1231,8 +1309,13 @@ def delete_model(
         if not isinstance(model, (Model, ModelReference)):
             raise TypeError("model must be a Model or a ModelReference")
 
+        path = model.path
         try:
-            self._call_api(retry, method="DELETE", path=model.path, timeout=timeout)
+            span_attributes = {"not_found_okay": not_found_ok, "path": path}
```
tswast, line 1314: `not_found_okay` only affects client-side logic, so I don't think it makes sense to include in the trace

```suggestion
            span_attributes = {"path": path}
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,122 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from contextlib import contextmanager
+from google.api_core.exceptions import GoogleAPICallError
+
+Logger = logging.getLogger(__name__)
```
tswast, line 19: Lowercase for variable names. Capitalized is reserved for class names.

```suggestion
logger = logging.getLogger(__name__)
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,122 @@
...
+try:
+    from opentelemetry import trace
+    from opentelemetry.instrumentation.utils import http_status_to_canonical_code
+    from opentelemetry.trace.status import Status
+
+    HAS_OPENTELEMETRY = True
+
+except ImportError:
+    Logger.info(
```
tswast, line 28:
```suggestion
    logger.info(
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -0,0 +1,122 @@
...
+    final_attributes = _get_final_span_attributes(attributes, client, job_ref)
+    if not HAS_OPENTELEMETRY:
+        yield None
+        return
+    tracer = trace.get_tracer(__name__)
+
+    # yield new span value
+    with tracer.start_as_current_span(name=name, attributes=final_attributes) as span:
+        try:
+            yield span
```
tswast, line 76: I'm a bit confused as to how this line could throw an exception. Could you add a unit test on client or job to verify the span status in this case?

tests/unit/test_client.py
```diff
@@ -1132,11 +1225,19 @@ def test_create_routine_w_conflict(self):
...
         routine = Routine(full_routine_id)
 
         with pytest.raises(google.api_core.exceptions.AlreadyExists):
-            client.create_routine(routine)
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.create_routine(routine)
+
+            final_attributes.assert_called_once_with(  # pragma: NO COVER
```
tswast, line 1238: I think you should be able to cover this line if outside of the `pytest.raises` block.

```suggestion
        final_attributes.assert_called_once_with(
```

tswast: Also, this might be a good test to verify that the status code is set as expected.

tests/unit/test_client.py
```diff
@@ -1902,7 +2113,14 @@ def test_set_iam_policy_invalid_policy(self):
         client = self._make_one(project=self.PROJECT, credentials=creds, _http=http)
 
         with self.assertRaises(TypeError):
-            client.set_iam_policy(self.TABLE_REF, invalid_policy_repr)
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.set_iam_policy(self.TABLE_REF, invalid_policy_repr)
+
+            final_attributes.assert_called_once_with(  # pragma: NO COVER
```
tswast, line 2121: Likewise, this line should be covered if outsized of the the `assertRaises` block.

```suggestion
        final_attributes.assert_called_once_with(
```

tests/unit/test_client.py
```diff
@@ -1920,7 +2138,14 @@ def test_set_iam_policy_w_invalid_table(self):
         )
 
         with self.assertRaises(TypeError):
-            client.set_iam_policy(table_resource_string, policy)
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.set_iam_policy(table_resource_string, policy)
+
+            final_attributes.assert_called_once_with(  # pragma: NO COVER
```
tswast, line 2146:
```suggestion
       final_attributes.assert_called_once_with(
```

google/cloud/bigquery/opentelemetry_tracing.py
aravinsiva: Added a test that throws the exception in `test_opentelemetry_tracing.py`. Unless you were referring to a different type of test. That test raises an exception within the context of the span.

tests/unit/test_opentelemetry_tracing.py
```diff
@@ -0,0 +1,215 @@
...
+        project="test_project", credentials=mock_credentials, location="test_location"
+    )
+
+    expected_attributes = {
+        "foo": "baz",
+        "db.system": "BigQuery",
+        "db.name": "test_project",
+        "location": "test_location",
+    }
+    with unittest.TestCase().assertRaises(GoogleAPICallError):
```
tswast, line 209: Could you also check that the status code is set on the span?

README.rst
tswast: I'd still prefer if these dependencies were defined in `extras`. You can add a line to exclude from `all` since it's not supported on Python 2.7.

noxfile.py
```diff
@@ -48,6 +48,14 @@ def default(session):
     else:
         session.install("ipython")
 
+    # opentelemetry was not added to [all] because opentelemetry does not support Python 2.
+    # Exporter does not need to be in nox thus it has been added to README documentation
+    if session.python != "2.7":
+        session.install(
```
tswast, line 54: It's important to include `opentelemetry` somewhere in `setup.py` so that the dependency resolver can check if a compatible version is installed.

```suggestion
        session.install("-e", ".[opentelemetry]")
```

tests/unit/test_client.py
```diff
@@ -246,20 +256,27 @@ def test__get_query_results_miss_w_explicit_project_and_timeout(self):
...
+                    None,
+                    project="other-project",
+                    location=self.LOCATION,
+                    timeout_ms=500,
+                    timeout=42,
+                )
+
+            final_attributes.assert_called_once_with(
+                {"path": path}, client, None
+            )  # pragma: NO COVER
```
tswast, line 275: Remove `pragma: NO COVER` and move outside of the `assertRaises` block

tests/unit/test_client.py
```diff
@@ -625,35 +656,63 @@ def test_get_dataset(self):
...
         with self.assertRaises(ServerError):
-            client.get_dataset(dataset_ref)
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.get_dataset(dataset_ref)
+
+            final_attributes.assert_called_once_with(
+                {"path": path}, client, None
+            )  # pragma: NO COVER
```
tswast, line 688: Remove `pragma: NO COVER` and move outside of the `assertRaises` block

tests/unit/test_client.py
```diff
@@ -2817,7 +3192,14 @@ def test_delete_dataset_w_not_found_ok_false(self):
...
         with self.assertRaises(google.api_core.exceptions.NotFound):
-            client.delete_dataset(self.DS_ID)
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.delete_dataset(self.DS_ID)
+
+            final_attributes.assert_called_once_with(
+                {"path": path}, client, None
+            )  # pragma: NO COVER
```
tswast, line 3202: Remove `pragma: NO COVER` and move outside of the `assertRaises` block

tests/unit/test_client.py
```diff
@@ -2938,14 +3341,20 @@ def test_delete_routine_w_not_found_ok_false(self):
...
         with self.assertRaises(google.api_core.exceptions.NotFound):
-            client.delete_routine("routines-project.test_routines.test_routine")
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.delete_routine("routines-project.test_routines.test_routine")
+
+            final_attributes.assert_called_once_with(
+                {"path": path}, client, None
+            )  # pragma: NO COVER
```
tswast, line 3354: Remove `pragma: NO COVER` and move outside of the `assertRaises` block

tests/unit/test_client.py
```diff
@@ -3012,7 +3433,14 @@ def test_delete_table_w_not_found_ok_false(self):
...
         with self.assertRaises(google.api_core.exceptions.NotFound):
-            client.delete_table("{}.{}".format(self.DS_ID, self.TABLE_ID))
+            with mock.patch(
+                "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
+            ) as final_attributes:
+                client.delete_table("{}.{}".format(self.DS_ID, self.TABLE_ID))
+
+            final_attributes.assert_called_once_with(
+                {"path": path}, client, None
+            )  # pragma: NO COVER
```
tswast, line 3443: Remove `pragma: NO COVER` and move outside of the `assertRaises` block

tests/unit/test_opentelemetry_tracing.py
aravinsiva: This test just asserts the exception was thrown there is another test that I wrote that verify's span status. Seen here:
https://github.com/googleapis/python-bigquery/pull/215/files#diff-87020ebeb7f433bac8fc936494fe2c75R1255

setup.py
```diff
@@ -62,12 +62,17 @@
...
 }
 
 all_extras = []
 
 for extra in extras:
-    if extra == "fastparquet":
+    if extra == "fastparquet" or extra == "opentelemetry":
         # Skip fastparquet from "all" because it is redundant with pyarrow and
         # creates a dependency on pre-release versions of numpy. See:
         # https://github.com/googleapis/google-cloud-python/issues/8549
```
tswast, line 78: Let's document why we're excluding `opentelemetry` so that it can be added back to `all` after this package drops Python 2 support.

```suggestion
    if extra in (
        # fastparquet is redundant with pyarrow and creates a dependency
        # on pre-release versions of numpy. See:
        # https://github.com/googleapis/google-cloud-python/issues/8549
        "fastparquet",
        # opentelemtry is not compatible with Python 2.
        "opentelemetry",
    ):
```

PR #219, opened 2020-08-10
tests: split out snippets builds

tseaver:
@tmatsuo Emulating PR #207. I don't know if I'm missing anything:  e.g., I don't quite understand what the `split_system_tests=True` does in the `synth.py` there.

Toward #191


.kokoro/presubmit/presubmit.cfg
```diff
@@ -5,3 +5,7 @@ env_vars: {
     key: "RUN_SYSTEM_TESTS"
     value: "false"
 }
+env_vars: {
+    key: "RUN_SNIPPETS"
```
tmatsuo, line 9: I think we need to allow this envvar in the Kokoro config in google3.
Also how about `RUN_SNIPPETS_TESTS` for consistensy?

tmatsuo (commented):
Sorry for the late reply. Once we agree on the env var name, I can make a change for allowing the env var.

.kokoro/presubmit/presubmit.cfg
tswast: `RUN_SNIPPETS_TESTS` sounds good to me

tswast (approved):

tmatsuo (approved):

Issue #221, opened 2020-08-12
Add tests in client & job for Opentelemetry Instrumentation


aravinsiva:
In order to complete the instrumentation in modular stable PR's these tests are added separate from the PR with the instrumentation. Task it to add tests that assert the calling of the opentelemetry_tracing module when different API calls are made. These tests that need to be modified exist in the test_client.py module & test_job.py module. 

**Please assign to me**

PR #239, opened 2020-08-22
feat: add instrumentation to list methods

aravinsiva:
PR adds opentelemetry instrumentation to all methods that don't directly make api calls and instead use an HTTPIterator.,

Fixes #<https://github.com/googleapis/python-bigquery/issues/221> 🦕


tests/unit/test_client.py
```diff
@@ -425,9 +425,14 @@ def test_list_projects_defaults(self):
         creds = _make_credentials()
         client = self._make_one(PROJECT_1, creds)
         conn = client._connection = make_connection(DATA)
-
         iterator = client.list_projects()
-        page = six.next(iterator.pages)
+
+        with mock.patch(
+            "google.cloud.bigquery.opentelemetry_tracing._get_final_span_attributes"
```
sjbrunst, line 431: Optional comment: this is a really long string that is repeated many times in this file. It's probably a good idea to put it in a const string in your test client.

Since that would affect a bunch of other tests in this file, it would be best to do in a separate PR.

Issue #240, opened 2020-08-24
Fastavro deprecated support for Python 2.7 breaking build


aravinsiva:
Fix the version of fastavro to a version that supports Python 2.7 that won't break the nox pipeline. Although there may be an upstream fix on this it would not hurt to fix this on our end as well.

HemangChothani:
Closing this issue  as resolved.

PR #241, opened 2020-08-24
fix: fix dependency issue in fastavro

aravinsiva:
Fixes #<https://github.com/googleapis/python-bigquery/issues/240> 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -92,7 +92,6 @@ def system(session):
     # Install all test dependencies, then install local packages in place.
     session.install("mock", "pytest", "psutil", "google-cloud-testutils")
     session.install("google-cloud-storage")
-    session.install("fastavro==0.24.2")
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -92,7 +92,7 @@ def system(session):
     # Install all test dependencies, then install local packages in place.
     session.install("mock", "pytest", "psutil", "google-cloud-testutils")
     session.install("google-cloud-storage")
-    session.install("fastavro")
+    session.install("fastavro==0.24.2")
```
tswast, line 95: Actually, I think we might not need `fastavro` anymore. We switched to Arrow as the serialization format a while ago. What happens if we just remove this line instead?

```suggestion
```

aravinsiva: yeah seems as though we don't need it. Ran through tests everything seems to be working fine

PR #244, opened 2020-08-26
refactor: restructing instrumentation format

aravinsiva:
Reformatting instrumentation to a cleaner implementation.

Fixes #<https://github.com/googleapis/python-bigquery/issues/245> 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -252,7 +252,6 @@ class Client(ClientWithProject):
             retry,
             name="BigQuery.getServiceAccountEmail",
             span_attributes=span_attributes,
-            client=self,
             method="GET",
             path=path,
             timeout=timeout,
@@ -297,7 +296,6 @@ class Client(ClientWithProject):
                 retry,
                 name="BigQuery.listProjects",
                 span_attributes=span_attributes,
-                client=self,
                 *args,
                 timeout=timeout,
                 **kwargs
@@ -376,7 +374,6 @@ class Client(ClientWithProject):
                 retry,
                 name="BigQuery.listDatasets",
                 span_attributes=span_attributes,
-                client=self,
                 *args,
                 timeout=timeout,
                 **kwargs
@@ -514,7 +511,6 @@ class Client(ClientWithProject):
                 retry,
                 name="BigQuery.createDataset",
                 span_attributes=span_attributes,
-                client=self,
                 method="POST",
                 path=path,
                 data=data,
@@ -566,7 +562,6 @@ class Client(ClientWithProject):
                 retry,
                 name="BigQuery.createRoutine",
                 span_attributes=span_attributes,
-                client=self,
                 method="POST",
                 path=path,
                 data=resource,
@@ -621,7 +616,6 @@ class Client(ClientWithProject):
                 retry,
                 name="BigQuery.createTable",
                 span_attributes=span_attributes,
-                client=self,
                 method="POST",
                 path=path,
                 data=data,
@@ -638,16 +632,18 @@ class Client(ClientWithProject):
         retry,
         name=None,
         span_attributes=None,
-        client=None,
         job_ref=None,
         **kwargs
     ):
-        with create_span(
-            name=name, attributes=span_attributes, client=client, job_ref=job_ref
-        ):
-            call = functools.partial(self._connection.api_request, **kwargs)
+
+        call = functools.partial(self._connection.api_request, **kwargs)
         if retry:
             call = retry(call)
+        if name is not None:
+            with create_span(
+                name=name, attributes=span_attributes, client=self, job_ref=job_ref
+            ):
+                return call()
         return call()
 
     def get_dataset(self, dataset_ref, retry=DEFAULT_RETRY, timeout=None):
@@ -682,7 +678,6 @@ class Client(ClientWithProject):
             retry,
             name="BigQuery.getDataset",
             span_attributes=span_attributes,
-            client=self,
             method="GET",
             path=path,
             timeout=timeout,
@@ -710,7 +705,6 @@ class Client(ClientWithProject):
             retry,
             name="BigQuery.getIamPolicy",
             span_attributes=span_attributes,
-            client=self,
             method="POST",
             path=path,
             data=body,
@@ -745,7 +739,6 @@ class Client(ClientWithProject):
             retry,
             name="BigQuery.setIamPolicy",
             span_attributes=span_attributes,
-            client=self,
             method="POST",
             path=path,
             data=body,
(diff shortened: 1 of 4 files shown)


google/cloud/bigquery/client.py
```diff
@@ -596,20 +617,35 @@ def create_table(self, table, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
...
             if not exists_ok:
                 raise
             return self.get_table(table.reference, retry=retry)
 
-    def _call_api(self, retry, **kwargs):
-        call = functools.partial(self._connection.api_request, **kwargs)
+    def _call_api(
+        self,
+        retry,
+        name=None,
```
sjbrunst, line 639: Should this parameter be called `span_name`? Since it's only used to create the span, the parameter should have name that makes its purpose more clear.

google/cloud/bigquery/job.py
```diff
@@ -34,7 +34,6 @@
 from google.cloud.bigquery.encryption_configuration import EncryptionConfiguration
 from google.cloud.bigquery.external_config import ExternalConfig
 from google.cloud.bigquery.external_config import HivePartitioningOptions
-from google.cloud.bigquery.opentelemetry_tracing import create_span
```
sjbrunst, line 37: Can we remove this import in client.py too? (GitHub doesn't let me comment too far from the diff)

tests/unit/test_job.py
```diff
@@ -2744,7 +2734,10 @@ def test_cancel_w_bound_client(self):
         final_attributes.assert_called_with({"path": PATH}, None, job)
 
         conn.api_request.assert_called_once_with(
-            method="POST", path=PATH, query_params={}, timeout=None,
+            method="POST",
+            path=PATH,
+            query_params={},
+            timeout=None,
```
sjbrunst, line 2740: (this is probably your IDE making changes): don't make whitespace-only changes too far from your main changes. It makes the file history more difficult to read and makes this PR longer. (If you want to clean up the whitespace, I'd suggest making a separate whitespace-only PR for easy reviewing.)

google/cloud/bigquery/client.py
```diff
@@ -596,20 +617,35 @@ def create_table(self, table, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
...
+        span_attributes=None,
+        client=None,
+        job_ref=None,
+        **kwargs
+    ):
+        with create_span(
+            name=name, attributes=span_attributes, client=client, job_ref=job_ref
```
sjbrunst, line 646: If `name` is `None`, will this span creation still work? It's a non-optional parameter to `create_span`: https://github.com/googleapis/python-bigquery/blob/fa9f9ca491c3f9954287102c567ec483aa6151d4/google/cloud/bigquery/opentelemetry_tracing.py#L43

aravinsiva: good point I will need to add some logic for that.

google/cloud/bigquery/client.py
tswast: +1 on calling this `span_name`, since it isn't used to actually make the API call

tests/unit/test_job.py
tswast: I suspect we may need to pin our version of `black`, as different versions can produce different formatting output.

google/cloud/bigquery/client.py
```diff
@@ -596,20 +617,35 @@ def create_table(self, table, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
...
             return self.get_table(table.reference, retry=retry)
 
-    def _call_api(self, retry, **kwargs):
-        call = functools.partial(self._connection.api_request, **kwargs)
+    def _call_api(
+        self,
+        retry,
+        name=None,
+        span_attributes=None,
+        client=None,
```
tswast, line 641: Client seems to be an unnecessary argument, as I don't think we'll be calling this with anything but `self`.

google/cloud/bigquery/client.py
```diff
@@ -596,20 +617,35 @@ def create_table(self, table, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
...
+        job_ref=None,
+        **kwargs
+    ):
+        with create_span(
+            name=name, attributes=span_attributes, client=client, job_ref=job_ref
+        ):
+            call = functools.partial(self._connection.api_request, **kwargs)
```
tswast, line 648: This isn't the right place for `create_span`. It should be below at the `return call()` line.

tests/unit/test_job.py
aravinsiva: Yes I think has been generating new changes to other files. So pining this version is definitely best.

tswast: Mailed https://github.com/googleapis/python-bigquery/pull/247

tswast:
Nit on commit name: `feat: restructing instrumentation format` implies this is a new feature, but it's actually a refactoring of an existing feature. `refactor: ` would be better since this doesn't actually change any behavior. See https://github.com/googleapis/release-please/blob/46d9d834abd6d11000476c6938b1b0e897c89b14/src/releasers/python.ts#L29-L42 for list of types we use.

google/cloud/bigquery/client.py
aravinsiva: thanks for clarification.

tests/unit/test_job.py
aravinsiva: waiting until #247 is merged to fix this. As black is fixing lines irrelevant to change.

google/cloud/bigquery/job.py
aravinsiva: Can't remove from client.py because it is used in the _call_api method in the module

google/cloud/bigquery/client.py
```diff
@@ -596,22 +612,33 @@ def create_table(self, table, exists_ok=False, retry=DEFAULT_RETRY, timeout=None
...
+
         call = functools.partial(self._connection.api_request, **kwargs)
         if retry:
             call = retry(call)
+        if span_name is not None:
+            with create_span(
+                name=span_name, attributes=span_attributes, client=self, job_ref=job_ref
+            ):
+                return call()
         return call()
```
tswast, line 642: Could you add a unit test just for `_call_api` itself to check this code branch? It can verify that `create_span` is not called if `span_name` is not specified.

aravinsiva: test's added in `test_client.py` seen on line 253 -> 299

tests/unit/test_client.py
```diff
@@ -250,6 +250,54 @@ def test__call_api_applying_custom_retry_on_timeout(self):
             [mock.call(foo="bar"), mock.call(foo="bar")],  # was retried once
         )
 
+    def test__call_api_span_creator_called(self):
```
tswast, line 253: Looks like this and the other test should exchange names. This is testing that it's not called.

PR #247, opened 2020-08-27
chore: use specific version for black

### LINKED ISSUE #248: Pin black to specific version
New releases of `black` can change the way the code is formatted ("blackened"), causing the CI checks to fail. We should thus use a deterministic version of `black`.

tswast:
Version matches https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2, set in https://github.com/googleapis/synthtool/pull/643

Closes #248.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -20,6 +20,7 @@ import shutil
 import nox
 
 
+BLACK_VERSION = "black==19.10b0"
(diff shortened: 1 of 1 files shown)


plamut (approved):
Thanks for this, I just started noticing black failures, probably because of a new release.

BTW, could you please just open and link an issue for this so that it ends up in the changelog? Thanks!

PR #250, opened 2020-08-28
chore: bump pyarrow to 1.0.0

### LINKED ISSUE #249: RowIterator to_dataframe requires pyarrow >= 1.0.0 to work
Currently the google-cloud-bigquery library requires pyarrow > 0.16.0, however the method RowIterator.to_dataframe adds the kwarg "timestamp_as_object", which is only supported in pyarrow >= 1.0.0. If install pyarrow >= 1.0.0, everything works as expected, however we are using other libraries which require pyarrow < 1.0.0.

So the requirements should either be updated to require pyarrow >= 1.0.0, or backported to support versions less than 1.

#### Environment details

  - OS type and version: Any
  - Python version: 3.6.9
  - pip version: 20.2.2
  - `google-cloud-bigquery` version: 1.27.2

#### Steps to reproduce

  1. Use pyarrow < 1.0.0
  2. Run RowIterator to_dataframe

#### Stack trace
```
#     result = future.result()
  File "<path>/python3.6/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "<path>/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "<path>/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "<path>", line 133, in run_query
    bqstorage_client=client_storage
  File "<path>/python3.6/site-packages/google/cloud/bigquery/table.py", line 1757, in to_dataframe
    df = record_batch.to_pandas(date_as_object=date_as_object, **extra_kwargs)
  File "pyarrow/array.pxi", line 503, in pyarrow.lib._PandasConvertible.to_pandas
TypeError: to_pandas() got an unexpected keyword argument 'timestamp_as_object'
```

HemangChothani:
Fixes #249

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ extras = {
         # as `google-api-core[grpc]`. We thus need to explicitly specify it here.
         # See: https://github.com/googleapis/python-bigquery/issues/83
         "grpcio >= 1.8.2, < 2.0dev",
-        "pyarrow >= 1.0.0, < 2.0dev",
+        "pyarrow >= 1.0.0, < 2.0dev; python_version >= '3.5'",
     ],
     "pandas": ["pandas>=0.17.1"],
     # Exclude PyArrow dependency from Windows Python 2.7.
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ extras = {
         # as `google-api-core[grpc]`. We thus need to explicitly specify it here.
         # See: https://github.com/googleapis/python-bigquery/issues/83
         "grpcio >= 1.8.2, < 2.0dev",
-        "pyarrow>=0.16.0, < 2.0dev",
+        "pyarrow >= 1.0.0, < 2.0dev",
     ],
     "pandas": ["pandas>=0.17.1"],
     # Exclude PyArrow dependency from Windows Python 2.7.


setup.py
```diff
@@ -43,7 +43,7 @@
         # as `google-api-core[grpc]`. We thus need to explicitly specify it here.
         # See: https://github.com/googleapis/python-bigquery/issues/83
         "grpcio >= 1.8.2, < 2.0dev",
-        "pyarrow>=0.16.0, < 2.0dev",
+        "pyarrow >= 1.0.0, < 2.0dev; python_version >= '3.5'",
```
plamut, line 46: Python 2 users will be likely stuck with `pyarrow` 0.16.0 (the last release compatible with Python 2), since bqstorage [still supports](https://github.com/googleapis/python-bigquery-storage/blob/master/setup.py#L33) that version.

IIRC that should still be fine, though, as we have a conditional check in the code to only use the `timestamp_as_object` parameter under Python 3 (that parameter is not supported in pyarrow versions compatible with Py2).

plamut (approved):

PR #212, opened 2020-08-03
refactor: update insert_rows.py for more efficient streaming

jmgreger:
Added in the use of selected_fields=schema to explicitly define the schema. This prevents excessive calling of table metadata when streaming large volumes of data.

There may be a better way to showcase this in the code, but I wanted to put this out there for review.

Fixes #211  🦕

PR #253, opened 2020-08-31
Update table_insert_rows.py

### LINKED ISSUE #211: Improve guidance for streaming inserts
Suggesting we update the insert_rows.py code to showcase how to explicitly define a schema to avoid excessive calls to get_table().

jmgreger:
Adding information to this example for more efficient streaming. Reduce calls to client.get_table()

This PR is meant to replace #212 due to issues from referencing a release initially. Comments and suggestions were incorporated into this PR. 

Fixes #211  🦕

### DIFF
--- a/samples/table_insert_rows.py
+++ b/samples/table_insert_rows.py
@@ -23,12 +23,28 @@ def table_insert_rows(table_id):
     client = bigquery.Client()
 
     # TODO(developer): Set table_id to the ID of the model to fetch.
-    # table_id = "your-project.your_dataset.your_table"
+    table_id = "your-project.your_dataset.your_table"
 
-    table = client.get_table(table_id)  # Make an API request.
+    # The client converts Python objects to JSON-serialization, but this requires a schema.
+    # A schema can be fetched by calling `client.get_table` as shown below.
+    # schema = client.get_table(table_id).schema
+    
+    # If inserting many rows, it is suggested that you cache the schema.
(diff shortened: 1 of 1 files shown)


samples/table_insert_rows.py
```diff
@@ -23,12 +23,26 @@ def table_insert_rows(table_id):
...
+        ]
+
+    # Populate data for entry
     rows_to_insert = [(u"Phred Phlyntstone", 32), (u"Wylma Phlyntstone", 29)]
 
-    errors = client.insert_rows(table, rows_to_insert)  # Make an API request.
-    if errors == []:
+    try:
+        # Stream data to BQ
+        client.insert_rows(table_id, selected_fields=schema, rows_to_insert)
```
plamut, line 46: Positional arguments cannot follow keyword arguments, please change the call.

plamut (changes requested):
On the surface the change seems sensible, although I cannot really comment on the wording in the comments, as I'm not a native English speaker.

Before merging, though, please make sure that the all the checks pass (linters, tests...), and that commit messages follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) guideline, thanks!

samples/table_insert_rows.py
tswast: Also, still need to check for errors. The API call returns success on parial failure, such as unable to insert a particular row.

PR #262, opened 2020-09-17
fix: validate job_config.source_format in load_table_from_dataframe

### LINKED ISSUE #261: Setting `job_config.source_format` is ignored by `load_table_from_dataframe`
Currently on PARQUET is supported. If someone sets the format to something other than that, we should probably raise a ValueError.

Might want to consider support CSV / JSON / AVRO, but also don't want to explode the number of (already numerous) code paths we have for DataFrame loading.

HemangChothani:
Fixes #261

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2174,7 +2174,15 @@ class Client(ClientWithProject):
         else:
             job_config = job.LoadJobConfig()
 
-        job_config.source_format = job.SourceFormat.PARQUET
+        if job_config.source_format:
+            if job_config.source_format is not job.SourceFormat.PARQUET:
+                raise ValueError(
+                    "Cannot pass `{}` as a ``source_format``, currently PARQUET is supported".format(
+                        job_config.source_format
+                    )
+                )
+        else:
+            job_config.source_format = job.SourceFormat.PARQUET
 
         if location is None:
             location = self.location
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2175,9 +2175,9 @@ class Client(ClientWithProject):
             job_config = job.LoadJobConfig()
 
         if job_config.source_format:
-            if job_config.source_format is not job.SourceFormat.PARQUET:
+            if job_config.source_format != job.SourceFormat.PARQUET:
                 raise ValueError(
-                    "Cannot pass `{}` as a ``source_format``, currently PARQUET is supported".format(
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/client.py
```diff
@@ -2174,7 +2174,15 @@ def load_table_from_dataframe(
         else:
             job_config = job.LoadJobConfig()
 
-        job_config.source_format = job.SourceFormat.PARQUET
+        if job_config.source_format:
+            if job_config.source_format is not job.SourceFormat.PARQUET:
+                raise ValueError(
+                    "Cannot pass `{}` as a ``source_format``, currently PARQUET is supported".format(
```
tswast, line 2180:
```suggestion
                    "Got unexpected source_format: '{}'. Currently, only PARQUET is supported".format(
```

tests/unit/test_client.py
```diff
@@ -7595,6 +7595,80 @@ def test_load_table_from_dataframe_w_custom_job_config(self):
...
+            write_disposition=job.WriteDisposition.WRITE_TRUNCATE,
+            source_format=job.SourceFormat.ORC,
+        )
+
+        with pytest.raises(ValueError) as exc:
+            client.load_table_from_dataframe(
+                dataframe, self.TABLE_REF, job_config=job_config, location=self.LOCATION
+            )
+
+        assert "Cannot pass" in str(exc.value)
```
tswast, line 7670:
```suggestion
        assert "Got unexpected source_format:" in str(exc.value)
```

tswast (commented):
Thanks for the quick response! A few suggested update to the error message.

google/cloud/bigquery/client.py
```diff
@@ -2174,7 +2174,15 @@ def load_table_from_dataframe(
         else:
             job_config = job.LoadJobConfig()
 
-        job_config.source_format = job.SourceFormat.PARQUET
+        if job_config.source_format:
+            if job_config.source_format is not job.SourceFormat.PARQUET:
```
plamut, line 2178: This (non)identity check is relying on interning strings, but we shouldn't rely on `"PARQUET"` strings to point (or not point) to the same string instance. Let's use the equality operator instead.

(`SourceFormat` is not an `enum`, it's just a plain class with a few string members)

plamut (changes requested):
Seems fine apart from that gotcha that might bite us at some future point.

tswast (approved):

PR #263, opened 2020-09-21
fix: update minimum dependency versions

tswast:
This PR updates the minimum dependency versions to match those that I
found to be actually runnable. Updates tests to use constraint files so
that at least one test session uses these minimum versions.

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!
  - Based on internal Python Client Library Testing Improvements docs.
  - In response to internal bug 166792569 covering insufficient pyarrow minimum dependency.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -51,7 +51,7 @@ def default(session):
 
     if session.python == "2.7":
         # The [all] extra is not installable on Python 2.7.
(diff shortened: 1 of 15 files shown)


shollyman (approved):
TIL about pytest.importorskip() 

Thanks for this!

PR #278, opened 2020-09-25
chore: Prepare for 2.0 release

plamut:
This PR should serve as a preview of the 2.0 release.

For a start, the BQ Storage `v1beta1` compatibility code is removed, as it will not be needed anymore once BQ Storage v2.0.0 is released.

**TODO list:**
- [x] Fix import paths in tests and samples that are  affected by `bigquery_v2` changes.
- [x] Describe `bigquery_v2` changes in UPGRADING guide.
- [x] Fix docs, include `bigquery_v2` in them, if necessary.


plamut:
BQ Storage 2.0.0 released, re-running.

google/cloud/bigquery_v2/services/__init__.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
tswast, line 1: This whole `services` directory (and sub-directories) won't actually function correctly, as there's not a public endpoint for the models service. Let's exclude these from this PR.

tests/unit/test_dbapi_connection.py
```diff
@@ -41,29 +41,26 @@ def _mock_client(self):
...
+        mock_client._transport.grpc_channel = mock.Mock(spec=["close"])
         return mock_client
 
     def test_ctor_wo_bqstorage_client(self):
         from google.cloud.bigquery.dbapi import Connection
 
         mock_client = self._mock_client()
-        mock_bqstorage_client = self._mock_bqstorage_client()
-        mock_client._create_bqstorage_client.return_value = mock_bqstorage_client
+        mock_client._create_bqstorage_client.return_value = None
```
tswast, line 55: Maybe create a sentinel `object()` here? I think we want to ensure that a BQ Storage client is created by default.

google/cloud/bigquery_v2/services/__init__.py
plamut: By "these", do you mean the models or everything that was re-generated under `bigquery_v2` with the microgenerator?

tests/unit/test_dbapi_connection.py
plamut: I'll see if a sentinel mock will work out of the box, because IIRC there is an "is None" check in the DB API code, thus we need to actually use `None`.

google/cloud/bigquery_v2/services/__init__.py
tswast: Just the `services` directory. We still need the `types`.

tests/unit/enums/test_standard_sql_data_types.py
```diff
@@ -59,15 +59,3 @@ def test_standard_sql_types_enum_members(enum_under_test, gapic_enum):
...
-def test_standard_sql_types_enum_docstring(enum_under_test, gapic_enum):
-    assert "STRUCT (int):" not in enum_under_test.__doc__
-    assert "BOOL (int):" in enum_under_test.__doc__
-    assert "TIME (int):" in enum_under_test.__doc__
-
-    # All lines in the docstring should actually come from the original docstring,
-    # except for the header.
-    assert "An Enum of scalar SQL types." in enum_under_test.__doc__
-    doc_lines = enum_under_test.__doc__.splitlines()
-    assert set(doc_lines[1:]) <= set(gapic_enum.__doc__.splitlines())
```
plamut, line 73: This was removed, because the new generated enum has no docstring, at least not at the time of writing.

plamut:
FWIW, the snippets tests in `samples/snippets` fail because the generated noxfile tries to install BigQuery v1.x, but the snippets require v2.0.0 that is still in the making. They pass locally if `INSTALL_LIBRARY_FROM_SOURCE` env variable is set and the  following patch is applied:

```diff
diff --git samples/snippets/noxfile.py samples/snippets/noxfile.py
index 5660f08..66589f5 100644
--- samples/snippets/noxfile.py
+++ samples/snippets/noxfile.py
@@ -158,14 +158,17 @@ def _session_tests(session, post_install=None):
         session.install("-r", "requirements-test.txt")
 
     if INSTALL_LIBRARY_FROM_SOURCE:
-        session.install("-e", _get_repo_root())
+        session.install("-e", _get_repo_root() + "[all]")
 
     if post_install:
         post_install(session)
 
diff --git samples/snippets/requirements.txt samples/snippets/requirements.txt
index 7fe8391..4753c47 100644
--- samples/snippets/requirements.txt
+++ samples/snippets/requirements.txt
@@ -1,4 +1,4 @@
-google-cloud-bigquery[pandas,bqstorage,pyarrow]==1.26.1
+# google-cloud-bigquery[pandas,bqstorage,pyarrow]==1.26.1
 google-auth-oauthlib==0.4.1
 ipython==7.16.1; python_version < '3.7'
 ipython==7.17.0; python_version >= '3.7'
```

tests/unit/enums/test_standard_sql_data_types.py
tswast: Bummer. We'll really want those at some point. Filed https://github.com/googleapis/gapic-generator-python/issues/625 against the microgenerator.

tswast:
> They pass locally if INSTALL_LIBRARY_FROM_SOURCE env variable is set and the following patch is applied:

Do we need to add the necessary extras to `requirements.txt`?

plamut:
@tswast Extras are needed, yes, otherwise the code complains that `pandas` is required (IIRC).

I did not check if each of the extras is actually needed, though.

tswast:
Since it looks like `requirements.txt` is used to install packages, even when `INSTALL_LIBRARY_FROM_SOURCE` is set, I'm a bit confused as to why pandas isn't getting installed.

https://github.com/googleapis/python-bigquery/blob/fbbe0cb0ea22161d81f1e5504bb89b55e4198634/samples/snippets/noxfile.py#L155

Maybe I missed some logic that filters out `google-cloud-bigquery[pandas,...]` from that installation?

plamut:
@tswast Ah, I think that's probably because I temporarily commented out the `google-cloud-python` line from `requirements.txt` completely, I just wanted to test with a local version of the library.

tests/unit/enums/test_standard_sql_data_types.py
plamut: Make sense. I will add ti back, but mark it as skipped. It will then be easy to resurrect it when the generator is fixed.

tests/unit/test_dbapi_connection.py
plamut: Checked the test, and the scenario it covers is when the `bqstorage` extra is not installed. The BigQuery client's factory method returns `None` in this case, hence the `None` return value of the mocked factory.

docs/bigquery_v2/types.rst
```diff
@@ -0,0 +1,5 @@
+Types for Google Cloud Bigquery v2 API
+======================================
+
+.. automodule:: google.cloud.bigquery_v2.types
```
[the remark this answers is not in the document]
plamut: Since this file is generated, I opted for the second option and configured docs build to ignore files in `docs/bigquery_v2`.

testing/constraints-3.6.txt
```diff
@@ -0,0 +1,16 @@
+fastparquet==0.4.1
```
plamut, line 1: The minimum is not specified, thus I used the current latest version, i.e. the one that gets installed right now by default.

testing/constraints-3.6.txt
```diff
@@ -0,0 +1,16 @@
...
+google-resumable-media==0.6.0
+grpcio==1.32.0
+ipython==5.5
+libcst==0.2.5
+llvmlite==0.34.0
+# pandas 0.23.0 is the first version to work with pyarrow to_pandas.
+pandas==0.23.0
+proto-plus==1.4.0
+pyarrow==1.0.0
+python-snappy==0.5.4
```
plamut, line 14: The minimum is not specified, thus I used the current latest version, i.e. the one that gets installed right now by default.

UPGRADING.md
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+
+# 2.0.0 Migration Guide
+
+The 2.0 release of the `google-cloud-bigquery` client drops support for Python
+versions below 3.6. The client surface itself has not changed, but the 1.x series
```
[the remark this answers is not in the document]
plamut: After re-generating `bigquery_v2` I noticed that `StandardSqlDataType` is not an enum anymore, thus references like `StandardSqlDataType.BOOL` do not work anymore. They now need to be referenced through the `TypeKind` enum, e.g. `StandardSqlDataType.TypeKind.BOOL`.

I will explain this to the guide.

docs/bigquery_v2/services.rst
```diff
@@ -0,0 +1,6 @@
+Services for Google Cloud Bigquery v2 API
+=========================================
+
+.. automodule:: google.cloud.bigquery_v2.services.model_service
```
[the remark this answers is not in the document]
plamut: Yes, pushed the change in the meantime - it's excluded in `docs/conf.py`, Sphinx ignores it (it does not complain about it not being included).

google/cloud/bigquery/model.py
```diff
@@ -55,7 +55,7 @@ class Model(object):
     def __init__(self, model_ref):
         # Use _proto on read-only properties to use it's built-in type
         # conversion.
-        self._proto = types.Model()
+        self._proto = types.Model()._pb
```
tswast, line 58: Could you explain why we're using `_pb`?

plamut: `json_format.ParseDict()` used in methods such as `from_api_repr()` needs a raw protobuf message to work. We get the raw message from a proto-plus wrapper class through the `_pb` attribute.

Unifying the internal representations to raw protobuf messages was needed to fix several then-failing unit tests, e.g. for `Rotuine` and `RoutineReference`.

It might be achievable to store the actual proto-plus wrapper instances, but we would then have to place conversions to the raw protobuf messages at quite a few places around the code (but I didn't go down that route, had to pick one).

tswast: Gotcha. There are a couple benefits to proto-plus (mostly around timestamp fields), but I think this is the pragmatic choice.

tswast (approved):

tswast (approved):

crwilcox:
Hammer merging to work around sample tests.

plamut:
@crwilcox Thanks!

Release PR (#284) will probably need the same hammer love.

PR #300, opened 2020-10-06
fix: broken create_job method

### LINKED ISSUE #297: Job retry solution (client.create_job) is broken
A solution was implemented in https://github.com/googleapis/python-bigquery/pull/71 but is now broken.

This is not caught by unit tests because the failure is within a mocked interface.

Original issue: https://github.com/googleapis/python-bigquery/issues/14

#### Environment details

  - Linux
  - Python 3.6, 3.7
  - Client version 1.25 - 2.0.0

#### Steps to reproduce

Run the code in the unit tests without mocks; rather than failing for nonexistent/unauthorized project/dataset (as you would expect), it fails with an AttributeError.:

#### Code example

```
configuration = {
            "load": {
                "destinationTable": {
                    "projectId": 'anyproject',
                    "datasetId": 'anydataset',
                    "tableId": "source_table",
                },
                "sourceUris": ["gs://test_bucket/src_object*"],
            }
        }

bqclient.create_job(configuration)
```

#### Stack trace

Fails with:

```
~/python/lib/python3.7/site-packages/google/cloud/bigquery/client.py in create_job(self, job_config, retry)
   1619             source_uris = _get_sub_prop(job_config, ["load", "sourceUris"])
   1620             return self.load_table_from_uri(
-> 1621                 source_uris, destination, job_config=load_job_config, retry=retry
   1622             )
   1623         elif "copy" in job_config:

~/python/lib/python3.7/site-packages/google/cloud/bigquery/client.py in load_table_from_uri(self, source_uris, destination, job_id, job_id_prefix, location, project, job_config, retry, timeout)
   1967 
   1968         load_job = job.LoadJob(job_ref, source_uris, destination, self, job_config)
-> 1969         load_job._begin(retry=retry, timeout=timeout)
   1970 
   1971         return load_job

~/python/lib/python3.7/site-packages/google/cloud/bigquery/job.py in _begin(self, client, retry, timeout)
    643             method="POST",
    644             path=path,
--> 645             data=self.to_api_repr(),
    646             timeout=timeout,
    647         )

~/python/lib/python3.7/site-packages/google/cloud/bigquery/job.py in to_api_repr(self)
   1694             )
   1695         _helpers._set_sub_prop(
-> 1696             configuration, ["load", "destinationTable"], self.destination.to_api_repr()
   1697         )
   1698 

AttributeError: 'dict' object has no attribute 'to_api_repr'
```

As you can see [here](https://github.com/googleapis/python-bigquery/blob/v2.0.0/tests/unit/test_client.py#L3609), the mocked client method `load_table_from_uri` is where the traceback begins.

The destination must be coerced to a `TableRef` before calling this method.

HemangChothani:
Fixes #297


google/cloud/bigquery/client.py
```diff
@@ -1625,11 +1626,9 @@ def create_job(self, job_config, retry=DEFAULT_RETRY):
                 job_config
             )
             destination = _get_sub_prop(job_config, ["copy", "destinationTable"])
+            destination = TableReference.from_api_repr(destination)
```
bhtucker, line 1629: Would you like to guard this check with `isinstance(TableReference, destination)`?

bhtucker (commented):
Since the root cause of this being broken for a while has been the test starting getting to autospec mocks very quickly, I wonder if it's possible to mock something a bit further along that still exercises the request marshalling. Seems like we should be able to at least have coverage through `to_api_repr` of the `Job` created by this method.

google/cloud/bigquery/client.py
HemangChothani: I mock further in test so now it cover `to_api_repr`.  I am not sure about the guard to check instance because `job_config` is dict and as per the example it has values in the string format.

google/cloud/bigquery/client.py
```diff
@@ -1625,11 +1626,9 @@ def create_job(self, job_config, retry=DEFAULT_RETRY):
...
             sources = []
             source_configs = _get_sub_prop(job_config, ["copy", "sourceTables"])
-
-            if source_configs is None:
-                source_configs = [_get_sub_prop(job_config, ["copy", "sourceTable"])]
```
tswast, line 1632: What's the reason for removing this code?

There is both a sourceTables and sourceTable property in the API representation. https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationTableCopy.FIELDS.source_table

tests/unit/test_client.py
```diff
@@ -3627,29 +3633,7 @@ def test_create_job_copy_config(self):
             }
         }
 
-        self._create_job_helper(
-            configuration, "google.cloud.bigquery.client.Client.copy_table",
-        )
-
-    def test_create_job_copy_config_w_single_source(self):
```
tswast, line 3634: Ditto on this. I see no reason to remove this test, as it's still documented as a possibility in the API.

tswast (changes requested):

google/cloud/bigquery/job.py
```diff
@@ -1871,7 +1870,15 @@ def to_api_repr(self):
         ]
 
         configuration = self._configuration.to_api_repr()
-        _helpers._set_sub_prop(configuration, ["copy", "sourceTables"], source_refs)
+        source_table = _helpers._get_sub_prop(
+            self._configuration._properties, ["copy", "sourceTable"]
+        )
+        if source_table:
+            _helpers._set_sub_prop(
```
tswast, line 1877: No need for this logic. We can always set the `sourceTables` with only a single item, but if given a job resource that has only `sourceTable` we need to be able to handle that.

tswast (changes requested):

google/cloud/bigquery/job.py
```diff
@@ -1871,6 +1870,12 @@ def to_api_repr(self):
         ]
 
         configuration = self._configuration.to_api_repr()
+        source_table = _helpers._get_sub_prop(
```
tswast, line 1873: I'm still very confused by this logic. I think it'll be unnecessary once https://github.com/googleapis/python-bigquery/pull/323 is merged.

tswast (changes requested):

tests/unit/test_client.py
```diff
@@ -3573,21 +3573,34 @@ def test_delete_table_w_not_found_ok_true(self):
...
+            "jobReference": {"projectId": self.PROJECT, "jobId": mock.ANY},
+            "configuration": job_config,
+        }
+        conn = client._connection = make_connection(RESOURCE)
+        client.create_job(job_config=job_config)
 
-        with load_patch as client_method, get_config_patch:
-            client.create_job(job_config=job_config)
-        client_method.assert_called_once()
+        if "copy" in job_config:
```
tswast, line 3590: Ditto. I'm sure confused by this logic, but I think whatever it was meant to do is unnecessary after https://github.com/googleapis/python-bigquery/pull/323

google/cloud/bigquery/job.py
HemangChothani: In the current situation i have to delete `sourceTable` from the already passed resource, else it throws an error like `can't have both sourceTable and sourceTables in config`. but it think after that PR merged don't need this type of check so i will wait for that PR to be merged and then update this PR accordingly.

tests/unit/test_client.py
```diff
@@ -3573,21 +3573,34 @@ def test_delete_table_w_not_found_ok_true(self):
...
+            "jobReference": {"projectId": self.PROJECT, "jobId": mock.ANY},
+            "configuration": job_config,
+        }
+        conn = client._connection = make_connection(RESOURCE)
+        client.create_job(job_config=job_config)
 
-        with load_patch as client_method, get_config_patch:
-            client.create_job(job_config=job_config)
-        client_method.assert_called_once()
+        # if "copy" in job_config:
```
tswast, line 3590: Is this commented out code needed?

tests/unit/test_client.py
```diff
@@ -3573,21 +3573,34 @@ def test_delete_table_w_not_found_ok_true(self):
...
+        #     if "sourceTable" in job_config["copy"]:
+        #         source = _helpers._get_sub_prop(job_config, ["copy", "sourceTable"])
+        #         _helpers._del_sub_prop(job_config, ["copy", "sourceTable"])
+        #         _helpers._set_sub_prop(job_config, ["copy", "sourceTables"], [source])
+        if "query" in job_config:
+            _helpers._del_sub_prop(job_config, ["query", "destinationTable"])
```
tswast, line 3596: Do we still need to remove the destination table property from the config after my refactoring?

I thought we allow None to be passed to the job constructors for most properties now.

tswast: I forget if we automatically remove the destination table to support query re-runs, but that seems like something we shouldn't remove by default, as explicit destination tables are also supported by the API.

tests/unit/test_client.py
HemangChothani: Sorry, my bad.

google/cloud/bigquery/client.py
```diff
@@ -1654,7 +1657,6 @@ def create_job(self, job_config, retry=DEFAULT_RETRY):
             )
         elif "query" in job_config:
             copy_config = copy.deepcopy(job_config)
-            _del_sub_prop(copy_config, ["query", "destinationTable"])
```
tswast, line 1657: I didn't realize we were already removing this property in past releases. We should restore this, since I don't want to break existing clients.

google/cloud/bigquery/client.py
```diff
@@ -48,11 +48,11 @@
 from google.cloud import exceptions
 from google.cloud.client import ClientWithProject
 
+from google.cloud.bigquery._helpers import _del_sub_prop
 from google.cloud.bigquery._helpers import _get_sub_prop
 from google.cloud.bigquery._helpers import _record_field_to_json
 from google.cloud.bigquery._helpers import _str_or_none
 from google.cloud.bigquery._helpers import _verify_job_config_type
-from google.cloud.bigquery._helpers import _del_sub_prop
```
tswast, line 55: I think this change needs to be reverted too, since `_del_sub_prop` is still used.

HemangChothani: I have added `_del_sub_prop` before `_get_sub_prop` import to maintain.

tswast (approved):

PR #314, opened 2020-10-09
deps: require pyarrow for pandas support

### LINKED ISSUE #265: drop fastparquet, add pyarrow to pandas extras deps
We've got a few code paths and tests that use fastparquet instead of pyarrow for serialization. This is a quite a bit of overhead and we end up directing people to pyarrow anyway because it's faster to download dataframes that way (even without the BQ Storage API, since pyarrow is faster at concatenating record buffers than pandas is at concatenating dataframes).

When we go to 2.0, I think that's probably a good time to drop fastparquet.

cguardia:

Fixes #265 🦕


tests/unit/test_client.py
```diff
@@ -7839,7 +7835,6 @@ def test_load_table_from_dataframe_unknown_table(self):
         )
 
     @unittest.skipIf(pandas is None, "Requires `pandas`")
-    @unittest.skipIf(fastparquet is None, "Requires `fastparquet`")
     def test_load_table_from_dataframe_no_pyarrow_warning(self):
```
tswast, line 7838: I'm a bit surprised to see this test passing. I guess we still have some code that falls back to the default pandas parquet rendering?

Can you look into if we can remove that code path?

Related: We should be able to simplify this docstring now: https://github.com/googleapis/python-bigquery/blob/cbcb4b86d2d01fb983c20271958b11665e8f7d03/google/cloud/bigquery/client.py#L2134-L2147

setup.py
```diff
@@ -53,7 +53,6 @@
         "pyarrow >= 1.0.0, < 2.0dev",
     ],
     "tqdm": ["tqdm >= 4.7.4, <5.0.0dev"],
-    "fastparquet": ["fastparquet", "python-snappy", "llvmlite>=0.34.0"],
```
tswast, line 56: I'd like to see us add "pyarrow" to the "pandas" extras now, since it's needed for both uploads and downloads to dataframe.

We can maybe refactor the `pyarrow >=1.0.0,<2.0dev` string into a variable since it's going to appear 3 times in setup.py now too

tswast (changes requested):

cguardia:
@tswast OK, this was a bit more involved than I expected at the beginning. Here goes my second attempt.

tswast (approved):
Thanks!

PR #332, opened 2020-10-19
WIP: use proto-plus instead of _pb

### LINKED ISSUE #319: Breaking Change: update model.py classes to use proto-plus instead of _pb
Currently the classes in `model.py` use the underlying protobuf `_pb` object for read-only properties instead of the proto-plus object.

https://github.com/googleapis/python-bigquery/blob/b8f502b14f21d1815697e4d57cf1225dfb4a7c5e/google/cloud/bigquery/model.py#L58

This avoided some possible breaking changes, but was mostly done to expedite the transition to the microgenerator.

There are some benefits to using proto-plus that would be worth taking advantage of. Most importantly, timestamp well-known types are properly translated into a datetime-like object with proto-plus.

Why now?

There are some breaking changes in the latest generated clients due to https://github.com/googleapis/gapic-generator-python/pull/595

* https://github.com/googleapis/python-bigquery/pull/316/files#diff-daa390c4795c10c42d5e8cec48f5339c

If we need to make a breaking change for these changes, we might as well update the classes to support the latest versions.

cguardia:
This is a proof of concept PR to fix #319. This almost works, but there are four tests failing. The failures are of two types:

1. The `message.from_json` method, which is the likeliest successor to the `json_format.ParseDict` call used with the `_pb` object, does not allow us to pass in the "ignore_unknown_fields" flag to the parser, thus causing errors in tests that try to ignore fields.

2. The json format created by the regular json library has some slight differences with the format generated by `json_format`, which causes tests trying to compare `repr` output to fail.

I don't like very much the use of the json library in this PR, and would prefer to avoid it if possible. I think perhaps making a PR to proto-plus to make the `to_json` and `from_json` methods more flexible would be a good idea, along with possibly adding `to_dict` and `from_dict` methods there, which seems to me could be generally helpful.

### DIFF
--- a/google/cloud/bigquery/model.py
+++ b/google/cloud/bigquery/model.py
@@ -18,7 +18,7 @@
 
 import copy
 
-from google.protobuf import json_format
+import json
 import six
 
 import google.cloud._helpers
@@ -55,7 +55,7 @@ class Model(object):
     def __init__(self, model_ref):
         # Use _proto on read-only properties to use it's built-in type
         # conversion.
-        self._proto = types.Model()._pb
+        self._proto = types.Model()
 
         # Use _properties on read-write properties to match the REST API
         # semantics. The BigQuery API makes a distinction between an unset
@@ -67,7 +67,7 @@ class Model(object):
             model_ref = ModelReference.from_string(model_ref)
 
         if model_ref:
-            self._proto.model_reference.CopyFrom(model_ref._proto)
+            self._proto.model_reference = model_ref._proto
 
     @property
     def reference(self):
@@ -305,9 +305,7 @@ class Model(object):
             start_time = datetime_helpers.from_microseconds(1e3 * float(start_time))
             training_run["startTime"] = datetime_helpers.to_rfc3339(start_time)
 
-        this._proto = json_format.ParseDict(
-            resource, types.Model()._pb, ignore_unknown_fields=True
-        )
+        this._proto = types.Model.from_json(json.dumps(resource))
         return this
 
     def _build_resource(self, filter_fields):
@@ -323,7 +321,7 @@ class Model(object):
         Returns:
             Dict[str, object]: Model reference represented as an API resource
         """
-        return json_format.MessageToDict(self._proto)
+        return json.loads(types.Model.to_json(self._proto))
 
 
 class ModelReference(object):
@@ -334,7 +332,7 @@ class ModelReference(object):
     """
 
     def __init__(self):
-        self._proto = types.ModelReference()._pb
+        self._proto = types.ModelReference()
         self._properties = {}
 
     @property
@@ -377,9 +375,7 @@ class ModelReference(object):
         # Keep a reference to the resource as a workaround to find unknown
         # field values.
         ref._properties = resource
-        ref._proto = json_format.ParseDict(
-            resource, types.ModelReference()._pb, ignore_unknown_fields=True
-        )
+        ref._proto = types.ModelReference.from_json(json.dumps(resource))
 
         return ref
 
@@ -418,7 +414,7 @@ class ModelReference(object):
         Returns:
             Dict[str, object]: Model reference represented as an API resource
         """
-        return json_format.MessageToDict(self._proto)
+        return json.loads(types.ModelReference.to_json(self._proto))
 
     def _key(self):
         """Unique key for this model.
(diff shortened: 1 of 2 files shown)


tswast:
I've filed https://github.com/googleapis/proto-plus-python/issues/153 and https://github.com/googleapis/proto-plus-python/issues/152 which will allow us to ignore unknown fields and avoid the use of the `json` module.

google/cloud/bigquery/model.py
```diff
@@ -305,9 +305,7 @@ def from_api_repr(cls, resource):
             start_time = datetime_helpers.from_microseconds(1e3 * float(start_time))
             training_run["startTime"] = datetime_helpers.to_rfc3339(start_time)
 
-        this._proto = json_format.ParseDict(
-            resource, types.Model()._pb, ignore_unknown_fields=True
-        )
+        this._proto = types.Model.from_json(json.dumps(resource))
```
tswast, line 308: Can we pass `resource` to the `Model` constructor here?

https://github.com/googleapis/proto-plus-python/issues/152#issuecomment-712290339

tswast: We'll still want https://github.com/googleapis/proto-plus-python/issues/153 to support `ignore_unknown_fields`, though.

cguardia: I already had tried that during my testing. It gives a KeyError on the metaclass during `__init__`, right here:

https://github.com/googleapis/proto-plus-python/blob/master/proto/message.py#L425

I was using this code:

    this._proto = types.Model(resource)

tswast:
https://github.com/googleapis/proto-plus-python/pull/154 adds `ignore_unknown_fields` to both the `Model` constructor and a `to_dict` method. Hopefully that means this PR won't be blocked by proto-plus once they make a release.

cguardia:
They already released. Still getting some test errors here, but I think they are related to assumptions about how the pb json_format conversion worked. Hopefully will have a working PR at some point tonight.

cguardia:
@tswast Speaking about hidden json_format pb conversion behavior: somewhere along the line it converted key names to camelCase automatically, so many checks we have on resources fail due to the ids now using snake_case and the manually constructed test resources using camelCase. In some cases, the camelCase names are part of the main code, like in job and client modules. Dov said this has been a problem elsewhere. I'm still not familiar enough with the bigquery codebase to be completely sure which way to go here: my instinct is to convert everything to snake_case, what do you think?

google/cloud/bigquery/model.py
tswast: We should be able to update this to `types.Model(resource, ignore_unknown_fields=True)` now.

tswast:
>  json_format pb conversion behavior: somewhere along the line it converted key names to camelCase automatically, so many checks we have on resources fail due to the ids now using snake_case and the manually constructed test resources using camelCase. 

@cguardia  You mean it's taking the `camelCase` keys we set in tests and converting them to `snake_case`? I suspect the API will accept either in most cases, since the language guide says

> Parsers accept both the lowerCamelCase name (or the one specified by the `json_name` option) and the original proto field name.

https://developers.google.com/protocol-buffers/docs/proto3#json

Does the same behavior happen with `to_dict`?

cguardia:
Not using JSON anymore. The tests are expecting CamelCase, but the calls to the constructor and to_dict are returning snake_case. I was thinking of getting rid of camelCase everywhere, but unsure if that would affect something else.

tswast:
> calls to the constructor and to_dict are returning snake_case. I was thinking of getting rid of camelCase everywhere, but unsure if that would affect something else.

This seems like a safe operation to me. We aren't ever sending that JSON representation to the API, so it shouldn't even affect that.

cguardia:
Well, this has been quite the rabbit hole. Turns out not only the tests, but some of the code as well, seem to be counting on the automatic case conversion, so that a lot of code would need to be changed to get rid of that. The `json_format` call also performs some data conversion, which again is expected to happen in some of the code. I believe we have three options:

1. Bite the bullet and go through all the code and tests until this is right. This is complicated by some API calls that want camelCase arguments, so each change has to be carefully done to avoid affecting those.
2. Write some helper code to do the case conversion on our side and handle this as an API requirement.
3. Use `json.loads` and `json.dumps` in combination with the `Model.to_json` code to get the conversion done like it was done before (mostly the current code with a few additions).

@tswast what do you think?

tswast:
2. and 3. sound like asking for trouble performance-wise. That several passes over the data in Python code to get things in the structure we want.

> Bite the bullet and go through all the code and tests until this is right.

I don't recall what code relies on this. Could you send a pointer? Are there really all that many tests for this? I recall that all of the properties we use `_proto` for are read-only.

cguardia:
I'm going to close this one, and open another with the current approach.

PR #41, opened 2020-02-21
feat: pass retry from Job.result() to Job.done()

IlyaFaer:
Closes https://github.com/googleapis/python-bigquery/issues/24

PR #340, opened 2020-10-23
test: make `_AsyncJob` tests mock at a lower layer

### LINKED ISSUE #339: _AsyncJob test failure
https://source.cloud.google.com/results/invocations/13cec9de-6f0f-4392-a32b-075930bd6c99/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fpresubmit/log

```
___________________ Test_AsyncJob.test_result_retry_to_done ____________________

self = <tests.unit.test_job.Test_AsyncJob testMethod=test_result_retry_to_done>

    def test_result_retry_to_done(self):
        client = _make_client(project=self.PROJECT)
        job = self._make_one(self.JOB_ID, client)
        done = job.done = mock.Mock()
        retry = mock.Mock()

>       job.result(retry=retry)

tests/unit/test_job.py:999:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
google/cloud/bigquery/job.py:819: in result
    self._begin(retry=retry, timeout=timeout)
google/cloud/bigquery/job.py:631: in _begin
    data=self.to_api_repr(),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <google.cloud.bigquery.job._AsyncJob object at 0x7fc21b43bcd0>

    def to_api_repr(self):
        """Generate a resource for the job."""
>       raise NotImplementedError("Abstract")
E       NotImplementedError: Abstract

google/cloud/bigquery/job.py:591: NotImplementedError
```

tswast:
This is intented to make the `_AsyncJob` tests more robust
to changes in retry behavior. It also more explicitly tests
the retry behavior by observing API calls rather than calls
to certain methods.

Fixes #339 🦕

### DIFF
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -529,9 +529,8 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
             Optional[str]:
                 the state (None until set from the server).
         """
-        status = self._properties.get("status")
-        if status is not None:
-            return status.get("state")
+        status = self._properties.get("status", {})
+        return status.get("state")
 
     def _set_properties(self, api_response):
         """Update properties from resource in body of ``api_response``
@@ -588,7 +587,7 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
 
     def to_api_repr(self):
         """Generate a resource for the job."""
-        raise NotImplementedError("Abstract")
+        return copy.deepcopy(self._properties)
 
     _build_resource = to_api_repr  # backward-compatibility alias
 
--- a/tests/unit/test_job.py
+++ b/tests/unit/test_job.py
@@ -20,6 +20,8 @@ import unittest
 import warnings
 
 import freezegun
+from google.api_core import exceptions
+import google.api_core.retry
 import mock
 import pytest
 import requests
@@ -70,6 +72,12 @@ def _make_connection(*responses):
     return mock_conn
 
 
+def _make_retriable_exception():
+    return exceptions.TooManyRequests(
+        "retriable exception", errors=[{"reason": "rateLimitExceeded"}]
+    )
+
+
 def _make_job_resource(
     creation_time_ms=1437767599006,
     started_time_ms=1437767600007,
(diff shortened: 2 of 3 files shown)


tests/unit/test_job.py
```diff
@@ -965,43 +976,95 @@ def test_done_already(self):
...
+            job_id=self.JOB_ID, project_id=self.PROJECT, started=True, ended=True
+        )
+        conn = _make_connection(
+            exceptions.NotFound("not normally retriable"),
+            begun_job_resource,
+            # The call to done() / reload() does not get the custom retry
+            # policy passed to it, so we don't throw a non-retriable
+            # exception here. See:
+            # https://github.com/googleapis/python-bigquery/issues/24
+            _make_retriable_exception(),
```
tswast, line 1027: @IlyaFaer You should be able to verify your changes in https://github.com/googleapis/python-bigquery/pull/41 by changing this from `_make_retriable_exception()` to `NotFound` as is done a few lines above.

google/cloud/bigquery/job.py
```diff
@@ -588,7 +587,7 @@ def _check_resource_config(cls, resource):
 
     def to_api_repr(self):
         """Generate a resource for the job."""
-        raise NotImplementedError("Abstract")
+        return copy.deepcopy(self._properties)
 
     _build_resource = to_api_repr  # backward-compatibility alias
```
steffnay, line 592: This all LGTM, but I'm curious what this backwards compatibility is for here. It's not clear to me why we were specifically testing to make sure this would raise NotImplementedError before.

steffnay (approved):

google/cloud/bigquery/job.py
tswast: > It's not clear to me why we were specifically testing to make sure this would raise NotImplementedError before.

I think that was solely to make sure the coverage tests pass. We're still doing a bit of funny business in the `*Job` classes, in that most (all?) subclasses omit the `statistics` property from their `to_api_repr`.

> I'm curious what this backwards compatibility is for here.

I don't remember, actually. I see a few hits for `_build_resource` in search. https://github.com/googleapis/python-bigquery/search?p=1&q=_build_resource I think `_build_resource` is usually the name we use for "update" methods that need to take a field mask and only populate certain fields. Jobs don't support update, so not sure why we ever had one.

Issue #18, opened 2019-07-21
BigQuery: make jobs awaitable


dkapitan:
I know BigQuery jobs are asynchronous by default. However, I am struggling to make my datapipeline async end-to-end.

Looking at [this JS example](https://medium.com/skyshidigital/what-bigquery-is-and-isnt-d092fa173d53), I thought it would be the most Pythonic to make a BigQuery job awaitable. However, I can't get that to work in Python i.e. errors when `await client.query(query)`. Looking at the source code, I don't see which method [returns an awaitable object](https://www.python.org/dev/peps/pep-0492/).

 I have little experience in writing async Python code and found [this example that wraps jobs in a `async def coroutine`](https://stackoverflow.com/questions/53065340/python-how-to-big-query-asynchronous-tasks).
```
class BQApi(object):                                                                                                 
    def __init__(self):                                                                                              
        self.api = bigquery.Client.from_service_account_json(BQ_CONFIG["credentials"])                               

    async def exec_query(self, query, **kwargs) -> bigquery.table.RowIterator:                                       
        job = self.api.query(query, **kwargs)                                                                        
        task = asyncio.create_task(self.coroutine_job(job))                                                          
        return await task                                                                                            

    @staticmethod                                                                                                    
    async def coroutine_job(job):                                                                                    
        return job.result()   
```

The [`google.api_core.operation.Operation`](https://googleapis.github.io/google-cloud-python/latest/core/operation.html) shows how to use `add_done_callback` to asynchronously wait for long-running operations. I have tried that, but the following yields `AttributeError: 'QueryJob' object has no attribute '_condition'` :

```
from concurrent.futures import ThreadPoolExecutor, as_completed
query1 = 'SELECT 1'
query2 = 'SELECT 2'

def my_callback(future):
    result = future.result()

operations = [bq.query(query1), bq.query(query2)]
[operation.add_done_callback(my_callback) for operation in operations]
results2 = []
for future in as_completed(operations):
  results2.append(list(future.result()))
```

Given that jobs are already asynchronous, would it make sense to add a method that returns an awaitable?

Or am I missing something and is there an Pythonic way to use the BigQuery client with the async/await pattern?

jsigee87:
Wanted to bump this thread, I am currently trying to write async BigQuery jobs myself and it would be great if the job were `awaitable`.

vgrabovets:
Any news here? You've already dropped support for Python 2.

yanhong-zhao-ef:
[...]

PR #341, opened 2020-10-23
feat: add `reload` argument to `*Job.done()` functions

tswast:
This enables checking the job status without making an API call.

It also fixes an inconsistency in `QueryJob`, where a job can be
reported as "done" without having the results of a `getQueryResults` API
call.

Follow-up to https://github.com/googleapis/python-bigquery/pull/340

### DIFF
--- a/google/cloud/bigquery/job.py
+++ b/google/cloud/bigquery/job.py
@@ -768,7 +768,7 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
             # set, do not call set_result/set_exception again.
             # Note: self._result_set is set to True in set_result and
             # set_exception, in case those methods are invoked directly.
-            if self.state != _DONE_STATE or self._result_set:
+            if not self.done(reload=False) or self._result_set:
                 return
 
             if self.error_result is not None:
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/job.py
```diff
@@ -3082,10 +3085,25 @@ def done(self, retry=DEFAULT_RETRY, timeout=None):
...
+                If ``True``, make an API call to refresh the job state of
+                unfinished jobs before checking. Default ``True``.
 
         Returns:
             bool: True if the job is complete, False otherwise.
         """
+        is_done = (
+            # Only consider a QueryJob complete when we know we have the final
+            # query results available.
+            self._query_results is not None
```
shollyman, line 3098: checking: What about DDL statements, where we have no results?  Or would this already retain the empty row iterator?

shollyman (approved):
Do the changes here indicate there's further changes for registering the done callback?  My mental model has them independent, but I wanted to make sure.

google/cloud/bigquery/job.py
tswast: DDL statements don't 404 for `jobs.getQueryResults`, do they? This is the full response object of `jobs.getQueryResults`, not just the rows.

tswast:
> Do the changes here indicate there's further changes for registering the done callback?

No callbacks here. Maybe when we implement the async client https://github.com/googleapis/python-bigquery/issues/18

`done` does get called by `result` via [PollingFuture._blocking_poll](https://github.com/googleapis/python-api-core/blob/85e14e3f3a6b3e3c3ae9626daf60183f4c05fb6a/google/api_core/future/polling.py#L107), which is why I can now remove the extra call to `_get_query_results` from `QueryJob.result()`.

tswast:
> Do the changes here indicate there's further changes for registering the done callback?

Maybe you meant -- How do we get the needed arguments to `get_query_results` from `result`? I'm planning to do something a little bit hacky in https://github.com/googleapis/python-bigquery/pull/347/files -- which maybe is a bad idea, as it means you can't call `result` from two threads simultaneously. Hmmm...

PR #347, opened 2020-10-27
perf: use `jobs.getQueryResults` to download result sets

tswast:
Since `getQueryResults` was already used to wait for the job to finish,
this avoids an additional call to `tabledata.list`. The first page of
results are cached in-memory.

Additional changes will come in the future to avoid calling the BQ
Storage API when the cached results contain the full result set.

Towards https://github.com/googleapis/python-bigquery/issues/362

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -80,12 +80,12 @@ _DEFAULT_CHUNKSIZE = 1048576  # 1024 * 1024 B = 1 MB
 _MAX_MULTIPART_SIZE = 5 * 1024 * 1024
 _DEFAULT_NUM_RETRIES = 6
 _BASE_UPLOAD_TEMPLATE = (
-    u"https://bigquery.googleapis.com/upload/bigquery/v2/projects/"
-    u"{project}/jobs?uploadType="
+    "https://bigquery.googleapis.com/upload/bigquery/v2/projects/"
+    "{project}/jobs?uploadType="
 )
-_MULTIPART_URL_TEMPLATE = _BASE_UPLOAD_TEMPLATE + u"multipart"
-_RESUMABLE_URL_TEMPLATE = _BASE_UPLOAD_TEMPLATE + u"resumable"
-_GENERIC_CONTENT_TYPE = u"*/*"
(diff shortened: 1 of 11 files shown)


tswast:
Based on https://github.com/googleapis/python-bigquery/pull/341

google/cloud/bigquery/job.py
```diff
@@ -2646,6 +2649,7 @@ def __init__(self, job_id, query, client, job_config=None):
             )
 
         self._query_results = None
+        self._get_query_results_kwargs = {}
```
tswast, line 2652: Does this need to be a thread-local variable?

tswast: Actually, the cached query results might need to be thread-local too. Imagine if two threads called `result` with different starting indexes and/or max results.

tswast: We'll also need some logic like

https://github.com/googleapis/google-cloud-go/blob/925033712191bce44fa99eb117d6531106042272/bigquery/iterator.go#L314

to see if we can use the cached page if `result` is called more than once

tswast: Done in latest commit.

google/cloud/bigquery/client.py
```diff
@@ -3172,6 +3191,64 @@ def list_rows(
...
+                before using ``retry``.
+                If multiple requests are made under the hood, ``timeout``
+                applies to each individual request.
+
+        Returns:
+            google.cloud.bigquery.table.RowIterator:
+                Iterator of row data
+                :class:`~google.cloud.bigquery.table.Row`-s.
+        """
+        row_iterator = RowIterator(
```
tswast, line 3240: Be sure to populate extra args with the field projection. We only need rows and page token.

tswast:
Per our discussion, I'll be splitting this into 2 PRs:

1. Call `getQueryResults` (no cache) from `RowIterator` -- make sure to add a projection to exclude the schema and other irrelevant job stats.
2. Cache the first page of results.

I'll base them on the refactoring to split up the giant job module here: https://github.com/googleapis/python-bigquery/pull/361

Issue #296, opened 2020-10-05
Add max_results option to QueryJob.to_dataframe and QueryJob.to_arrow methods


tswast:
Currently, pandas-gbq calls `QueryJob.result()` and `Client.list_rows()` directly

https://github.com/pydata/pandas-gbq/blob/46c579ac21879b431c8568b49e68624f4a5ea25e/pandas_gbq/gbq.py#L561-L564

This is because the `max_results` parameter is needed, but not available in `to_dataframe`.

Currently, this is not a problem except for some duplicate code, but it may keep pandas-gbq from benefiting from the "fast query path" changes currently being designed.

plamut:
@tswast Isn't `max_results` already set when a query job is created? If one then says:
```py
rows_iterator = query_job.result()
rows_iterator.to_dataframe(max_results=42)
```
The `max_results` argument should override the same setting on the query job? And stop iterating through pages/rows once that many rows have been fetched and yielded to the user code?

Also, `max_results` is not compatible with BQ Storage client, thus if that is used, the code should just fall back to the REST API with a warning?

tswast:
> The max_results argument should override the same setting on the query job?

I don't see `max_results` here? https://github.com/googleapis/python-bigquery/blob/790d11bdbbf50d2f30c09ba4660d420ba2212866/google/cloud/bigquery/client.py#L3140-L3150

> Also, max_results is not compatible with BQ Storage client, thus if that is used, the code should just fall back to the REST API with a warning?

I'm pretty sure that's the current behavior.

> rows_iterator.to_dataframe(max_results=42)

Oh! No, I didn't mean here. I meant on the `QueryJob`. There's a `QueryJob.to_dataframe()` method and a `QueryJob.to_arrow()` method that call `result()`. I'd like them to be able to pass a `max_results` to that.

tswast:
(Looking at this, probably the solution in pandas-gbq is to call `.result()` with a `max_results` argument. Not sure why we weren't doing that.)

plamut:
Ah, sorry, I meant the row iterator, yes.

> (Looking at this, probably the solution in pandas-gbq is to call .result() with a max_results argument. Not sure why we weren't doing that.)

I thought there was a reason for that in Pandas GBQ and that you later wanted to pass `max_results` to the rows iterator that is returned by `query_job.result()` :laughing: 

[...]

PR #352, opened 2020-10-29
feat: add progress bar for to_arrow method

### LINKED ISSUE #343: feat: progress bar during query in `QueryJob.to_dataframe` and `QueryJob.to_arrow`

When running a query via the `%%bigquery` magics or waiting for it to finish via [QueryJob.to_dataframe](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJob.html#google.cloud.bigquery.job.QueryJob.to_dataframe) or [QueryJob.to_arrow](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJob.html#google.cloud.bigquery.job.QueryJob.to_arrow), an argument `progress_bar_type` is accepted.

Currently, this only shows the progress of the query results download. It would be great if it would also give an indicator while the query is executing.

When a value is passed to `progress_bar_type`, show some kind of progress bar. Ideally, it would work similarly to the UI. For example,

* Show the job state in the progress bar description. For example, is it currently "pending" (queued, waiting for resources) or "running" (actually executing).
* Show how many "stages" there are via length of the [query_plan](https://github.com/googleapis/python-bigquery/blob/d093cd4ee255bbcbe5a6a198e819b876b4aa51f9/google/cloud/bigquery/job.py#L2841).
* Find the latest incomplete stage.
* Use [parallel_inputs](https://github.com/googleapis/python-bigquery/blob/d093cd4ee255bbcbe5a6a198e819b876b4aa51f9/google/cloud/bigquery/job.py#L3514) as the total amount of work (per stage)
* Use [completed_parallel_inputs](https://github.com/googleapis/python-bigquery/blob/d093cd4ee255bbcbe5a6a198e819b876b4aa51f9/google/cloud/bigquery/job.py#L3521) as the amount of work completed so far.

To populate this, instead of calling `result()` once:

* Call `result(timeout=[a few seconds])` every few seconds.
* Call `job.reload()` to fetch the latest job statistics.
* Update the progress bar.
* Repeat.

* Some kind of spinner that only shows that some time has elapsed. (This is actually harder than it sounds because `tqdm` doesn't actually support spinners, and if it ever does, it sounds like it'll be via a different API than the current progress bars. tqdm/tqdm#427 tqdm/tqdm#925

 **Additional context**

* pandas-gbq feature request: https://github.com/pydata/pandas-gbq/issues/327
* Example of BigQuery UI progress bar:

  bq-progress-small

HemangChothani:
Fixes #343 

- PR open for the feedback and suggestions.

- Currently implemented for 'QueryJob.to_arrow' method 

- When the `cacheHit` for result is true at time `query_plan` is blank so can't implement progress bar.


google/cloud/bigquery/job.py
```diff
@@ -20,6 +20,9 @@
 import copy
 import re
 import threading
+import tqdm
```
tswast, line 23: This needs to be an optional import. See:

https://github.com/googleapis/python-bigquery/blob/20f473bfff5ae98377f5d9cdf18bfe5554d86ff4/google/cloud/bigquery/table.py#L39-L42

google/cloud/bigquery/job.py
```diff
@@ -3275,6 +3283,27 @@ def result(
         rows._preserve_order = _contains_order_by(self.query)
         return rows
 
+    def _get_progress_bar(self, progress_bar_type, description, total, unit):
```
tswast, line 3286: Seems like this is very close to the logic in `table.py`

https://github.com/googleapis/python-bigquery/blob/20f473bfff5ae98377f5d9cdf18bfe5554d86ff4/google/cloud/bigquery/table.py#L1374

I'd suggest creating a `_tqdm_helpers.py` module similar to our pandas helpers to hold this logic for both Table and Job logic.

google/cloud/bigquery/job.py
```diff
@@ -3337,7 +3339,49 @@ def to_arrow(
 
         ..versionadded:: 1.17.0
         """
-        return self.result().to_arrow(
+        if self.query_plan and progress_bar_type:
```
tswast, line 3342: Is this logic necessary if we are calling `to_arrow`? Can't we rely on `to_arrow`'s progress bar support?

google/cloud/bigquery/job.py
```diff
@@ -3406,7 +3450,46 @@ def to_dataframe(
         Raises:
             ValueError: If the `pandas` library cannot be imported.
         """
-        return self.result().to_dataframe(
+        query_plan = self.query_plan
+        if query_plan and progress_bar_type:
```
tswast, line 3454: `query_plan` can get updated as the job progresses. I'd prefer if this always created a progress bar, but had a generic message when the job is still pending and there isn't a query plan.

tswast (changes requested):

google/cloud/bigquery/job.py
HemangChothani: Sorry, didn't get this point, are you talking about table.to_arrow() progress bar support?

tswast: Oh, I was getting this confused with `RowIterator`, where `to_dataframe` relies on `to_arrow`'s progress bar support. I do wonder if some of this "waiting for the query to finish" logic could be refactored into a `_tqdm_helpers.py` function, since it should be identical between `to_dataframe` and `to_arrow`

google/cloud/bigquery/job.py
```diff
@@ -3404,7 +3452,51 @@ def to_dataframe(
...
+                    progress_bar.set_description(
+                        "Query executing stage {} and status {} : {:0.2f}s".format(
+                            current_stage.name,
+                            current_stage.status,
+                            time.time() - start_time,
+                        )
+                    )
+
+                    try:
+                        query_result = self.result(timeout=0.5)
```
tswast, line 3476: We should make constants for this (`PROGRESS_BAR_UPDATE_INTERVAL`, for example)

google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -0,0 +1,106 @@
...
+            return tqdm.tqdm_gui(desc=description, total=total, unit=unit)
+    except (KeyError, TypeError):
+        # Protect ourselves from any tqdm errors. In case of
+        # unexpected tqdm behavior, just fall back to showing
+        # no progress bar.
+        warnings.warn(_NO_TQDM_ERROR, UserWarning, stacklevel=3)
+    return None
+
+
+def _query_job_result_helper(query_job, progress_bar_type=None):
```
tswast, line 56: Let's use active verbs in this helper name.

```suggestion
def wait_for_query(query_job, progress_bar_type=None):
```

google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -0,0 +1,106 @@
...
+    """Return query result and display a progress bar while the query running, if tqdm is installed."""
+    if progress_bar_type:
```
tswast, line 58: We can reduce the level of indentation if we exit early. Also, I think we'll want to pass through some keyword arguments to `result()`, correct?

```suggestion
    if progress_bar_type is None:
        return query_job.result()
```

Aside (not relevant for this PR): we'll eventually want to pass additional arguments to `result()` whenever we implement https://github.com/googleapis/python-bigquery/issues/296

google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -0,0 +1,106 @@
...
+
+_NO_TQDM_ERROR = (
+    "A progress bar was requested, but there was an error loading the tqdm "
+    "library. Please install tqdm to use the progress bar functionality."
+)
+
+_PROGRESS_BAR_UPDATE_INTERVAL = 0.5
+
+
+def _get_progress_bar(progress_bar_type, description, total, unit):
```
tswast, line 34: We want to use these helpers from other modules, let's remove the (redundant because of private module) leading `_`.

```suggestion
def get_progress_bar(progress_bar_type, description, total, unit):
```

tswast (commented):
Looking good. A few nits regarding naming and indentation.

google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _query_job_result_helper(query_job, progress_bar_type=None):
+    """Return query result and display a progress bar while the query running, if tqdm is installed."""
+    if progress_bar_type:
+        start_time = time.time()
+        progress_bar = _get_progress_bar(
+            progress_bar_type, "Query is running", 1, "query"
+        )
+        if query_job.query_plan:
```
tswast, line 63: This should be one level deeper. I'd like to see the `while True` loop, even when `query_plan` is not initially populated.

google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -0,0 +1,105 @@
...
+                    )
+                    break
+                except concurrent.futures.TimeoutError:
+                    if current_stage.status == "COMPLETE":
+                        if i < total - 1:
+                            progress_bar.update(i + 1)
+                            i += 1
+                    continue
+            else:
+                query_result = query_job.result()
```
tswast, line 97: We need a timeout here. It's not clear to me why the above `try` block is even in the `if` statement.

google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -0,0 +1,105 @@
...
+    else:
+        start_time = time.time()
+        progress_bar = get_progress_bar(
+            progress_bar_type, "Query is running", 1, "query"
+        )
+        i = 0
+        while True:
+            if query_job.query_plan:
+                total = len(query_job.query_plan)
+                query_job.reload()  # Refreshes the state via a GET request.
```
tswast, line 69: Wouldn't we only want to call `reload` after `result` times out?

Also, why would we only call `reload` inside this `if` statement? The job might not have a `query_plan` until **after** `reload` is called in many cases.

tswast (changes requested):
I'd expect the `while` loop to contain the following steps:

1. Call `result()` with a short-ish timeout.
2. Call `reload()`
3. Update the progress bar (different cases for `query_plan` present or not)

google/cloud/bigquery/_tqdm_helpers.py
tswast: Looks like the only difference is the presence of `total`, which could be set to a default value in the case where `query_plan` is not present.

HemangChothani:
system test failed not related to changes.

tswast (approved):
Love it!

P.S. I have opened https://github.com/googleapis/python-bigquery/pull/388 to fix those failing tests.

Issue #362, opened 2020-11-03
Query performance optimizations


tswast:
This issue tracks the "fast query path" changes for the Python client(s):

- [x] https://github.com/googleapis/python-bigquery/pull/363 -- Update `QueryJob` to use `getQueryResults` in `RowIterator`. Project down to avoid fetching schema and other unnecessary job stats in `RowIterator`.
- [ ] https://github.com/googleapis/python-bigquery/pull/374 -- Update `QueryJob` and `RowIterator` to cache the first page of results, which we fetch as part of the logic to wait for the job to finish. Discard the cache if `maxResults` or `startIndex` are set.
- [x] https://github.com/googleapis/python-bigquery/pull/375 -- Update DB-API to avoid direct call to list_rows()
- [x] https://github.com/googleapis/python-bigquery/pull/384 -- Update `to_dataframe` and related methods in RowIterator to not call BQ Storage API if cached results are the only page.
- [ ] Update DB-API to not call BQ Storage API if cached results are the only page.
- [ ] Update Client.query to call jobs.query backend API method for acceptable `job_config`s.
- [ ] (optional?) Avoid call to jobs.get in certain cases, such as `QueryJob.to_dataframe` and `QueryJob.to_arrow`
  - Add "reload" argument to QueryJob.result() -- default to True.
  - Update RowIterator to call get_job to fetch the destination table ID before attempting use of BQ Storage API (if destination table ID isn't available).

tswast:
When we reintroduce row caching (reverted in https://github.com/googleapis/python-bigquery/pull/400), it should be opt-in to account for time/memory tradeoff https://github.com/googleapis/python-bigquery/issues/394 and longer calls to `result()` causing infinite loops in pandas-gbq https://github.com/pydata/pandas-gbq/issues/343 and the magics.

tswast:
I've conducted some microbenchmarks, and for a large class of row sizes, the BigQuery Storage API is faster than `getQueryResults`. Closing this work item out, as I don't think this pattern of long-running API requests is well suited to the typical use of the Python client (especially not with pandas DataFrames).

PR #361, opened 2020-11-03
refactor: break job into multiple modules

tswast:
Original paths are retained for backwards compatibility.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #358  🦕

PR #363, opened 2020-11-03
perf: use `jobs.getQueryResults` to download result sets

tswast:
Replaces `tabledata.list` when `RowIterator` is used for query results.
This likely also fixes a few edge cases around BigQuery scripting jobs.

Based on https://github.com/googleapis/python-bigquery/pull/361
Towards #362  🦕


google/cloud/bigquery/client.py
```diff
@@ -3169,6 +3170,77 @@ def list_rows(
...
+
+        row_iterator = RowIterator(
+            client=self,
+            api_request=functools.partial(self._call_api, retry, timeout=timeout),
+            path=f"/projects/{project}/queries/{job_id}",
+            schema=schema,
+            max_results=max_results,
+            page_size=page_size,
+            table=destination,
+            extra_params=params,
```
shollyman, line 3242: My assumption is that location gets plumbed through extra params when relevant?

shollyman (approved):
Looks good.  If you end up reusing _list_rows_from_query_results for the done polling you may need to consider that explicit project list, but it seems you're being distinct between the job polling and iteration usages here.

google/cloud/bigquery/client.py
tswast: Good catch. I don't believe we are populating location here.

tswast: populating in latest commit

PR #365, opened 2020-11-04
feat: added system test and sample for dataframe contains array

### LINKED ISSUE #19: BigQuery: Upload pandas DataFrame containing arrays
The support for python Bigquery API indicates that arrays are possible, however, when passing from a pandas dataframe to bigquery there is a pyarrow struct issue.

The only way round it seems its to drop columns then use JSON Normalise for a separate table.

```python

project = 'lake'
client = bigquery.Client(credentials=credentials, project=project)
dataset_ref = client.dataset('XXX')
table_ref = dataset_ref.table('RAW_XXX')
job_config.autodetect = True

client.load_table_from_dataframe(appended_data, table_ref,job_config=job_config).result()
```

This is the error recieved. NotImplementedError: struct

The reason I wanted to use this API as it indicates Nested Array support, which is perfect for our data lake in BQ but I assume this doesn't work?

HemangChothani:
Fixes #19

### DIFF
--- a/samples/load_table_dataframe_array_contains.py
+++ b/samples/load_table_dataframe_array_contains.py
@@ -1,0 +1,40 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def load_table_dataframe_array_contains(table_id):
+
+    # [START bigquery_load_table_dataframe_array_contains]
+
+    from google.cloud import bigquery
+    import pandas
+
+    # Construct a BigQuery client object.
+    client = bigquery.Client()
+
+    # TODO(developer): Set table_id to the ID of the table to create.
+    # table_id = "your-project.your_dataset.your_table_name"
+
+    dataframe = pandas.DataFrame({"A": [[1, 2, 3], [4, 5, 6], [7, 8, 9]]})
+    job = client.load_table_from_dataframe(dataframe, table_id)  # Make an API request.
+    job.result()  # Wait for the job to complete.
+
+    table = client.get_table(table_id)  # Make an API request.
+    print(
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -46,12 +46,12 @@
...
         "grpcio >= 1.32.0, < 2.0dev",
-        "pyarrow >= 1.0.0, < 2.0dev",
+        "pyarrow >= 2.0.0, < 3.0dev",
     ],
     "pandas": [
         "pandas>=0.23.0",
-        # pyarrow 1.0.0 is required for the use of timestamp_as_object keyword.
-        "pyarrow >= 1.0.0, < 2.0dev",
+        # pyarrow 2.0.0 is required for the use of arrays in dataframe to load the table .
+        "pyarrow >= 2.0.0, < 3.0dev",
```
tswast, line 54: Let's not bump the minimum version here. Most features do work with 1.0, and pyarrow is a core library that is very useful to have a wide range of support.

tests/system.py
```diff
@@ -1126,6 +1126,62 @@ def test_load_table_from_dataframe_w_struct_datatype(self):
...
+                None,
+                (
+                    bigquery.SchemaField(
+                        "list",
+                        "RECORD",
+                        "REPEATED",
+                        None,
+                        (
+                            bigquery.SchemaField(
+                                "item", "INTEGER", "NULLABLE", None, (), None
```
tswast, line 1154: Hmm... This is a bit of a surprising schema. It appears to match the behavior we were encountering previously. This feature is not supported if we cannot upload directly to a REPEATED INTEGER column.

tswast (changes requested):

samples/load_table_dataframe_array_contains.py
```diff
@@ -0,0 +1,40 @@
...
+    import pandas
+
+    # Construct a BigQuery client object.
+    client = bigquery.Client()
+
+    # TODO(developer): Set table_id to the ID of the table to create.
+    # table_id = "your-project.your_dataset.your_table_name"
+
+    dataframe = pandas.DataFrame({"A": [[1, 2, 3], [4, 5, 6], [7, 8, 9]]})
+    job = client.load_table_from_dataframe(dataframe, table_id)  # Make an API request.
```
tswast, line 30: Without an explicit schema, this sample is no different from the generic load_table_from_dataframe sample.

I was imagining system test XOR sample, as they are testing the same behavior.

tswast:
I've sent https://github.com/googleapis/python-bigquery/pull/368 to capture just the desired `setup.py` changes.

It's possible there are some kinds of arrays (such as arrays of records) that are supported, but it appears arrays of scalars still aren't handled correctly.

PR #378, opened 2020-11-10
feat: allow routine references

### LINKED ISSUE #353: Implement: Authorized UDF support for dataset Access
## Overview

Add support for authorized UDFs.

The Access property of a dataset now allows references to **Routine** entities, similar to the existing View support.  Only UDF routines are currently supported.  This feature is enabled and can be implemented immediately.

## References

Example impl: https://github.com/googleapis/google-cloud-go/pull/3084

What we document in the discovery layer for dataset's `access` field:
```
"access": {
          "type": "array",
          "description": "[Optional] An array of objects that define dataset access for one or more entities. You can set this property when inserting or updating a dataset in order to control who is allowed to access the data. If unspecified at dataset creation time, BigQuery adds default dataset access for the following entities: access.specialGroup: projectReaders; access.role: READER; access.specialGroup: projectWriters; access.role: WRITER; access.specialGroup: projectOwners; access.role: OWNER; access.userByEmail: [dataset creator email]; access.role: OWNER;",
          "items": {
            "properties": {
              "view": {
                "$ref": "TableReference",
                "description": "[Pick one] A view from a different dataset to grant access to. Queries executed against that view will have read access to tables in this dataset. The role field is not required when this field is set. If that view is updated by any user, access to the view needs to be granted again via an update operation."
              },
              "groupByEmail": {
                "description": "[Pick one] An email address of a Google Group to grant access to. Maps to IAM policy member \"group:GROUP\".",
                "type": "string"
              },
              "userByEmail": {
                "description": "[Pick one] An email address of a user to grant access to. For example: fred@example.com. Maps to IAM policy member \"user:EMAIL\" or \"serviceAccount:EMAIL\".",
                "type": "string"
              },
              "domain": {
                "description": "[Pick one] A domain to grant access to. Any users signed in with the domain specified will be granted the specified access. Example: \"example.com\". Maps to IAM policy member \"domain:DOMAIN\".",
                "type": "string"
              },
              "iamMember": {
                "type": "string",
                "description": "[Pick one] Some other type of member that appears in the IAM Policy but isn't a user, group, domain, or special group."
              },
              "role": {
                "description": "[Required] An IAM role ID that should be granted to the user, group, or domain specified in this access entry. The following legacy mappings will be applied: OWNER  roles/bigquery.dataOwner WRITER  roles/bigquery.dataEditor READER  roles/bigquery.dataViewer This field will accept any of the above formats, but will return only the legacy format. For example, if you set this field to \"roles/bigquery.dataOwner\", it will be returned back as \"OWNER\".",
                "type": "string"
              },
              "specialGroup": {
                "type": "string",
                "description": "[Pick one] A special group to grant access to. Possible values include: projectOwners: Owners of the enclosing project. projectReaders: Readers of the enclosing project. projectWriters: Writers of the enclosing project. allAuthenticatedUsers: All authenticated BigQuery users. Maps to similarly-named IAM members."
              },
              "routine": {
                "$ref": "RoutineReference",
                "description": "[Pick one] A routine from a different dataset to grant access to. Queries executed against that routine will have read access to views/tables/routines in this dataset. Only UDF is supported for now. The role field is not required when this field is set. If that routine is updated by any user, access to the routine needs to be granted again via an update operation."
              }
            },
            "type": "object"
          }
        },
```

Internal request issue: 170647386

cguardia:
The Access property of a dataset now allows references to Routine entities.

Fixes #353 🦕

### DIFF
--- a/google/cloud/bigquery/dataset.py
+++ b/google/cloud/bigquery/dataset.py
@@ -79,8 +79,9 @@ class AccessEntry(object):
     """Represents grant of an access role to an entity.
 
     An entry must have exactly one of the allowed :attr:`ENTITY_TYPES`. If
-    anything but ``view`` is set, a ``role`` is also required. ``role`` is
-    omitted for a ``view``, because ``view`` s are always read-only.
+    anything but ``view`` or ``routine`` are set, a ``role`` is also required.
+    ``role`` is omitted for ``view`` and ``routine``, because they are always
+    read-only.
 
     See https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets.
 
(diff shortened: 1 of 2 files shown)


tswast (approved):
Thanks!

PR #384, opened 2020-11-11
perf: avoid extra API calls from `to_dataframe` if all rows are cached

tswast:
Follow-up to previous PR, which cached the first page of
`getQueryResults`. If the first page is the only page (no `pageToken`),
then it is unnecessary to make extra API calls from `to_dataframe` or
`to_arrow` to the BigQuery Storage API.

Towards #362  🦕


shollyman (approved):
The rows are coming...from inside the house!

Issue #377, opened 2020-11-10
DBAPI - running a series of statements that depend on each other (DECLARE, SET, SELECT, ...)


mistercrunch:
see original issue here: https://github.com/mxmzdlv/pybigquery/issues/74

I'm a committer in Apache Superset and we integrate generally fairly well with BigQuery using the DBAPI driver provided here. Using the python DBAPI driver for BigQuery, we're having an issue running a chain of statements that depend on each other in a common session. I'm using a simple connection and cursor behind the scene.

It appears that the statements are not chained and the session/connection isn't persisted. Any clue on how to solve this? All other DBAPI 

Other DBAPI implementations allow this type of chaining, and using the same logic with this driver does not work properly.

Here's a simple script exemplifying this:

```sql
DECLARE d DATE DEFAULT CURRENT_DATE();
CREATE TEMP FUNCTION Add(x INT64, y INT64) AS (x + y);
SELECT Add(3, 4);
``` 

#### Environment details

  - OS type and version: Linux
  - Python version: 3.7.x
  - pip version: `20.2.4`
  - `google-cloud-bigquery` version: `pip show google-cloud-bigquery` 1.28.0

#### Steps to reproduce
code:
```python
with closing(engine.raw_connection()) as conn:
    with closing(conn.cursor()) as cursor:
        cursor.execute("DECLARE d DATE DEFAULT CURRENT_DATE()")
        cursor.execute("CREATE TEMP FUNCTION Add(x INT64, y INT64) AS (x + y)")
        cursor.execute("SELECT Add(3, 4)")
```

eugeniamz:
I confirm the issue. The Driver executes the statements in separate sessions, from the project history you can see that 

```
DECLARE d DATE DEFAULT CURRENT_DATE();
CREATE TEMP FUNCTION Add(x INT64, y INT64) AS (x + y);
```

are separated entries

<img width="938" alt="2020-11-10_07-20-08" src="https://user-images.githubusercontent.com/58375897/98674074-464b4700-2326-11eb-8d6e-a24d8fff10a2.png">

and because of executing as separated entries 

`CREATE TEMP FUNCTION Add(x INT64, y INT64) AS (x + y);
`
fails as temp function should be followed with a select as the error explains : 

<img width="666" alt="2020-11-10_07-20-49" src="https://user-images.githubusercontent.com/58375897/98673789-dd63cf00-2325-11eb-9bc6-49d5a0d94c36.png">


[...]

PR #386, opened 2020-11-12
fix(DBAPI): keep cursor hot when statement is a script

mistercrunch:
all the details here: https://github.com/googleapis/python-bigquery/issues/377

I guess technically a script isn't DML :trollface:

### DIFF
--- a/google/cloud/bigquery/dbapi/cursor.py
+++ b/google/cloud/bigquery/dbapi/cursor.py
@@ -222,7 +222,7 @@ class Cursor(object):
 
         is_dml = (
             self._query_job.statement_type
-            and self._query_job.statement_type.upper() != "SELECT"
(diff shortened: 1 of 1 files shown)


tswast:
Makes sense to me.

Actually, we can probably consider removing the `is_dml` logic altogether now that we moved to `getQueryResults` instead of `tabledata.list` (https://github.com/googleapis/python-bigquery/pull/375). The destination table of a DML query would return a non-null value that was unreadable or would return nonsense with DML (and some DDL) queries. Before we do that, I'll want to double-check that we have a system test for this case. (I think a DDL to CREATE VIEW would be a good choice for that)

mistercrunch:
Superseded by https://github.com/googleapis/python-bigquery/pull/387

PR #387, opened 2020-11-12
fix(dbapi): allow rows to be fetched from scripts

tswast:
The `is_dml` logic is not needed now that we moved to `getQueryResults` instead of `tabledata.list` (https://github.com/googleapis/python-bigquery/pull/375).

Previously, the destination table of a DML query would return a non-null value that was unreadable or would return nonsense with DML (and some DDL) queries. 

Towards #377  🦕

### DIFF
--- a/google/cloud/bigquery/dbapi/cursor.py
+++ b/google/cloud/bigquery/dbapi/cursor.py
@@ -220,14 +220,6 @@ class Cursor(object):
             self._query_data = iter([])
             return
 
-        is_dml = (
-            self._query_job.statement_type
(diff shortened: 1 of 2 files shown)


shollyman (approved):
Nice!

mistercrunch:
happy

mistercrunch:
Big thanks for making this happen, this enabled us to ship this: https://github.com/apache/incubator-superset/pull/11904

We're planning more work around creating a great experience around BigQuery in Superset, looking forward to more collaboration!

PR #393, opened 2020-11-19
fix: timestamp precision in insert_rows

### LINKED ISSUE #392: Timestamp loss of precision in insert_rows
Certain values of `datetime` lose a single microsecond on their way into BigQuery. This appears to happen because of the JSON rendering of timestamps as a floating point number of seconds.

#### Environment details

  - OS type and version: Ubuntu Bionic
  - Python version: 3.6.9
  - pip version: 9.0.1
  - `google-cloud-bigquery` version: 1.28.0

#### Steps to reproduce

Here is an example of a `datetime` value that loses a microsecond:
```py
datetime(2020, 11, 17, 1, 6, 52, 353795, tzinfo=datetime.timezone(datetime.timedelta(0), 'UTC'))
```

As a full integration repro, you could try something like this:

  1. Create a BigQuery table with a `TIMESTAMP` column.
  2. Use `insert_rows` to insert the value above.
  3. Inspect the results in the Console, where it will show as `2020-11-17 01:06:52.353794 UTC` instead of `2020-11-17 01:06:52.353795 UTC`.

#### Code example

An easier repro hints at the root cause. By looking at the output of `_timestamp_to_json_row`, it appears that the problematic values are rounded down to microsecond precision.

```python
import datetime
from google.cloud import bigquery as bq
ts = datetime.datetime(2020, 11, 17, 1, 6, 52, 353795, tzinfo=datetime.timezone(datetime.timedelta(0), 'UTC'))
print(bq._helpers._timestamp_to_json_row(ts))
```
This will print:
```
1605575212.3537948
```

Perhaps the solution is to add a half microsecond (0.5e-6) so that truncating results in rounding to the nearest microsecond.

HemangChothani:
Fixes #392

### DIFF
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -23,7 +23,7 @@ import six
 from google.cloud._helpers import UTC
 from google.cloud._helpers import _date_from_iso8601_date
 from google.cloud._helpers import _datetime_from_microseconds
-from google.cloud._helpers import _microseconds_from_datetime
+from google.cloud._helpers import _RFC3339_MICROS
 from google.cloud._helpers import _RFC3339_NO_FRACTION
 from google.cloud._helpers import _to_bytes
 
@@ -313,13 +313,9 @@ def _timestamp_to_json_parameter(value):
 
 
 def _timestamp_to_json_row(value):
-    """Coerce 'value' to an JSON-compatible representation.
-
-    This version returns floating-point seconds value used in row data.
-    """
+    """Coerce 'value' to an JSON-compatible representation."""
     if isinstance(value, datetime.datetime):
-        value = _microseconds_from_datetime(value) * 1e-6
-        value = round(value, 6)
+        value = value.strftime(_RFC3339_MICROS)
     return value
 
 
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -319,6 +319,7 @@ def _timestamp_to_json_row(value):
     """
     if isinstance(value, datetime.datetime):
         value = _microseconds_from_datetime(value) * 1e-6
+        value = round(value, 6)
     return value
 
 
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/_helpers.py
```diff
@@ -319,6 +319,7 @@ def _timestamp_to_json_row(value):
     """
     if isinstance(value, datetime.datetime):
         value = _microseconds_from_datetime(value) * 1e-6
+        value = round(value, 6)
```
tswast, line 322: I'm curious how this does anything. Doesn't `_microseconds_from_datetime` return an integer?

tswast (changes requested):
Please add a system test for this. It'll need to query the table that rows are inserted into in order to ensure the values roundtrip correctly.

google/cloud/bigquery/_helpers.py
HemangChothani: Yes, it returns an integer as we can see here https://github.com/googleapis/python-cloud-core/blob/5768ce33a182c97d9c3b2cd77d2fec6a39326b68/google/cloud/_helpers.py#L215

tests/unit/test__helpers.py
```diff
@@ -728,17 +728,17 @@ def test_w_string(self):
         self.assertEqual(self._call_fut(ZULU), ZULU)
 
     def test_w_datetime(self):
-        from google.cloud._helpers import _microseconds_from_datetime
+        from google.cloud._helpers import _RFC3339_MICROS
 
         when = datetime.datetime(2016, 12, 20, 15, 58, 27, 339328)
-        self.assertEqual(self._call_fut(when), _microseconds_from_datetime(when) / 1e6)
+        self.assertEqual(self._call_fut(when), when.strftime(_RFC3339_MICROS))
```
tswast, line 734: Please update these tests to compare against the actual string produced. I trust that `when.strftime(_RFC3339_MICROS)` is doing the right thing, but I'd be much more comfortable seeing the actual formatted string here and in the other unit tests.

tswast (approved):

PR #399, opened 2020-11-24
feat: support CSV format in `load_table_from_dataframe` pandas connector

### LINKED ISSUE #383: load_table_from_dataframe -- support alternative serialization formats
Currently, pandas-gbq doesn't use `load_table_from_dataframe` because it is using CSV for serialization, which has better support for TIME, DATE, DATETIME than Parquet. See: https://github.com/googleapis/python-bigquery/issues/56, https://github.com/googleapis/python-bigquery/issues/382

Struct / Array support could also be improved (though this is partially due to problems in the pyarrow library). https://github.com/googleapis/python-bigquery/issues/19

I propose supporting the CSV serialization type.

cguardia:
for load_table_from_dataframe

Fixes #383 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2111,8 +2111,10 @@ class Client(ClientWithProject):
 
         .. note::
 
-            Due to the way REPEATED fields are encoded in the ``parquet`` file
-            format, a mismatch with the existing table schema can occur, and
+            REPEATED fields are NOT supported when using the CSV source format.
+            They are supported when using the PARQUET source format, but
+            due to the way they are encoded in the ``parquet`` file,
+            a mismatch with the existing table schema can occur, so
             100% compatibility cannot be guaranteed for REPEATED fields when
             using the parquet format.
 
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2197,15 +2197,16 @@ class Client(ClientWithProject):
         else:
             job_config = job.LoadJobConfig()
 
-        if job_config.source_format:
-            if job_config.source_format != job.SourceFormat.PARQUET:
-                raise ValueError(
-                    "Got unexpected source_format: '{}'. Currently, only PARQUET is supported".format(
-                        job_config.source_format
-                    )
-                )
-        else:
+        supported_formats = (job.SourceFormat.CSV, job.SourceFormat.PARQUET)
+        if job_config.source_format is None:
+            # default value
             job_config.source_format = job.SourceFormat.PARQUET
+        if job_config.source_format not in supported_formats:
+            raise ValueError(
+                "Got unexpected source_format: '{}'. Currently, only PARQUET and CSV are supported".format(
+                    job_config.source_format
+                )
+            )
 
         if location is None:
             location = self.location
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/client.py
```diff
@@ -2197,15 +2197,16 @@ def load_table_from_dataframe(
...
 
-        if job_config.source_format:
-            if job_config.source_format != job.SourceFormat.PARQUET:
-                raise ValueError(
-                    "Got unexpected source_format: '{}'. Currently, only PARQUET is supported".format(
-                        job_config.source_format
-                    )
-                )
-        else:
+        supported_formats = (job.SourceFormat.CSV, job.SourceFormat.PARQUET)
```
tswast, line 2200: Nit: we can use a set literal here.

```suggestion
        supported_formats = {job.SourceFormat.CSV, job.SourceFormat.PARQUET}
```

google/cloud/bigquery/client.py
```diff
@@ -2245,39 +2246,71 @@ def load_table_from_dataframe(
                 stacklevel=2,
             )
 
-        tmpfd, tmppath = tempfile.mkstemp(suffix="_job_{}.parquet".format(job_id[:8]))
+        tmpfd, tmppath = tempfile.mkstemp(
+            suffix="_job_{}.{}".format(job_id[:8], job_config.source_format.lower(),)
```
tswast, line 2250: I think `flake8` will complain about the trailing comma here.

google/cloud/bigquery/client.py
```diff
@@ -2245,39 +2246,71 @@ def load_table_from_dataframe(
...
+                    _pandas_helpers.dataframe_to_parquet(
+                        dataframe,
+                        job_config.schema,
+                        tmppath,
+                        parquet_compression=parquet_compression,
+                    )
+                else:
+                    dataframe.to_parquet(tmppath, compression=parquet_compression)
+
+                with open(tmppath, "rb") as parquet_file:
```
tswast, line 2271: Isn't from here down the same between CSV and Parquet? In both cases we're calling `load_table_from_file`. The file object variable name `parquet_file` could be renamed to `tmpfile` or similar.

tests/system.py
```diff
@@ -1165,6 +1165,103 @@ def test_load_table_from_json_basic_use(self):
...
+            bigquery.SchemaField("float_col", "FLOAT"),
+            bigquery.SchemaField("geo_col", "GEOGRAPHY"),
+            bigquery.SchemaField("int_col", "INTEGER"),
+            bigquery.SchemaField("num_col", "NUMERIC"),
+            bigquery.SchemaField("str_col", "STRING"),
+            bigquery.SchemaField("time_col", "TIME"),
+            bigquery.SchemaField("ts_col", "TIMESTAMP"),
+        )
+        table_schema = scalars_schema + (
+            # TODO: Array columns can't be read due to NULLABLE versus REPEATED
```
tswast, line 1186: I don't expect ARRAY or STRUCT fields to ever be supported in CSV format. These TODOs can be removed.

tests/system.py
```diff
@@ -1165,6 +1165,103 @@ def test_load_table_from_json_basic_use(self):
...
+                ),
+                (
+                    "dt_col",
+                    [
+                        datetime.datetime(1, 1, 1, 0, 0, 0),
+                        None,
+                        datetime.datetime(9999, 12, 31, 23, 59, 59, 999999),
+                    ],
+                ),
+                ("float_col", [float("-inf"), float("nan"), float("inf")]),
```
tswast, line 1210: I'd like to see some floating point values with lots of precision, similar to those I added in https://github.com/pydata/pandas-gbq/pull/336

We'll probably want to test with > 3 rows in that case. Perhaps those floating point values can get their own system test.

cguardia:
@tswast sorry for the delay. I think this addresses your comments.

google/cloud/bigquery/client.py
```diff
@@ -2113,7 +2113,8 @@ def load_table_from_dataframe(
 
             Due to the way REPEATED fields are encoded in the ``parquet`` file
             format, a mismatch with the existing table schema can occur, and
-            100% compatibility cannot be guaranteed for REPEATED fields.
+            100% compatibility cannot be guaranteed for REPEATED fields when
+            using the parquet format.
```
tswast, line 2117: This could mislead people to think that REPEATED fields are supported for CSV (not that it would make any sense). Perhaps we need a second note describing that only scalar data types are supported with CSV?

cguardia: @tswast I updated the note to try to make that clear.

tswast (approved):

PR #374, opened 2020-11-05
perf: cache first page of `jobs.getQueryResults` rows

tswast:
Replaces and simplifies #347 . The cache is only used if `max_results`, `page_size`, and `start_index` are not set.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Towards #362 🦕

PR #400, opened 2020-11-24
perf: don't fetch rows when waiting for query to finish

### LINKED ISSUE #394: increased memory usage in 2.4.0
Version 2.4.0 of the library is allocating much more memory that the previous version, 2.3.1, when running multiple queries.
In particular, it seems that the `QueryJob` object is retaining the results of the query internally, and that memory is not deallocated.

I think that the problem is related to https://github.com/googleapis/python-bigquery/pull/374.

#### Environment details

  - macOS 11.0.1 (also observing this on Linux in a production environment)
  - Python version: 3.8.6
  - pip version: 20.1.1 
  - `google-cloud-bigquery` version: 2.4.0

#### Steps to reproduce

Run the script in the code example with `google-cloud-bigquery` 2.4.0 and 2.3.1 versions.
You will also need to install:
```
google-cloud-bigquery-storage==2.1.0
pandas==1.1.4
psutil==5.7.3
```

The outputs on my machine are:

With 2.4.0:
```
Initial memory used: 77 MB
Memory used: 642 MB
Memory used: 875 MB
Memory used: 1117 MB
Memory used: 1342 MB
Memory used: 1568 MB
Memory used: 1792 MB
Memory used: 2039 MB
Memory used: 2265 MB
Memory used: 2505 MB
Memory used: 2725 MB
```

With 2.3.1:

```
Initial memory used: 77 MB
Memory used: 97 MB
Memory used: 98 MB
Memory used: 99 MB
Memory used: 99 MB
Memory used: 99 MB
Memory used: 99 MB
Memory used: 100 MB
Memory used: 101 MB
Memory used: 101 MB
Memory used: 101 MB
```

#### Code example

Please note that we are storing a reference to the `QueryJob` objects, but **not** to the resulting DataFrames.

```python
import os

import psutil

if __name__ == '__main__':
    client = bigquery.Client()

    process = psutil.Process(os.getpid())
    print(f"Initial memory used: {process.memory_info().rss / 1e6:.0f} MB")

    jobs = []

    for i in range(10):
        job = client.query("SELECT x FROM UNNEST(GENERATE_ARRAY(1, 1000000)) AS x")
        job.result().to_dataframe()
        jobs.append(job)
        print(f"Memory used: {process.memory_info().rss / 1e6:.0f} MB")
```

tswast:
When there are large result sets, fetching rows while waiting for the
query to finish can cause the API to hang indefinitely. (This may be due
to an interaction between connection timeout and API timeout.)

This reverts commit 86f6a516d1c7c5dc204ab085ea2578793e6561ff (#374).

Fixes https://github.com/pydata/pandas-gbq/issues/343
Fixes #394 🦕

### DIFF
--- a/tests/unit/test_table.py
+++ b/tests/unit/test_table.py
@@ -1630,6 +1630,40 @@ class TestRowIterator(unittest.TestCase):
 
         api_request.assert_called_once_with(method="GET", path=path, query_params={})
 
+    def test_iterate_with_cached_first_page(self):
+        from google.cloud.bigquery.schema import SchemaField
+
+        first_page = {
+            "rows": [
+                {"f": [{"v": "Whillma Phlyntstone"}, {"v": "27"}]},
+                {"f": [{"v": "Bhetty Rhubble"}, {"v": "28"}]},
+            ],
+            "pageToken": "next-page",
+        }
+        schema = [
+            SchemaField("name", "STRING", mode="REQUIRED"),
+            SchemaField("age", "INTEGER", mode="REQUIRED"),
+        ]
+        rows = [
+            {"f": [{"v": "Phred Phlyntstone"}, {"v": "32"}]},
+            {"f": [{"v": "Bharney Rhubble"}, {"v": "33"}]},
+        ]
+        path = "/foo"
+        api_request = mock.Mock(return_value={"rows": rows})
+        row_iterator = self._make_one(
+            _mock_client(), api_request, path, schema, first_page_response=first_page
+        )
+        rows = list(row_iterator)
+        self.assertEqual(len(rows), 4)
+        self.assertEqual(rows[0].age, 27)
+        self.assertEqual(rows[1].age, 28)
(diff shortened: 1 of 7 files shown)


tests/unit/test_table.py
```diff
@@ -1630,6 +1630,40 @@ def test_iterate(self):
 
         api_request.assert_called_once_with(method="GET", path=path, query_params={})
 
+    def test_iterate_with_cached_first_page(self):
```
tswast, line 1633: I add these tests to fix coverage.

Perhaps instead I should also be reverting the changes from https://github.com/googleapis/python-bigquery/pull/384 ?

shollyman: The alternative I'd consider is doing a flag-based feature, so users can turn on the latency changes (or flip default, and provide an opt-out while we diagnose these other cases).

shollyman (approved):

PR #401, opened 2020-11-25
feat: add support for unrecognized model types

### LINKED ISSUE #334: Make Model constructor robust against new model types
Follow-up to https://github.com/googleapis/python-bigquery/issues/293

[model_type](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.model.Model.html#google.cloud.bigquery.model.Model.model_type) appears to be the only enum. Maybe we pull this out of the `_proto` resource and save it as a string, instead?

Alternatively, maybe if we fail to parse, we can set to MODEL_TYPE_UNSPECIFIED?

steffnay:
If fails to parse model type enum, sets to MODEL_TYPE_UNSPECIFIED.

Fixes #334  🦕

### DIFF
--- a/google/cloud/bigquery/model.py
+++ b/google/cloud/bigquery/model.py
@@ -310,8 +310,7 @@ class Model(object):
                 resource, types.Model()._pb, ignore_unknown_fields=True
             )
         except json_format.ParseError:
-            if this.model_type == 0:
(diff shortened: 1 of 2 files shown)


tswast (approved):
Wonderful!

PR #402, opened 2020-11-25
fix: preserve timestamp microsecond precision with rows from REST API

### LINKED ISSUE #395: Use DataFormatOptions to download timestamps with full precision

When fetching timestamp values, the [tabledata.list](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list) and [jobs.getQueryResults](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) APIs return a floating point value by default. 

This can cause [unexpected floating point issues](https://issuetracker.google.com/issues/147629356) when rounding timestamps in SQL, such as with the `TIMESTAMP_TRUNC(COL_TIMESTAMP, MILLISECOND)` function.

The backend API adds a [`formatOptions.useInt64Timestamp`](https://cloud.google.com/bigquery/docs/reference/rest/v2/DataFormatOptions#FIELDS.use_int64_timestamp) option. 

* Wherever `tabledata.list`, `jobs.query`, or `jobs.getQueryResults` are called, set the `formatOptions.useInt64Timestamp` option to `True`.
* Wherever timestamps in rows API responses are handled, update it to use integer math to extract microseconds and seconds.
* Add system tests to verify that timestamps rounded to every digit of supported precision (e.g. 9, 9.9, 9.99, ..., 9.999999) are returned exactly as set in the table/query.

N/A

 **Additional context**

Precision issues for the other direction (to BigQuery) for similar floating-point reasons: https://github.com/googleapis/python-bigquery/issues/392

HemangChothani:

Fixes #395

### DIFF
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -81,19 +81,18 @@ def _bytes_from_json(value, field):
 def _timestamp_from_json(value, field):
     """Coerce 'value' to a datetime, if set or not nullable."""
     if _not_null(value, field):
-        # value will be a float in seconds, to microsecond precision, in UTC.
+        # value will be a integer in seconds, to microsecond precision, in UTC.
 
-        if value is not None:
-            if type(value) == int:
-                value = value
-            elif type(value) == float:
-                value = 1e6 * float(value)
-            elif value.isdigit():
-                value = int(value)
-            else:
-                value = 1e6 * float(value)
-
-        return _datetime_from_microseconds(value)
+        # if value is not None:
+        #     if type(value) == int:
+        #         value = value
+        #     elif type(value) == float:
+        #         value = 1e6 * float(value)
+        #     elif value.isdigit():
+        #         value = int(value)
+        #     else:
+        #         value = 1e6 * float(value)
+        return _datetime_from_microseconds(int(value))
 
 
 def _timestamp_query_param_from_json(value, field):
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -82,7 +82,18 @@ def _timestamp_from_json(value, field):
     """Coerce 'value' to a datetime, if set or not nullable."""
     if _not_null(value, field):
         # value will be a float in seconds, to microsecond precision, in UTC.
-        return _datetime_from_microseconds(1e6 * float(value))
+
+        if value is not None:
+            if type(value) == int:
+                value = value
+            elif type(value) == float:
+                value = 1e6 * float(value)
+            elif value.isdigit():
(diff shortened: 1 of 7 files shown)


google/cloud/bigquery/_helpers.py
```diff
@@ -82,7 +82,18 @@ def _timestamp_from_json(value, field):
     """Coerce 'value' to a datetime, if set or not nullable."""
     if _not_null(value, field):
         # value will be a float in seconds, to microsecond precision, in UTC.
-        return _datetime_from_microseconds(1e6 * float(value))
```
tswast, line 85: This will need to be modified to parse from an rfc timestamp string.

HemangChothani: Sorry, i didn't get the point. Before passing `params["formatOptions.useInt64Timestamp"] = True` in `client.list_row`, received a value in string with decimal  in `_timestamp_from_json`  method , but after passing it, received a value in string without decimal and call the method `_datetime_from_microseconds(int(value))` which returns datetime object.

tswast: Oh I see. Yes, that's expected.

tswast: In this case, we should update the comment.

Also, are there cases where floating point values are still passed in? If not, we should clean this up now. If so, let's add a TODO to identify those cases and file an issue to clean them up.

google/cloud/bigquery/_helpers.py
```diff
@@ -82,7 +82,18 @@ def _timestamp_from_json(value, field):
...
+
+        if value is not None:
+            if type(value) == int:
+                value = value
+            elif type(value) == float:
+                value = 1e6 * float(value)
+            elif value.isdigit():
+                value = int(value)
+            else:
+                value = 1e6 * float(value)
```
tswast, line 94: Why is this case needed? Couldn't the above `isdigit()` check be removed and we always fall back to converting to integer? I don't expect floating point values to be serialized as strings in JSON.

google/cloud/bigquery/_helpers.py
```diff
@@ -81,8 +81,18 @@ def _bytes_from_json(value, field):
 def _timestamp_from_json(value, field):
     """Coerce 'value' to a datetime, if set or not nullable."""
     if _not_null(value, field):
-        # value will be a float in seconds, to microsecond precision, in UTC.
-        return _datetime_from_microseconds(1e6 * float(value))
+        # value will be a integer in seconds, to microsecond precision, in UTC.
+
+        # if value is not None:
```
tswast, line 86: Please remove commented out code.

tswast (approved):

PR #409, opened 2020-12-01
chore: require samples checks

busunkim96:
Make samples kokoro sessions required


busunkim96:
@tswast Are there any other sessions (snippets, system?) that should be marked required?

tswast:
Snippets also get included in cloud.google.com docs, so makes sense to treat them the same as samples.

tswast (approved):

PR #420, opened 2020-12-07
docs: refresh view samples

tswast:
These snippets are copied from snippets.py and refreshed
to include imports and client construction. Will follow up
with a PR to remove the redundant snippets once the docs
on cloud.google.com are updated.

I notice that the tests for these snippets have been skipped.
I believe that was due to a temporary rollout issue with
table update. If that actually is flakier than I thought, we
can more easily add retries or disable just updates in this
new structure.

https://github.com/googleapis/python-bigquery/blob/96a1c5b3c72855ba6ae8c88dfd0cdb02d2faf909/docs/snippets.py#L487

Fixes internal issue 173133642 🦕

### DIFF
--- a/samples/snippets/conftest.py
+++ b/samples/snippets/conftest.py
@@ -1,0 +1,27 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


tmatsuo:
Re: snippet-bot's comment

I guess you're moving samples under `samples` dir. The bot is warning because some region tags are referenced from the devsite doc. This will cause 404 errors on our devsite pages.

I think the correct course of action is to 1) duplicate the sample, 2) update devsite, then 3) remove the old tags.

tswast:
Thanks @tmatsuo for looking. Yes, I am trying to move these samples. I'll try your suggested process.

shollyman (approved):

PR #423, opened 2020-12-08
ci: skip docfx in main 'Kokoro' presubmit

busunkim96:
The `docfx` build is failing in continuous (reasons are unclear as we can't seem to reproduce the error locally). [log](https://source.cloud.google.com/results/invocations/8c0ad1bc-87fc-40ef-b730-484000fbd57e/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fdocs%2Fdocs/log)

This PR makes the following changes:
* skip the `docfx` session in the main `Kokoro` presubmit by not listing it in the default sessions list
* use `NOX_SESSION` env var to run `docs` and `docfx` in `docs-presubmit` (needs an internal config change in 346339439)

After this is approved I will open a PR in synthtool to make the changes permanent.

PR #425, opened 2020-12-09
ci: use python3 instead of python3.6 in build.sh

busunkim96:


tswast:
Looks like `docs-presubmit` is trying to run the default nox sessions. I sent internal CL 348695855 to add NOX_SESSION to the allowed env vars.

busunkim96:
docs-presubmit is now properly running just the docs jobs.

plamut:
I see the _"no python3.6 error"_ is now gone, but the docs check now fails with a different error - is that in scope of this PR, too?

(FWIW, the `docfx` nox session succeeds locally)

busunkim96:
@plamut I opened #423 and this PR to properly split off docfx (not run it in required check `Kokoro` build since it will fail and block merge). We're tracking the  `docfx` failure in https://github.com/googleapis/doc-templates/issues/94.

tswast (approved):

PR #428, opened 2020-12-09
docs: add GEOGRAPHY data type code samples

tswast:
These are added to a separate directory in order to isolate the GeoJSON
and WKT dependencies from the other code samples.

Towards internal 171804665


tests/system.py
```diff
@@ -2414,9 +2414,8 @@ def test_querying_data_w_timeout(self):
 
         query_job = Config.CLIENT.query(
             """
-            SELECT name, SUM(number) AS total_people
-            FROM `bigquery-public-data.usa_names.usa_1910_current`
-            GROUP BY name
+            SELECT COUNT(*)
```
tswast, line 2417: Not really associated with this sample, but the system test flaked, so I figured I'd swap out this query with one from go/bq-long-queries

shollyman: Sounds good.

samples/geography/noxfile.py
```diff
@@ -0,0 +1,246 @@
...
+from __future__ import print_function
+
+import os
+from pathlib import Path
+import sys
+
+import nox
+
+
+# WARNING - WARNING - WARNING - WARNING - WARNING
```
shollyman, line 24: This file is generated?

shollyman (approved):
I was today years old when I found out that streaming inserts will accept either WKT or geojson as the field contents.

samples/geography/noxfile.py
tswast: Yerp. Copied from here: https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_samples/noxfile.py.j2

PR #442, opened 2020-12-21
fix: use debug logging level for OpenTelemetry message

### LINKED ISSUE #441: Too noise logging about telemetry
Hello,

In the Apache Airflow project, we use the BigQuery library, but recently we've started to see annoying log message when the library is loaded. It is enough that the library is loaded and there is an message every time. 

In my opinion, this message should be of a lower level (DEBUG) so that it is not displayed much less often or is displayed only when the client is initialized. 
```
import logging
logging.basicConfig(level=logging.INFO)

```
Output: 
```
INFO:google.cloud.bigquery.opentelemetry_tracing:This service is instrumented using OpenTelemetry. OpenTelemetry could not be imported; please add opentelemetry-api and opentelemetry-instrumentation packages in order to get BigQuery Tracing data.
```

Related issue: https://github.com/apache/airflow/issues/13131

CC: @tswast

tswast:

Fixes #441 🦕

### DIFF
--- a/google/cloud/bigquery/opentelemetry_tracing.py
+++ b/google/cloud/bigquery/opentelemetry_tracing.py
@@ -23,11 +23,11 @@ try:
     from opentelemetry.trace.status import Status
 
     HAS_OPENTELEMETRY = True
-    WARNED_OPENTELEMETRY = True
+    _warned_telemetry = True
 
 except ImportError:
     HAS_OPENTELEMETRY = False
-    WARNED_OPENTELEMETRY = False
+    _warned_telemetry = False
 
 _default_attributes = {
     "db.system": "BigQuery"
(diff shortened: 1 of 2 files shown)


tswast:
Updated based on https://github.com/apache/airflow/issues/13131 to not warn at import time.

mik-laj (approved):

shollyman (approved):

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -64,8 +59,18 @@ def create_span(name, attributes=None, client=None, job_ref=None):
...
     final_attributes = _get_final_span_attributes(attributes, client, job_ref)
     if not HAS_OPENTELEMETRY:
+        if not WARNED_OPENTELEMETRY:
+            logger.debug(
+                "This service is instrumented using OpenTelemetry. "
+                "OpenTelemetry could not be imported; please "
+                "add opentelemetry-api and opentelemetry-instrumentation "
+                "packages in order to get BigQuery Tracing data."
+            )
+            WARNED_OPENTELEMETRY = True
```
plamut, line 72: (nit) It's a bit unexpected to change a "constant", perhaps rename it? Also, since it's an internal state-keeping flag that users should neither know about nor modify it, we should probably mark it as such (e.g. `_warned_telemetry`)

plamut (commented):
I agree that the warning should not be issued at import time, but instead when trying to actually use BigQuery ([discussion](https://github.com/apache/airflow/issues/13131)).

DEBUG level is probably indeed better, as `opentelemetry` is not used by default (it's listed under extras), thus shouldn't bother users that don't choose to install it.

We should choose a different name for the flag, though, IMHO.

google/cloud/bigquery/opentelemetry_tracing.py
tswast: Good call. I've renamed in the latest commit.

PR #445, opened 2020-12-22
feat: add type hint for public methods

### LINKED ISSUE #157: feat: add type hints to methods
Having typings in the library would help and saw that there is a PR opened for that. However, the documentation is somewhat lacking.

It is hard to tell what the methods return and are all the parameters while looking at the docs https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html

For example, in the tutorial: what is the difference between these 2 table objects created?
```
table = bigquery.Table(table_id, schema=schema)
table = client.create_table(table)  # Make an API request.
```

HemangChothani:
Fixes #157


setup.cfg
```diff
@@ -17,3 +17,13 @@
...
+[pytype]
+python_version = 3.8
+inputs =
+    google/cloud/
+exclude =
+    tests/
+    google/cloud/bigquery_v2/
+output = .pytype/
+# Workaround for https://github.com/google/pytype/issues/150
+disable = pyi-error
```
HemangChothani, line 29: This file is synthtool generated, but for the reference and pass current tests changed the current file.

google/cloud/bigquery/client.py
```diff
@@ -75,6 +75,19 @@
 from google.cloud.bigquery.table import TableReference
 from google.cloud.bigquery.table import RowIterator
 
+# Types needed only for Type Hints
+import datetime
+import pandas
```
tswast, line 80: Is there any way we can avoid this import? I would like to keep pandas as an optional dependency.

I recall sometimes strings can be used for type hints.

google/cloud/bigquery/client.py
```diff
@@ -212,7 +225,10 @@ def close(self):
         self._http.close()
 
     def get_service_account_email(
-        self, project=None, retry=DEFAULT_RETRY, timeout=None
+        self,
+        project: str = None,
+        retry: retries.Retry = DEFAULT_RETRY,
+        timeout: float = None,
     ):
```
tswast, line 232: I recall there's a way to annotate the return type, too.

HemangChothani: Yes, i will specify return type  for all the public methods.

google/cloud/bigquery/client.py
HemangChothani: I have tried, but lint was failing so removed pandas type hints

setup.cfg
tswast: https://github.com/google/pytype/issues/150 is marked as closed. Can you provide a little more context as to why this is needed?

tswast (commented):
Looking good, thanks! A few questions.

plamut:
Bump, just checking the status of this?

Generally speaking, though, type hints would be welcome, even though static type checks have not been added to Kokoro (yet).

plamut:
Superseded by #613.

PR #456, opened 2020-12-30
chore: add constraints file

busunkim96:
Add constraints file to test lower bounds. These files will not be used until the noxfile is changed in googleapis/synthtool#869.


rohitpc363 (approved):

tswast (approved):
Thanks!

PR #458, opened 2021-01-07
refactor: simplify AutoStrEnum definition

### LINKED ISSUE #457: Simplify AutoStrEnum definition in Jupyter magic's argument parser
The current definition of [AutoStrEnum](https://github.com/googleapis/python-bigquery/blob/0337ea0bde966c3ccb94960493a6fa6f2bee49b4/google/cloud/bigquery/magics/line_arg_parser/lexer.py#L139-L167) is rather complicated. With the library now only supporting Python 3.6+, we can simplify it and just use the `_generate_next_value_()` hook.

plamut:
Closes #457.

With now only Python 3.6 supported, we can use the `_generate_next_value()` hook instead of metaclass magic. No new tests, as the existing tests already cover this code.

PR checklist:

### DIFF
--- a/google/cloud/bigquery/magics/line_arg_parser/lexer.py
+++ b/google/cloud/bigquery/magics/line_arg_parser/lexer.py
@@ -136,40 +136,17 @@ token_types = OrderedDict(
 )
 
 
-# The _generate_next_value_() enum hook is only available in Python 3.6+, thus we
-# need to do some acrobatics to implement an "auto str enum" base class. Implementation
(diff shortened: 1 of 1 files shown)


plamut:
I'm not entirely sure that the docs-presubmit failure is about, something about the missing `python3.6` command when setting up the tests.

I see the same error on other PRs, thus it doesn't seem to be caused by the changes here.

tswast (approved):
Nice!

PR #461, opened 2021-01-07
chore: Remove six dependency

### LINKED ISSUE #460: Remove six dependency
Now that we don't have to support Python 2.7 anymore, we can get rid of `six`.

If the number of cleanup changes turns out to be very high, it is probably better to remove `six` gradually, in which case this ticket should be treated merely as a reminder.

plamut:
Closes #460.

This PR gets rid of `six` that we don't need anymore after dropping Python 2 support.

PR checklist:


tswast (approved):
Thanks!

PR #465, opened 2021-01-08
chore: Bound maximum supported Python version

### LINKED ISSUE #464: Bound the maximum supported Python version
Related: #462.

The BigQuery library does not support Python 3.9 yet, as the `bqstorage` and `pandas` extras are blocked by the `pyarrow` dependency. Pyarrow [added](https://issues.apache.org/jira/browse/ARROW-10224) the Python 3.9 support, but the change has not been released yet (awaiting `pyarrow==3.0.0`).

We need to reflect this in `setup.py`'s ``python_requires`.

plamut:
Closes #464.

Bounding the Python versions gives a much more clear error message on installation attempts, as opposed to compile errors when trying to install `pyarrow` from source (pulls in the `numpy==1.16.0` dependency which fails to compile).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -52,7 +52,7 @@ dependencies.
 
 Supported Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^
-Python >= 3.6
+Python >= 3.6, < 3.9
 
 Unsupported Python Versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


tswast (approved):
I have mixed feelings about this, but I think it's the right call.

Interestingly, there are conda builds for several versions of pyarrow, just not wheels https://anaconda.org/conda-forge/pyarrow/files

PR #467, opened 2021-01-12
fix: QueryJob.exception() *returns* the errors, not raises them

### LINKED ISSUE #451: `google.cloud.bigquery.job.QueryJob.exception` has wrong behavior
Hello,

According to the [documentation](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJob.html#google.cloud.bigquery.job.QueryJob.exception) of `google.cloud.bigquery.job.QueryJob.exception`, this method should returns an exception (and not raises) if something went wrong. But according to my tests this is not always the case:

```python

client = bigquery.Client()

# Correct behavior: returns an exception and does not raise.
query = 'WRONG QUERY'
job = client.query(query)
exc = job.exception()

# Wrong behavior: raises an exception instead of returning it.
query = 'ASSERT (SELECT 2 > 3) AS "Error !!!"'
job = client.query(query)
try:
    job.exception()
except:
    print('raised!')
```

Am I misunderstanding the documentation or is this the correct behavior?

I'm using `google-cloud-bigquery==2.6.1` with Python 3.8.6. Let me know if you need more informations.

Thanks for your help,

plamut:
Fixes #451.

This PR makes sure that `QueryJob.exception()` actually _returns_ exceptions as documented, not raises them.

Checklist:


google/cloud/bigquery/job/query.py
```diff
@@ -988,15 +988,25 @@ def done(self, retry=DEFAULT_RETRY, timeout=None, reload=True):
...
-        self._reload_query_results(retry=retry, timeout=timeout)
+        try:
+            self._reload_query_results(retry=retry, timeout=timeout)
+        except Exception as exc:
+            # The job state might still be "RUNNING", but if an exception bubbles
+            # up (either a RetryError or a non-retryable error), we declare that we
+            # are done for good (and the blocking pull should not continue polling).
+            # Ditto in a similar try-except block below.
+            self.set_exception(exc)
+            return True
```
plamut, line 1009: Reviewers: Please confirm that returning `True` is fine in such case.

If returning `self.state == _DONE_STATE`, the query from the issue description can cause a long retry loop, as the backend seems to always return the state `RUNNING` for the job, even though an error in the query can (presumably?) be detected immediately and there's no need to keep the query running.

tswast: I think we need to reload the job object in this case. `jobs.get` has different error conditions than `jobs.getQueryResults`.

tswast: That is, `jobs.get` (i.e. `self.reload()`) should return successfully, even for failed jobs.

plamut: IIRC I tried that, but the state was still "RUNNING". I can check once more tomorrow, just in case.

Should `self.reload()` indeed not work, what would be the preferred backup plan?

google/cloud/bigquery/job/query.py
```diff
@@ -989,19 +989,30 @@ def done(self, retry=DEFAULT_RETRY, timeout=None, reload=True):
 
         Returns:
             bool: True if the job is complete, False otherwise.
+
+        Raises:
+            google.api_core.exceptions,GoogleAPICallError:
+                If a non-retryable error ocurrs while reloading the job metadata,
+                or if too many retry attempts fail.
```
plamut, line 996: Despite catching and handling reload-query-results errors, we should nevertheless let any job reload errors bubble up, IMO. We don't really know how to handle them and what to return if we cannot even get job info in the first place.

google/cloud/bigquery/job/query.py
```diff
@@ -989,19 +989,30 @@ def done(self, retry=DEFAULT_RETRY, timeout=None, reload=True):
...
         # If an explicit timeout is not given, fall back to the transport timeout
         # stored in _blocking_poll() in the process of polling for job completion.
         transport_timeout = timeout if timeout is not None else self._transport_timeout
 
+        try:
+            self._reload_query_results(retry=retry, timeout=transport_timeout)
+        except exceptions.GoogleAPIError:
+            # Reloading also updates error details on self, no need for an
+            # explicit self.set_exception() call.
+            self.reload(retry=retry, timeout=transport_timeout)
```
plamut, line 1013: Perhaps this should be changed to `exceptions.GoogleAPICallError` so that `RetryError` bubbles up immediately? Or could it still be beneficial to call `self.reload()` even if query results could not be fetched?

tswast: > Or could it still be beneficial to call self.reload() even if query results could not be fetched?

Yes, this would be beneficial, as the job resource will contain extra information about why the job failed.

google/cloud/bigquery/job/query.py
tswast: Hmmm... This is a departure from https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.done which doesn't throw any exceptions.

I agree that bubbling up the error would give more information to the user that we otherwise wouldn't know how to interpret, though. The alternative would be to return `False`, but then I think we risk infinite loops by our blocking poll that uses this method.

Or can we call `set_exception` and return False in this case? We'd want a unit test that confirms `result()` doesn't get into an infinite loop.

tswast (changes requested):

google/cloud/bigquery/job/query.py
plamut: OK, I'll leave it as is then, thanks.

tswast (approved):
Thanks!

tswast (changes requested):
Retracting approval until

> Or can we call set_exception and return False in this case? We'd want a unit test that confirms result() doesn't get into an infinite loop.

is handled.

google/cloud/bigquery/job/query.py
plamut: As discussed in the chat, returning `False` would cause the `done()` method to be repeated when run as part of a `_blocking_poll()`, even if exception was set on the job. However, we can still set the exception and return `True` to terminate the polling.

Terminating the polling is actually fine, because `self._reload_query_results()` and `self.reload()` already do their own retry cycle and if that has failed, we should give up and signal that the job instance - which is actually _future_ - is done, even if the result of the future is an exception.

We must also catch any errors that happen on job reload itself and set the job/future state accordingly.

google/cloud/bigquery/job/query.py
```diff
@@ -988,25 +988,41 @@ def done(self, retry=DEFAULT_RETRY, timeout=None, reload=True):
...
 
+        try:
+            self._reload_query_results(retry=retry, timeout=transport_timeout)
+        except exceptions.GoogleAPIError:
+            # Reloading also updates error details on self, thus no need for an
+            # explicit self.set_exception() call if reloading succeeds.
+            try:
+                self.reload(retry=retry, timeout=transport_timeout)
+            except exceptions.GoogleAPIError as exc:
+                self.set_exception(exc)
```
plamut, line 1012: Thought - should that be the original exception that made us reload the job metadata, i.e. the one cause by `_reload_query_results()`?

plamut:
Hmm, `TestQueryJob.test_result_w_retry` seems to be flaky ([failed run](https://source.cloud.google.com/results/invocations/eaaf9106-140d-4a44-ad23-b549799bd998/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fpresubmit/log)), it passes locally - let's see if it fails on a re-run.

google/cloud/bigquery/job/query.py
plamut, line 1012: Thought - should that be the original exception that made us reload the job metadata, i.e. the one caused by `_reload_query_results()`?

plamut:
I was able to reproduce flakiness on my machine by reducing the retry deadline in affected tests by ~25 %, which means the latter was indeed the culprit. Because `done()` now does a bit of extra work, it appears that was just enough of a slowdown to start causing the failures (at least under Python 3.7).

Although not ideal, increasing the deadline should be a solution sufficient with high enough probability.

tests/unit/job/test_query.py
```diff
@@ -356,6 +357,84 @@ def test_done_w_timeout_and_longer_internal_api_timeout(self):
...
+            self._properties["status"]["state"] = "DONE"
+            self.set_exception(copy.copy(bad_request_error))
+
+        fake_reload_method = types.MethodType(fake_reload, job)
+
+        with mock.patch.object(job, "reload", new=fake_reload_method):
+            is_done = job.done()
+
+        assert is_done
+        assert isinstance(job._exception, exceptions.BadRequest)
```
tswast, line 379: why use the private property in this and the other tests? any objections to calling `job.exception()` here and the other tests?

google/cloud/bigquery/job/query.py
plamut: Makes sense, the original exception should indeed be considerably more informative.

tests/unit/job/test_query.py
plamut: The reasoning was that `job.exception()` can execute additional logic, and errors in that method would make the tests for `done()` fail, too, even if there was nothing wrong with the `done()` method itself.

One could argue that the chosen unit of test is too small, and that the class itself should represent a unit as opposed to its individual methods, but addressing that would require quite some refactoring (we already tinker with internal `_properties` , for instance). 

Here, practicality beats purity IMHO, thus the "cheating" by examining the internal state of the class. Or do you have a strong opinion on this?

tswast: Works for me. I agree that ideally we'd have higher-level tests than this, but makes sense to stay with existing conventions, especially given our 100% coverage requirement.

tswast (approved):

PR #471, opened 2021-01-14
fix: reading the labels attribute on Job instances

### LINKED ISSUE #469: Reading the labels set on a job fails
The ability to specify labels on created BigQuery jobs was supposedly added in googleapis/google-cloud-python#5654, but it doesn't work. There is also a comment in the documentation from that same PR referring to Client.update_job() which doesn't even exist.

#### Environment details

# python --version
Python 3.7.3

# pip freeze | grep google
google-api-core==1.24.1
google-auth==1.24.0
google-cloud-bigquery==2.6.2
google-cloud-bigtable==1.6.1
google-cloud-core==1.5.0
google-cloud-pubsub==2.2.0
google-cloud-storage==1.35.0
google-crc32c==1.1.0
google-resumable-media==1.2.0
googleapis-common-protos==1.52.0
grpc-google-iam-v1==0.12.3

#### Code example

```python
job_config = bigquery.QueryJobConfig(labels={"app": "test"})
query_job = client.query("SELECT 1", job_config=job_config)

# no labels seen below
print(query_job.labels)

# try and get a job later, also no labels
job = client.get_job(query_job.job_id)
print(job.labels)

```

plamut:
Fixes #469.

This PR fixes reading the labels on Job instances and adds a system test that would have caught this bug.

PR checklist:

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -233,7 +233,7 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
     @property
     def labels(self):
         """Dict[str, str]: Labels for the job."""
-        return self._properties.setdefault("labels", {})
+        return self._properties.setdefault("configuration", {}).setdefault("labels", {})
 
     @property
     def etag(self):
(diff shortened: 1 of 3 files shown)


tswast (approved):
Thanks!

Can confirm in https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfiguration that this is how it's supposed to be (labels are a property of configuration)

PR #474, opened 2021-01-19
fix: error using empty array of structs parameter

### LINKED ISSUE #463: Support arrays of record/structs as query parameters

 Setting nested repeated fields via the python-bigquery API works (despite not being documented by [ArrayQueryParameter](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.query.ArrayQueryParameter.html#google.cloud.bigquery.query.ArrayQueryParameter)).
[ArrayQueryParamter.to_api_repr](https://googleapis.dev/python/bigquery/latest/_modules/google/cloud/bigquery/query.html#ArrayQueryParameter.to_api_repr) has code to handle "RECORD" or "STRUCT" but doesn't check if the array is empty and so throw exceptions.
  
I'd a way to define the schema for the array of records so that it doesn't matter that it's empty - I can still use the BQ python client. There appears to be solutions  for typed languages https://github.com/googleapis/google-cloud-java/issues/2678 as well as dynamic languages https://github.com/googleapis/google-cloud-ruby/pull/4269 and losely typed languages https://github.com/googleapis/nodejs-bigquery/issues/400.

Our current solution involves naming each individual field in the repeated struct  (eg fieldA_0, fieldB_0, fieldA_1,...)
and sending over MANY query parameters. As there's a limit to the number of named parameters the API allows, this is far from perfect and makes our code significantly harder to read and understand.

 **Additional context**
```python
client = bigquery.Client()

def query(params):
    sql = "SELECT * FROM UNNEST(@params) as param"
    query_parameter = bigquery.ArrayQueryParameter(
                "params",
                "RECORD",
                [
                    bigquery.StructQueryParameter.positional(
                        bigquery.ScalarQueryParameter("key", "STRING", key),
                        bigquery.ScalarQueryParameter("value", "STRING", value),
                    )
                    for key, value in (params.items())
                ],
    )
    for row in client.query(sql, job_config=bigquery.QueryJobConfig(query_parameters=[query_parameter])):
        print(dict(row))

query({"one": "a", "two": "b"}) # works
query({}) # IndexError: list index out of range
```

plamut:
Fixes #463.

This PR adds an optional type info parameter to `ArrayQueryParameter` so that array query parameters with STRUCT items can be used in queries even if they are empty. This is a solution similar to [the one used](https://github.com/googleapis/google-cloud-ruby/pull/4269#issuecomment-550562108) in e.g. the Ruby client.

(for non-struct items the type can always be determined by using the `array_type` argument)

PR checklist:


google/cloud/bigquery/query.py
```diff
@@ -186,31 +186,56 @@ class ArrayQueryParameter(_AbstractQueryParameter):
...
             Name of type of array elements.  One of `'STRING'`, `'INT64'`,
-            `'FLOAT64'`, `'NUMERIC'`, `'BOOL'`, `'TIMESTAMP'`, or `'DATE'`.
-
-        values (List[appropriate scalar type]): The parameter array values.
+            `'FLOAT64'`, `'NUMERIC'`, `'BOOL'`, `'TIMESTAMP'`, `'DATE'`,
+            or `'STRUCT'`/`'RECORD'`.
+
+        values (List[appropriate type]): The parameter array values.
+
+        struct_item_type (Optional[google.cloud.bigquery.query.StructQueryParameter]):
```
tswast, line 194: Could this be an iterable of [SchemaField](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.schema.SchemaField.html#google.cloud.bigquery.schema.SchemaField)? Since we only want type information, not actual values, it seems the query parameter classes are the wrong choice for this argument.

tswast (changes requested):

google/cloud/bigquery/query.py
plamut: Are those serialized into a compatible API representation and the backend won't notice? If yes, then let's do that.

I suppose tests should sufficiently protect us if these two representations drift apart at some point.

tswast: Good point. They aren't 100% the same. No "mode" in the query parameter types because it's following the standard SQL semantics. Instead of this implementation, perhaps we should introduce a [QueryParameterType](https://cloud.google.com/bigquery/docs/reference/rest/v2/QueryParameter#QueryParameterType) to match the API backend?

Then, `array_type` could take `Union[str, QueryParameterType]` instead of just `str`.

plamut: Both using a `*QueryParameter` instance with an empty value is or a `SchemaField` instance are a bit of a hack that might eventually bite use, thus introducing `QueryParameterType` sounds like the "most correct" solution, but it will require some extra work.

If that's fine, I can take a stab, WDYT?

google/cloud/bigquery/query.py
```diff
@@ -48,6 +48,226 @@ def __ne__(self, other):
...
+    def to_api_repr(self):
+        """Construct JSON API representation for the parameter type.
+
+        Returns:
+            Dict: JSON mapping
+        """
+        raise NotImplementedError
+
+
+class ScalarQueryParameterType(_AbstractQueryParameterType):
```
tswast, line 78: I don't think we should do this in this PR, but I'm tempted to ask for an additional set of constants in `google.cloud.bigquery.enums` that has all the scalar types defined as objects (without name & description). I imagine that will be useful when we eventually create an "array of structs query parameter" code sample.

google/cloud/bigquery/query.py
```diff
@@ -48,6 +48,226 @@ def __ne__(self, other):
...
+            else ""
+        )
+        return f"{self.__class__.__name__}({self._array_type!r}{name}{description})"
+
+
+class StructQueryParameterType(_AbstractQueryParameterType):
+    """Type representation for struct query parameters.
+
+    Args:
+        subtypes (Iterable[Union[ \
```
tswast, line 193: "fields" would be a more appropriate name for this.

https://cloud.google.com/bigquery/docs/parameterized-queries?hl=en#using_structs_in_parameterized_queries

google/cloud/bigquery/query.py
```diff
@@ -48,6 +48,226 @@ def __ne__(self, other):
...
+        name (Optional[str]):
+            The name of the query parameter. Primarily used if the type is
+            one of the subfields in ``StructQueryParameterType`` instance.
+        description (Optional[str]):
+            The query parameter description. Primarily used if the type is
+            one of the subfields in ``StructQueryParameterType`` instance.
+    """
+
+    def __init__(self, *subtypes, name=None, description=None):
+        self._subtypes = [type_ for type_ in subtypes]  # make a shallow copy
```
tswast, line 206: Looking at the unit tests, I think we might want a public "fields" property.

tswast (changes requested):

google/cloud/bigquery/query.py
```diff
@@ -263,22 +498,38 @@ def to_api_repr(self):
...
+            self.array_type, StructQueryParameterType
+        ):
             reprs = [value.to_api_repr() for value in values]
-            a_type = reprs[0]["parameterType"]
             a_values = [repr_["parameterValue"] for repr_ in reprs]
+
+            if reprs:
+                a_type = reprs[0]["parameterType"]
+            else:
+                # The constructor disallows STRUCT/RECORD type when empty values.
```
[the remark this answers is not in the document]
plamut: Admittedly, the comment is already much less obvious than when I wrote it, thus it's good to clarify the assertion's purpose.

google/cloud/bigquery/query.py
plamut: Sounds reasonable, let's do it in a follow-up PR. :+1:

google/cloud/bigquery/query.py
```diff
@@ -48,6 +48,226 @@ def __ne__(self, other):
...
+        type_ = resource["type"]
+        return cls(type_)
+
+    def to_api_repr(self):
+        """Construct JSON API representation for the parameter type.
+
+        Returns:
+            Dict: JSON mapping
+        """
+        return {"type": self._type}
```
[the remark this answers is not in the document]
plamut: Yes, that's the reason. A `QueryParameterType` in its [APi representation](https://cloud.google.com/bigquery/docs/reference/rest/v2/QueryParameter#QueryParameterType) can have an optional name and description attached to it, but only when used inside a `structTypes` list .

Using another wrapper felt too convoluted, and it's IMHO more natural to treat `QueryParameterType` as a single entity that can be used both as a top-level item, or as a struct field.

google/cloud/bigquery/query.py
plamut: Aha, let's then use the name used in the docs. :+1:

google/cloud/bigquery/enums.py
```diff
@@ -215,6 +216,26 @@ class SqlTypeNames(str, enum.Enum):
     DATETIME = "DATETIME"
 
 
+class SqlParameterScalarTypes:
```
tswast, line 219: Beautiful!

tswast (approved):

PR #480, opened 2021-01-21
fix: invalid conversion of timezone-aware datetime values to JSON

### LINKED ISSUE #476: BigQuery python cli fails to change time, BUT does changes timezone for TIMESTAMP which leads to an incorrect time
This code works with with the python package
   **google-cloud-bigquery==2.2.0** 
but it does not work with this package
   **google-cloud-bigquery==2.6.0**

What you expected to happen:

I expect the client library to take a datetime of
 '_2017-11-01 04:08 AM -07:00_'    and convert it to
 '_2017-11-01 11:08 AM -00:00_' in big query. but instead it uploads it to
 '_2017-11-01 04:08 AM -00:00_'

Steps to reproduce:
```
import datetime
from dateutil.tz import tzoffset
from google.cloud.bigquery import SchemaField, Table

schema = [SchemaField('c', 'TIMESTAMP', 'NULLABLE', None, (), None)] 
problem_date = datetime.datetime(2017, 11, 1, 4, 8, tzinfo=tzoffset(None, -25200))
table_name = ‘my_test_table’
dataset = <an existing dataset>

bigquery = bigquery.Client(credentials=<somecreds>, project=<ourproj>)

dataset_ref = bigquery.dataset(dataset, project=<ourproj>)
table_ref = dataset_ref.table(table_name)
table = bigquery.create_table(Table(table_ref, schema=schema))
errors = bigquery.insert_rows(table, [[problem_date]])
```

#### Other info
-------------------------------
I worked with the google BigQuery team and here is their [feedback](https://issuetracker.google.com/issues/175849093).
Hello,

I have been investigating the issue you described and it seems like this change of behaviour of the python client library was introduced in the version 2.5.0 [1] with the pull request #393 [2], whose objective was to fix a previous issue with floating points and timestamps in the method `insert_rows`.

I have been able to reproduce your issue in new versions after 2.5.0 while it does not happen in the previous ones. The suggested way to proceed here would be to open an issue in the Github repository [3] so you can transmit your findings to the Engineering Team, since the code is open source .

Alternatively, you might want to add the following line of code, so the timezone is converted to UTC before trying to insert the row into BigQuery. Please let me know if this workaround worked for you:
```
problem_date = datetime.datetime(2017, 11, 1, 4, 8, tzinfo=tzoffset(None, -25200))
problem_date = problem_date.astimezone(datetime.timezone.utc)
```

[1]:
https://github.com/googleapis/python-bigquery/releases/tag/v2.5.0
[2]:
https://github.com/googleapis/python-bigquery/pull/393
[3]:
https://github.com/googleapis/python-bigquery/issues

-------------------------------

We will look into the feasibility of the workaround

plamut:
Fixes #476.

This PR fixes converting timezone-aware datetime values to an UTC JSON string.

PR checklist:

### DIFF
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -315,6 +315,10 @@ def _timestamp_to_json_parameter(value):
 def _timestamp_to_json_row(value):
     """Coerce 'value' to an JSON-compatible representation."""
     if isinstance(value, datetime.datetime):
+        # For naive datetime objects UTC timezone is assumed, thus we format
+        # those to string directly without conversion.
+        if value.tzinfo is not None:
+            value = value.astimezone(UTC)
         value = value.strftime(_RFC3339_MICROS)
     return value
 
@@ -322,6 +326,10 @@ def _timestamp_to_json_row(value):
 def _datetime_to_json(value):
     """Coerce 'value' to an JSON-compatible representation."""
     if isinstance(value, datetime.datetime):
+        # For naive datetime objects UTC timezone is assumed, thus we format
+        # those to string directly without conversion.
+        if value.tzinfo is not None:
+            value = value.astimezone(UTC)
         value = value.strftime(_RFC3339_MICROS_NO_ZULU)
     return value
 
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,6 @@ dependencies = [
     "google-cloud-core >= 1.4.1, < 2.0dev",
     "google-resumable-media >= 0.6.0, < 2.0dev",
     "protobuf >= 3.12.0",
-    "python-dateutil >= 2.8.1",  # used in tests, otherwise dependency of "pandas" extra
 ]
 extras = {
     "bqstorage": [
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ dependencies = [
     "google-cloud-core >= 1.4.1, < 2.0dev",
     "google-resumable-media >= 0.6.0, < 2.0dev",
     "protobuf >= 3.12.0",
+    "python-dateutil >= 2.8.1",  # used in tests, otherwise dependency of "pandas" extra
 ]
 extras = {
     "bqstorage": [
--- a/tests/unit/test__helpers.py
+++ b/tests/unit/test__helpers.py
@@ -726,7 +726,7 @@ class Test_timestamp_to_json_row(unittest.TestCase):
         ZULU = "2016-12-20 15:58:27.339328+00:00"
         self.assertEqual(self._call_fut(ZULU), ZULU)
 
-    def test_w_datetime(self):
+    def test_w_datetime_no_zone(self):
         when = datetime.datetime(2016, 12, 20, 15, 58, 27, 339328)
         self.assertEqual(self._call_fut(when), "2016-12-20T15:58:27.339328Z")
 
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -34,6 +34,7 @@
     "google-cloud-core >= 1.4.1, < 2.0dev",
     "google-resumable-media >= 0.6.0, < 2.0dev",
     "protobuf >= 3.12.0",
+    "python-dateutil >= 2.8.1",  # used in tests, otherwise dependency of "pandas" extra
```
tswast, line 37: When adding dependencies, could you also add to https://github.com/googleapis/python-bigquery/blob/master/testing/constraints-3.6.txt so that we know the minimum version we specify does actually work?

google/cloud/bigquery/_helpers.py
```diff
@@ -315,13 +315,21 @@ def _timestamp_to_json_parameter(value):
 def _timestamp_to_json_row(value):
     """Coerce 'value' to an JSON-compatible representation."""
     if isinstance(value, datetime.datetime):
+        # For naive datetime objects UTC timezone is assumed, thus we format
```
tswast, line 318: Was this always our assumption? Based on the unit tests, I suppose so?

setup.py
tswast: Also, 2.8.1 is the latest version of python-dateutil. Are you sure we can't use a wider range?

tswast: Also also, do we actually need this dependency? I see it's commonly used and BSD-licensed, so should be OK, but I like to avoid new deps where possible.

If truly test-only, we can put this in the noxfile instead.

tswast (changes requested):

google/cloud/bigquery/_helpers.py
plamut: Can't say if "always", but the `_datetime_to_json()` helper has been unchanged since 2017 and was supposedly doing its job fine with naive datetime objects in this time - at least I don't remember any reported issues related to timezones. The `strftime()` conversion just uses the `datetime` fields at their face value and ignores any `tzinfo` set.

Considering that naive datetimes are used most of the time (as opposed to timezone-aware datetime objects) and that users did not complain, the UTC timezone assumption was apparently correct behavior. The logic just failed to cover the case when somebody actually used a timezone-aware instance.

And indeed, unit tests expect the same, a naive `datetime` instance is supposed to be converted to a string representation with a zero offset from UTC ([example](https://github.com/googleapis/python-bigquery/blob/530e1e8d8fe8939e914a78ff1b220907c1b87af7/tests/unit/test__helpers.py#L729-L731)).

setup.py
plamut: I used it because the issue reporter's example used its `tzoffset()` helper to specify a timezone offset. But on a second though you are right, adding a dependency just for that is an overkill, even if we currently do implicitly install it as a sub-dependency.

I'll just manually define a custom `datetime.tzinfo` subclass like in some other tests, and we don't even have to bother much about the logic of its `dst()` method, as dateutil's [tzoffset](https://github.com/dateutil/dateutil/blob/24caabfb63c7dcbf6499b3ee5b6140e6b7da0d8a/dateutil/tz/tz.py#L132-L158) simply returns zero.

tswast (approved):

PR #482, opened 2021-01-22
fix: use explicitly given project over the client's default project for load jobs

### LINKED ISSUE #473: Creating a load job in a project other than the client's project results in error
The following code
```
    bq_client = bigquery.Client("project_A")
    job = bq_client.load_table_from_json(data, table, job_config=job_config, project="project_B")
```
would give the error 400 "Value project_B in content does not agree with value project_A. This can happen when a value set through a parameter is inconsistent with a value set in the request."

It seems the upload URL is always set to the client's project.
https://github.com/googleapis/python-bigquery/blob/80944f080bcc4fda870a6daf1d884de616d39ae7/google/cloud/bigquery/client.py#L2594

plamut:
Fixes #473.

This PR fixes the ignored `project` parameter for load jobs (the client's default project was incorrectly always used).

PR checklist:

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2136,11 +2136,11 @@ class Client(ClientWithProject):
         try:
             if size is None or size >= _MAX_MULTIPART_SIZE:
                 response = self._do_resumable_upload(
-                    file_obj, job_resource, num_retries, timeout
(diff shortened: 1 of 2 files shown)


tswast (approved):
Good catch, thanks!

PR #484, opened 2021-01-24
fix: do not remove `destinationTable` from job config (#483)

### LINKED ISSUE #483: Query jobs destinationTable attribute is explicitly removed
# What happened:

I run a job with a query and a destination table and the table is not generated from the file.
```python
bq_client.create_job({
    'query': {
        'query': rendered_query,
        'destinationTable': {
            'projectId': GCP_PROJECT,
            'datasetId': DATASET_ID,
            'tableId': TRANSFORMED_TABLE_ID
        },
        'writeDisposition': 'WRITE_TRUNCATE'
})
```

My code executes successfully and produces the following job in the console
image

# How to fix it
There is a line where the `destinationTable` is removed from the job config. Removing the line that deletes the property in the job configuration will do.
https://github.com/googleapis/python-bigquery/blob/530e1e8d8fe8939e914a78ff1b220907c1b87af7/google/cloud/bigquery/client.py#L1729

Can anyone explain why is the `destinationTable` property removed?

In the meanwhile I'm opening a PR

dinigo:

Fixes #483  🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -47,7 +47,6 @@ import google.cloud._helpers
 from google.cloud import exceptions
 from google.cloud.client import ClientWithProject
 
-from google.cloud.bigquery._helpers import _del_sub_prop
 from google.cloud.bigquery._helpers import _get_sub_prop
 from google.cloud.bigquery._helpers import _record_field_to_json
 from google.cloud.bigquery._helpers import _str_or_none
@@ -1726,7 +1725,6 @@ class Client(ClientWithProject):
             )
         elif "query" in job_config:
             copy_config = copy.deepcopy(job_config)
-            _del_sub_prop(copy_config, ["query", "destinationTable"])
             query_job_config = google.cloud.bigquery.job.QueryJobConfig.from_api_repr(
                 copy_config
             )


plamut:
@dinigo Thanks for the PR.

The changes need to be covered with the corresponding test(s), but it's not entirely clear what is the expected behavior here (as discussed on the ticket). Let's wait until that gets clarified.

tswast:
As discussed in https://github.com/googleapis/python-bigquery/issues/483 the `create_job` function was initially added so that customers could retry jobs. Since the destination table isn't writable when it is an anonymous temporary table, we delete this property.

I agree that this was too aggressive, and to be a general `create_job` function we should only strip `destinationTable` when the temporary table `datasetId` begins with `_` and the `tableId` begins with `anon`.

As @plamut unit tests need to be updated to cover this change in logic.

tswast:
Now that we're considering a 3.0.0 release, I'd like us to revisit this request.

I believe to implement https://github.com/googleapis/python-bigquery/issues/539 @jimfulton will be adding a property to `XYZJob` classes with a copy of the original request that created the job. This could be used instead of trying to add/remove properties from the job resource that we get from the response.

Closing in favor of a future PR, tied to the 3.0.0 release.

PR #487, opened 2021-01-27
Install the latest version of pyarrow (3.0.0)

macbre:
https://pypi.org/project/pyarrow/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ extras = {
         # grpc.Channel.close() method isn't added until 1.32.0.
         # https://github.com/grpc/grpc/pull/15254
         "grpcio >= 1.32.0, < 2.0dev",
-        "pyarrow >= 1.0.0, < 3.0dev",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@
         # grpc.Channel.close() method isn't added until 1.32.0.
         # https://github.com/grpc/grpc/pull/15254
         "grpcio >= 1.32.0, < 2.0dev",
-        "pyarrow >= 1.0.0, < 3.0dev",
+        "pyarrow >= 1.0.0, < 4.0dev",
```
plamut, line 48: We actually need `pyarrow==3.0.0` to support Python 3.9, meaning that the lower bound should be bumped, too.

plamut (commented):
Thanks for the PR, but in parallel I made a similar change in #488 along with a few other changes to support Python 3.9 - let's add the pyarrow version bump in that PR then.

plamut:
Closing as a subset of #488, but thanks nevertheless, @macbre

Issue #462, opened 2021-01-07
Run checks to use latest supported Python (3.9)


plamut:
In `setup.py` and in README we declare that we support `Python >= 3.6`, but the nox test sessions only use Python up to 3.8. Python 3.9 is also missing from the `classifiers` list in `setup.py`.

We should bump the maximum versions to `3.9` where applicable. If tests fail in 3.9, we should bound the `python_requires` version in `setup.py` until the failures are fixed (though - bound it in any case and manually increase only after we confirm that it works with the new minor Python release?).

If we first need to setup Kokoro environment to support Python 3.9 test environment, let's do that and keep this ticket open as a reminder.

cc: @tswast @busunkim96

tswast:
This was blocked by grpcio and arrow have been released for both platforms now 
(https://github.com/grpc/grpc/issues/24344#issuecomment-725303543, https://issues.apache.org/jira/browse/ARROW-10224)

plamut:
As a quick check, I tried to install the library with Python 3.9, but failed, because `pyarrow` pulled in `numpy==1.16.0`, but the latter failed to compile on my machine (Ubuntu), and there is no wheel for it. It didn't help if I installed `numpy==1.19.5`, pyarrow was still stubbornly pulling in `1.16.0`.

TL; DR - I was not able to install BigQuery localy with Python3.9, and thus could not run the tests either.

tswast:
I misunderstood the ticket. They have released 3.9 wheels to nightly but not public PyPI yet. See my thread talking to myself in: https://github.com/googleapis/python-bigquery-storage/pull/99

tswast:
Arrow 3.0 release with Python 3.9 support is in the works. https://github.com/apache/arrow/pull/9245

kszucs:
The wheels are available now on pypi.

plamut:
@kszucs That's great, thanks a lot!

galuszkak:
@plamut I think this ticket can be closed.

PR #447, opened 2020-12-22
feat: add BIGNUMERIC support to pandas helpers and DB-API

HemangChothani:
Fixes #367

PR #488, opened 2021-01-27
deps: declare support for Python 3.9

plamut:
Towards #462.

With wheels for `pyarrow==3.0.0` released we can now officially support Python 3.9 - well, at least after when we add Python 3.9 checks to Kokoro.

PR checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,12 +45,11 @@ extras = {
         # grpc.Channel.close() method isn't added until 1.32.0.
         # https://github.com/grpc/grpc/pull/15254
         "grpcio >= 1.32.0, < 2.0dev",
-        "pyarrow >= 1.0.0, < 3.0dev",
+        "pyarrow >= 3.0.0, < 4.0dev",
     ],
     "pandas": [
         "pandas>=0.23.0",
-        # pyarrow 1.0.0 is required for the use of timestamp_as_object keyword.
-        "pyarrow >= 1.0.0, < 3.0dev",
+        "pyarrow >= 3.0.0, < 4.0dev",
     ],
     "tqdm": ["tqdm >= 4.7.4, <5.0.0dev"],
     "opentelemetry": [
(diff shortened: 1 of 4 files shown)


plamut:
Oh, we need to bump the version in `samples/`, too.

plamut:
FWIW, I successfully ran unit and system test with Python 3.9 locally (Ubuntu).

setup.py
```diff
@@ -45,12 +45,11 @@
...
         # https://github.com/grpc/grpc/pull/15254
         "grpcio >= 1.32.0, < 2.0dev",
-        "pyarrow >= 1.0.0, < 3.0dev",
+        "pyarrow >= 3.0.0, < 4.0dev",
     ],
     "pandas": [
         "pandas>=0.23.0",
-        # pyarrow 1.0.0 is required for the use of timestamp_as_object keyword.
-        "pyarrow >= 1.0.0, < 3.0dev",
+        "pyarrow >= 3.0.0, < 4.0dev",
```
tswast, line 52: Ditto.

setup.py
```diff
@@ -45,12 +45,11 @@
         # grpc.Channel.close() method isn't added until 1.32.0.
         # https://github.com/grpc/grpc/pull/15254
         "grpcio >= 1.32.0, < 2.0dev",
-        "pyarrow >= 1.0.0, < 3.0dev",
+        "pyarrow >= 3.0.0, < 4.0dev",
```
tswast, line 48: Can we keep this at 1.0 as a minimum? Or were there breaking changes we needed to account for?

Since pyarrow is a core library, the wider we can keep these dependencies, the fewer conflicts we'll get.

tswast (approved):
LGTM, assuming Python 3.6 tests pass with `pyarrow==1.0.0`

tswast:
I pushed a commit that adds some variables for python versions to the noxfile. This is to align more closely with the central "synthtool" template.

Also, I remembered that we'll probably need `pyarrow>=3.0` for https://github.com/googleapis/python-bigquery/pull/447, but I'd still like to avoid that requirement if at all possible, since BIGNUMERIC is less commonly used than is pyarrow 1&2 right now.

PR #522, opened 2021-02-15
feat: add determinism level for javascript UDFs

### LINKED ISSUE #503: BigQuery feature: support determinism level for javascript UDFs
# Summary

Ensure BigQuery client library supports the ability for a javascript UDF (exposed via the Routine resource) to communicate **determinism level**.

The functionality here is to allow users to inform the query engine about whether a function is deterministic or not, which can improve query performance.   Existing clients should already support the routines methods and resource type, so this should largely be a case of exposing a new field of the Routine and updating existing testing.

For integration testing, please note that only javascript UDFs allow determinism to be set; the backend will likely error when you try to set determinism for SQL UDFs or stored procedures.

# References

Public REST reference includes information about the routine resource structure:

https://cloud.google.com/bigquery/docs/reference/rest/v2/routines#resource:-routine

Here's an example implementation that adds a new field to the Routine resource type in Go:

https://github.com/googleapis/google-cloud-go/pull/3534

plamut:
Closes #503.

This PR adds support for the determinism level setting for user-defined routines.

PR checklist:


tests/system/test_client.py
```diff
@@ -2682,6 +2682,7 @@ def test_create_routine(self):
             )
         ]
         routine.body = "return maxValue(arr)"
+        routine.determinism_level = "DETERMINISTIC"
```
plamut, line 2685: Should determinism level be an enum?

I can see some dedicated string "enum" classes, e.g. `WriteDisposition` , while some other enum values (as defined by API) are set by just using plain strings, e.g. `RoutineArgument.mode`.

tswast: Probably **should**. We usually get a few pieces of sample feedback in our cloud.google.com docs if we use raw strings anywhere instead of enum-like constants. It is more to maintain, though.

plamut: Added an enum-like class similar to e.g. `WriteDisposition` to use instead of raw strings. Used such a class instead of a "real" enum both for consistency, and for keeping the logic unchanged.

tswast (approved):
Thanks!

PR #524, opened 2021-02-16
feat: expose reservation usage stats on jobs

### LINKED ISSUE #507: BigQuery Feature: reservation usage in job statistics
# Summary

This FR tracks work to expose reservation usage information that may be reported as part of BigQuery job statistics, namely [`JobStatistics.reservationUsage`](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobStatistics)

# Details

More detailed information about this work has previously been shared out of band with library maintainers.

plamut:
Closes #507.

This PR exposes reservation usage stats on job objects.

PR checklist:

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -14,6 +14,7 @@
 
 """Base classes and helpers for job classes."""
 
+from collections import namedtuple
 import copy
 import http
 import threading
@@ -71,6 +72,16 @@ def _error_result_to_exception(error_result):
     return exceptions.from_http_status(
         status_code, error_result.get("message", ""), errors=[error_result]
     )
+
+
+ReservationUsage = namedtuple("ReservationUsage", "name slot_ms")
+ReservationUsage.__doc__ = " Job resource usage for a reservation."
+ReservationUsage.name.__doc__ = (
+    'Reservation name or "unreserved" for on-demand resources usage.'
+)
+ReservationUsage.slot_ms.__doc__ = (
+    "Total slot milliseconds used by the reservation for a particular job."
+)
 
 
 class _JobReference(object):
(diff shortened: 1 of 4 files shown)


tswast (commented):
We need to add to https://github.com/googleapis/python-bigquery/blob/master/docs/reference.rst under "job-related types" too.

I know in most libraries we use automodule, but it probably makes sense to continue to have a central index for BigQuery, since we have a lot more modules than most.

google/cloud/bigquery/job/base.py
```diff
@@ -73,6 +74,16 @@ def _error_result_to_exception(error_result):
     )
 
 
+ReservationUsage = namedtuple("ReservationUsage", "name slot_ms")
```
tswast, line 77: Interesting!

This is a little different from our usual "class with `_properties` dictionary" approach, but does seem to save us some boilerplate. I'll allow it. :-)

Do you think we should do this for our other read-only classes?

tswast (approved):

google/cloud/bigquery/job/base.py
```diff
@@ -73,6 +74,16 @@ def _error_result_to_exception(error_result):
...
+ReservationUsage.__doc__ = " Job resource usage for a reservation."
```
plamut, line 78:
```suggestion
ReservationUsage.__doc__ = "Job resource usage for a reservation."
```

google/cloud/bigquery/job/base.py
plamut: I though a namedtuple would better communicate that it's a read-only (output-only) property, although our "_properties classes"  can also have read-only properties - but yeah, a namedtuple requires much less boilerplate.

(FWIW, I personally view these output-only classes merely as named read-only containers with descriptive attribute names, and using a class with a bunch of read-only properties is just an implementation detail)

We can eventually refactor other output-only classes to namedtuples for consistency, yes, but I wouldn't do it just yet. There's not that much immediate value in it, as that code has already been written, and there might be some legit use cases out there that we don't know about... let's first wait and see how a namedtuple performs when released into the wild.

PR #527, opened 2021-02-18
feat: add BIGNUMERIC support

### LINKED ISSUE #367: BIGNUMERIC support
BIGNUMERIC is a new type with even larger decimal precision than NUMERIC. We should ensure we can go to/from `decimal.decimal` and this new data type.

plamut:
Closes #367.

This is a continuation of #447 that adds extra tests and a few final touches.

PR checklist:


plamut:
@HemangChothani If you see this, can you please confirm to the bot that you consent using your commits? (this PR builds on top of your work)

Thanks!

plamut:
Tests will have to be refactored to not assume `pyarrow>=3.0.0` is available (they fail when they are run with `pyarrow==1.0.0`)  - the BIGNUMERIC fixtures and assertions need to be moved out and skipped conditionally.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -129,9 +133,13 @@ def pyarrow_timestamp():
         pyarrow.binary().id: "BYTES",
         pyarrow.string().id: "STRING",  # also alias for pyarrow.utf8()
         pyarrow.decimal128(38, scale=9).id: "NUMERIC",
+    }
+
+    if int(pyarrow.__version__.split(".")[0]) >= 3:
```
tswast, line 138: This should work, but I'd prefer if we use a more explicit version comparison.

For example: https://github.com/googleapis/python-api-core/blob/7271b23afddb032e49e957525704d0cd5bfa4c65/google/api_core/grpc_helpers.py#L47

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -129,9 +133,13 @@ def pyarrow_timestamp():
...
+        BQ_TO_ARROW_SCALARS["BIGNUMERIC"] = pyarrow_bignumeric
         # The exact decimal's scale and precision are not important, as only
         # the type ID matters, and it's the same for all decimal128 instances.
```
tswast, line 141: Comment seems off, as this is a decimal256

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -184,7 +189,11 @@ def bigquery_scalar_type(value):
     elif isinstance(value, numbers.Real):
         return "FLOAT64"
     elif isinstance(value, decimal.Decimal):
-        return "NUMERIC"
+        scalar_object = pyarrow.scalar(value)
+        if isinstance(scalar_object, pyarrow.Decimal128Scalar):
```
tswast, line 193: Let's add a comment that we check for numeric before bignumeric in order to support pyarrow < 3.0

tswast (commented):
Looking good! One suggestion regarding the version comparison. We also will need some `skipif`s. Might be able to do those based on the pattern we have in `google-auth`, with a private constant that indicates if the BIGNUMERIC feature is available.

google/cloud/bigquery/_pandas_helpers.py
tswast: Also, we can add a constant as is done in `google-api-core` to indicate to tests and other logic if the BIGNUMERIC feature is available.

plamut: Makes sense.

Re: version comparison - I initially wanted to avoid using `pkg_resources.parse_version()`, because it depends on setuptools, but on a second thought we already implicitly require `setuptools` to install the package, thus relying on `pkg_resources` being present is fine.

google/cloud/bigquery/_pandas_helpers.py
plamut: Copy-pasta, will fix it. Thanks for catching it!

setup.py
```diff
@@ -48,6 +49,7 @@
         "pyarrow >= 1.0.0, < 4.0dev",
     ],
     "pandas": ["pandas>=0.23.0", "pyarrow >= 1.0.0, < 4.0dev",],
+    "bignumeric_type": ["pyarrow >= 3.0.0, < 4.0dev"],
```
tswast, line 52: I like that we're documenting the supported versions in a machine-checkable way, but I'm also wary of adding extras. Would we consider it a breaking change to remove this in the future when we bump every minimum pyarrow to 3.0?

tswast (approved):
One question, but otherwise looks good

setup.py
plamut: Hmm... I would say not. The existing code itself will still run fine and will not break, perhaps only _installing_ a new library version might break some deployment pipelines when the `bignumeric_type` extra is removed.

The following are the broad categories of BigQuery installations:

- BQ installed without any extra - these should be fine (even if they use BIGNUMERIC, the latter implies that they already have an up-to-date `pyarrow` version that supports it). And fresh installations get the latest supported dependency versions anyway.
- BQ installed with all extras by using the catch-all `"all"` extra - these should be fine, as they already use `pyarrow` 3+ pulled in by `bignumeric_type`. Removing that extra will not downgrade `pyarrow`.
- BQ specifically installed with the `bignumeric_type` extra - installation will emit a warning about the missing extra (at least when using `pip`). This _might_ break a pipeline if it's made sensitive to any warnings, but it's probably not that common a case.

Do you think that they are any other cases that I might have missed?

tswast:
"CLAs are signed, but unable to verify author consent"

Since these commits were pulled from https://github.com/googleapis/python-bigquery/pull/447, I'm going to call that author consent and mark the label as true.

setup.py
tswast: I think you're right that most people will get this via `all`

PR #544, opened 2021-03-09
feat: make QueryJob.done() method more performant

### LINKED ISSUE #534: Improve performance of `QueryJob.done()` method

As reported in internal issue 180972119, with a name like `done()`, I would expect it to be a very lightweight call. At most, I'd expect it to refresh the job state. It is **not** a lightweight call. It calls the `jobs.getQueryResults` REST API, which can hang for 10 seconds or more.

`QueryJob.done()` should return `True` if the job has finished or reload the job and return the correct answer.

This requires moving the repeated call to `jobs.getQueryResults` to somewhere else in the stack of functions between `QueryJob.result()` and `QueryJob.done()`.

Perhaps a flag to "go fast" to `done()`? This feels like it's making the developer interface unnecessarily complicated, though.

 **Additional context**

Related to customer support case 26956233

plamut:
Fixes #534.

This PR removes refreshing query results from the `QueryJob.done()` method, the latter is now just the `done()` method inherited from the `_AsyncJob` base class that at most reloads the job itself and checks if its state is DONE.

Since blocking poll from the `PollingFuture` base class repeatedly invokes `done()`, the change would cause too many job reload requests while waiting for the query results. The `QueryJob` class thus overrides the `_done_or_raise()` method repeatedly used by the blocking poll so that the polling is actually performed by fetching the query results. The latter call can block for up to 10 seconds, meaning that fewer polling requests are made than if reload the job was used.

#### How to test
Set logging level to DEBUG to see what HTTP requests are made. Then run a query job that normally takes more than 10 seconds to complete, for example:
```sql
SELECT
CONCAT(
    'https://stackoverflow.com/questions/',
    CAST(id as STRING)) as url,
view_count, 1 AS foo
FROM `bigquery-public-data.stackoverflow.posts_questions`"""
"""
ORDER BY view_count DESC
```
> **Tip:** If running the query in multiple test runs in a row, change `1 AS foo` to a different value so that the query is re-run and cached query results are **not** used)

While the query is running, call `query_job.result()` - query results should be fetched, but with a reasonable amount of requests.

Beside testing the `.result()` method, the `.done()` method should be checked, too - if it is run repeatedly while the query is running, each call should finish "fast", i.e. it should block for significantly less time than 10 seconds, because all it needs to do is to reload the job data itself.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -19,6 +19,7 @@ import copy
 import re
 
 from google.api_core import exceptions
+from google.api_core.future import polling as polling_future
 import requests
 
 from google.cloud.bigquery.dataset import Dataset
@@ -42,7 +43,6 @@ from google.cloud.bigquery.table import TimePartitioning
 from google.cloud.bigquery._tqdm_helpers import wait_for_query
 
 from google.cloud.bigquery.job.base import _AsyncJob
-from google.cloud.bigquery.job.base import _DONE_STATE
 from google.cloud.bigquery.job.base import _JobConfig
 from google.cloud.bigquery.job.base import _JobReference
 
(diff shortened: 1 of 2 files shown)


tswast (approved):
Magnificent!

PR #448, opened 2020-12-22
refactor: split system tests across modules

tswast:
This should make it easier to see which pandas tests
we have and which we are missing.

Despite the branch name, not exactly related to https://github.com/googleapis/python-bigquery/issues/56, but I realized I was having trouble determining if this pandas DATETIME use case was tested when all system tests were in the same file.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Towards #366

Issue #366, opened 2020-11-04
process: use templated noxfile + refactor tests


tswast:
It'd be nice to stay consistent with the other libraries in how we perform tests.

Blocked by:

- https://github.com/googleapis/python-bigquery/pull/203 - move samples to templated subdir(s)
- https://github.com/googleapis/synthtool/issues/838 -- extras in template
- https://github.com/googleapis/synthtool/issues/839 -- constraints in template

tswast:
This issue may be considered to be blocking: https://github.com/googleapis/python-bigquery/pull/336 as the SNIPPETS session is a custom session, which is not present in other libraries.

chalmerlowe:
Closing. A number of tests and such have been relocated to more appropriate locations.

PR #548, opened 2021-03-11
refactor: split pandas system tests to new module

tswast:

Follow-up to https://github.com/googleapis/python-bigquery/pull/448

Towards #366

### DIFF
--- a/tests/system/test_pandas.py
+++ b/tests/system/test_pandas.py
@@ -1,0 +1,794 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


tests/system/test_pandas.py
```diff
@@ -0,0 +1,794 @@
...
+    row_tuples = [r.values() for r in sorted_rows]
+    expected = [
+        tuple(None if col != col else col for col in data_row)
+        for data_row in dataframe.itertuples(index=False)
+    ]
+
+    assert len(row_tuples) == len(expected)
+
+    for row, expected_row in zip(row_tuples, expected):
+        assert len(row) == len(expected_row)  # column order does not matter
```
plamut, line 697: This assertion is weaker than the original - [assertCountEqual()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual) not only checks that the lengths of two sequences match, but also that the counts of their individual elements match.

If the elements are hashable, the easiest way to make an equivalent assertion is probably using the `collections.Counter` class:
```py
assert Counter(row) == Counter(expected_row)
```

plamut (changes requested):
I like how the system tests are getting some structure. :+1: I think it's good, but I did spot an incorrectly converted assertion.

BTW, in the future, I suggest keeping related changes in their own commits, as it makes it easier to glance over the diff and directly see the key differences. That assertion conversion could have been easily overlooked.

tests/system/test_pandas.py
tswast: Nice catch!

plamut (approved):

PR #551, opened 2021-03-15
fix: DB API depends on pyarrow when decimal query parameters are used

### LINKED ISSUE #549: DB-API depends on pyarrow when decimal query parameters are used
It seems we introduced a hard dependency on pyarrow to detect data types of query parameters in the DB-API module. While `pyarrow` is recommended as it's needed to use the BigQuery Storage API, I don't think we want to depend on it for basic functionality.

I believe we added this hard dependency here: https://github.com/googleapis/python-bigquery/pull/527/files#diff-2221cd9b6b85f2ed1f943c474e8a8628e6ca28ec9a96c0aa5873cd6a66826bdeR193

We use `pyarrow` to disambiguate between `NUMERIC` and `BIGNUMERIC`. I'm not seeing a good way to check the precision / scale in the [decimal](https://docs.python.org/3/library/decimal.html) type, so in the case that `pyarrow` is not available, we should guess the type is `NUMERIC`.

#### Stack trace

```
__________________________________________________________________________________ test_dml ___________________________________________________________________________________
test/test_sqlalchemy_bigquery.py:430: in test_dml
    engine.execute(table_dml.insert(ONE_ROW_CONTENTS_DML))
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/base.py:2235: in execute
    return connection.execute(statement, *multiparams, **params)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/base.py:1011: in execute
    return meth(self, multiparams, params)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/sql/elements.py:298: in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/base.py:1124: in _execute_clauseelement
    ret = self._execute_context(
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/base.py:1316: in _execute_context
    self._handle_dbapi_exception(
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/base.py:1514: in _handle_dbapi_exception
    util.raise_(exc_info[1], with_traceback=exc_info[2])
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/util/compat.py:182: in raise_
    raise exception
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/base.py:1276: in _execute_context
    self.dialect.do_execute(
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/sqlalchemy/engine/default.py:608: in do_execute
    cursor.execute(statement, parameters)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/google/cloud/bigquery/dbapi/_helpers.py:263: in with_closed_check
    return method(self, *args, **kwargs)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/google/cloud/bigquery/dbapi/cursor.py:165: in execute
    query_parameters = _helpers.to_query_parameters(parameters)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/google/cloud/bigquery/dbapi/_helpers.py:168: in to_query_parameters
    return to_query_parameters_dict(parameters)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/google/cloud/bigquery/dbapi/_helpers.py:147: in to_query_parameters_dict
    param = scalar_to_query_parameter(value, name=name)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/google/cloud/bigquery/dbapi/_helpers.py:51: in scalar_to_query_parameter
    parameter_type = bigquery_scalar_type(value)
../../miniconda3/envs/pybigquery/lib/python3.9/site-packages/google/cloud/bigquery/dbapi/_helpers.py:193: in bigquery_scalar_type
    scalar_object = pyarrow.scalar(value)
E   AttributeError: 'NoneType' object has no attribute 'scalar'
```

plamut:
Fixes #549.

This PR removes the hard dependency of DB API on the optional `pyarrow` dependency.

**PR checklist**:

### DIFF
--- a/google/cloud/bigquery/dbapi/_helpers.py
+++ b/google/cloud/bigquery/dbapi/_helpers.py
@@ -189,6 +189,10 @@ def bigquery_scalar_type(value):
     elif isinstance(value, numbers.Real):
         return "FLOAT64"
     elif isinstance(value, decimal.Decimal):
+        if pyarrow is None:
+            # We don't try to manually guess if it's maybe a BIGNUMERIC.
+            return "NUMERIC"
+
         # We check for NUMERIC before BIGNUMERIC in order to support pyarrow < 3.0.
         scalar_object = pyarrow.scalar(value)
         if isinstance(scalar_object, pyarrow.Decimal128Scalar):
--- a/google/cloud/bigquery/dbapi/_helpers.py
+++ b/google/cloud/bigquery/dbapi/_helpers.py
@@ -19,14 +19,13 @@ import decimal
 import functools
 import numbers
 
-try:
-    import pyarrow
-except ImportError:  # pragma: NO COVER
-    pyarrow = None
-
 from google.cloud import bigquery
 from google.cloud.bigquery import table
 from google.cloud.bigquery.dbapi import exceptions
+
+
+_NUMERIC_SERVER_MIN = decimal.Decimal("-9.9999999999999999999999999999999999999E+28")
+_NUMERIC_SERVER_MAX = decimal.Decimal("9.9999999999999999999999999999999999999E+28")
 
 
 def scalar_to_query_parameter(value, name=None):
@@ -189,16 +188,20 @@ def bigquery_scalar_type(value):
     elif isinstance(value, numbers.Real):
         return "FLOAT64"
     elif isinstance(value, decimal.Decimal):
-        if pyarrow is None:
-            # We don't try to manually guess if it's maybe a BIGNUMERIC.
-            return "NUMERIC"
-
-        # We check for NUMERIC before BIGNUMERIC in order to support pyarrow < 3.0.
-        scalar_object = pyarrow.scalar(value)
-        if isinstance(scalar_object, pyarrow.Decimal128Scalar):
+        vtuple = value.as_tuple()
+        # NUMERIC values have precision of 38 (number of digits) and scale of 9 (number
+        # of fractional digits), and their max absolute value must be strictly smaller
+        # than 1.0E+29.
+        # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
+        if (
+            len(vtuple.digits) <= 38  # max precision: 38
+            and vtuple.exponent >= -9  # max scale: 9
+            and _NUMERIC_SERVER_MIN <= value <= _NUMERIC_SERVER_MAX
+        ):
             return "NUMERIC"
         else:
             return "BIGNUMERIC"
+
     elif isinstance(value, str):
         return "STRING"
     elif isinstance(value, bytes):
--- a/tests/unit/test_dbapi__helpers.py
+++ b/tests/unit/test_dbapi__helpers.py
@@ -17,7 +17,6 @@ import decimal
 import math
 import operator as op
 import unittest
-from unittest import mock
 
 try:
     import pyarrow
@@ -26,7 +25,6 @@ except ImportError:  # pragma: NO COVER
 
 import google.cloud._helpers
 from google.cloud.bigquery import table
-from google.cloud.bigquery._pandas_helpers import _BIGNUMERIC_SUPPORT
 from google.cloud.bigquery.dbapi import _helpers
 from google.cloud.bigquery.dbapi import exceptions
 from tests.unit.helpers import _to_pyarrow
@@ -40,7 +38,6 @@ class TestQueryParameters(unittest.TestCase):
             (123, "INT64"),
             (-123456789, "INT64"),
             (1.25, "FLOAT64"),
-            (decimal.Decimal("1.25"), "NUMERIC"),
             (b"I am some bytes", "BYTES"),
             ("I am a string", "STRING"),
             (datetime.date(2017, 4, 1), "DATE"),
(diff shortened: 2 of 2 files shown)


google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -189,6 +189,10 @@ def bigquery_scalar_type(value):
     elif isinstance(value, numbers.Real):
         return "FLOAT64"
     elif isinstance(value, decimal.Decimal):
+        if pyarrow is None:
+            # We don't try to manually guess if it's maybe a BIGNUMERIC.
+            return "NUMERIC"
```
plamut, line 194: Just to confirm, we indeed don't want to inspect the number of digits and try to deduce whether the type is actually BIGNUMERIC? As that would essentially mean duplicating pyarrow's logic?

tswast: Actually, I double-checked the docs, and there is a `Decimial.as_tuple()` method that would be very useful for this.

```python
vtuple = value.as_tuple()
# NUMERIC values have precision of 38 (number of digits)
# and scale of 9 (number of fractional digits), so there can
# be at most 38 - 9 = 29 whole number digits.
# https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
if vtuple.digits <= 38 and vtuple.exponent <= 29:
    return "NUMERIC"
else:
    return "BIGNUMERIC"
```

tswast: Actually, I double-checked the docs, and there is a [`Decimial.as_tuple()` method](https://docs.python.org/3/library/decimal.html#decimal.Decimal.as_tuple) that would be very useful for this.

```python
vtuple = value.as_tuple()
# NUMERIC values have precision of 38 (number of digits)
# and scale of 9 (number of fractional digits), so there can
# be at most 38 - 9 = 29 whole number digits.
# https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
if vtuple.digits <= 38 and vtuple.exponent <= 29:
    return "NUMERIC"
else:
    return "BIGNUMERIC"
```

tswast (changes requested):
Based on my latest comment, I think we can remove `pyarrow` entirely from this logic.

google/cloud/bigquery/dbapi/_helpers.py
plamut: That's precisely the Decimal method that caught my eyes, thus I asked. :+1: 

Yes, let's do that.

plamut: Following the offline debate and posting here for the reference - interpreting the exponent might be somewhat trickier:

```py
>>> Decimal("1.0E+30").as_tuple()
DecimalTuple(sign=0, digits=(1, 0), exponent=29)
>>> Decimal("1E+30").as_tuple()
DecimalTuple(sign=0, digits=(1,), exponent=30)
```
The same number in both cases, but with different exponents. It does not fit into backend's `NUMERIC` (`BIGNUMERIC` should be used), but checking if `exponent <= 29` is not sufficient by itself.

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -189,12 +188,20 @@ def bigquery_scalar_type(value):
...
+        # NUMERIC values have precision of 38 (number of digits) and scale of 9 (number
+        # of fractional digits), and their max absolute value must be strictly smaller
+        # than 1.0E+29.
+        # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
+        if (
+            len(vtuple.digits) <= 38  # max precision: 38
+            and vtuple.exponent >= -9  # max scale: 9
+            and _NUMERIC_SERVER_MIN <= value <= _NUMERIC_SERVER_MAX
+        ):
             return "NUMERIC"
```
plamut, line 201: The logic differs from the [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types), because testing the actual backend response showed that `9.9999999999999999999999999999999999999E+29` is not accepted as a valid NUMERIC value:
```sql
SELECT CAST((SELECT "9.9999999999999999999999999999999999999E+29" AS literal) AS NUMERIC);
```
Even `1.0E+29` is too large, but `9.999...9E+28` works, meaning that there's probably an off-by-one error in the docs for the exponent.

tests/unit/test_dbapi__helpers.py
```diff
@@ -71,6 +72,33 @@ def test_scalar_to_query_parameter(self):
             self.assertEqual(named_parameter.type_, expected_type, msg=msg)
             self.assertEqual(named_parameter.value, value, msg=msg)
 
+    def test_decimal_to_query_parameter(self):  # TODO: merge with previous test
```
plamut, line 75: I should have removed that, it's just a subset of the `test_scalar_to_query_parameter` test above it.

PR #557, opened 2021-03-17
fix: avoid policy tags 403 error in `load_table_from_dataframe`

tswast:
In internal issue 182204971, as customer is encountering a 403 error for missing permissions to set policy tags on a table when trying to append a dataframe to a table with `load_table_from_dataframe`. This is because we get the schema from the table and then pass it back to the API. We only need to set the field names (and maybe type + mode) in this case.

Fixes internal bug 182204971 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2291,9 +2291,13 @@ class Client(ClientWithProject):
                     name
                     for name, _ in _pandas_helpers.list_columns_and_indexes(dataframe)
                 )
-                # schema fields not present in the dataframe are not needed
                 job_config.schema = [
-                    field for field in table.schema if field.name in columns_and_indexes
+                    # Field description and policy tags are not needed to
+                    # serialize a data frame.
+                    SchemaField(field.name, field.field_type, mode=field.mode)
+                    # schema fields not present in the dataframe are not needed
+                    for field in table.schema
+                    if field.name in columns_and_indexes
                 ]
 
         job_config.schema = _pandas_helpers.dataframe_to_bq_schema(
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2294,7 +2294,12 @@ class Client(ClientWithProject):
                 job_config.schema = [
                     # Field description and policy tags are not needed to
                     # serialize a data frame.
-                    SchemaField(field.name, field.field_type, mode=field.mode)
+                    SchemaField(
+                        field.name,
+                        field.field_type,
+                        mode=field.mode,
+                        fields=field.fields,
+                    )
                     # schema fields not present in the dataframe are not needed
                     for field in table.schema
                     if field.name in columns_and_indexes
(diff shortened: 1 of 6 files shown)


google/cloud/bigquery/schema.py
```diff
@@ -73,14 +74,18 @@ def __init__(
...
-        self._field_type = field_type
-        self._mode = mode
-        self._description = description
+        self._properties = {
+            "name": name,
+            "type": field_type,
+        }
+        if mode is not None:
+            self._properties["mode"] = mode.upper()
+        if description is not _DEFAULT_VALUE:
```
tswast, line 87: @shollyman This is one of the key changes: we no longer set the resource value for "description" if it's not explicitly set.

We already omit `policy_tags` from the resource if none (though arguably it should get the same treatment so that someone can unset policy tags from Python)

google/cloud/bigquery/client.py
```diff
@@ -2291,9 +2291,18 @@ def load_table_from_dataframe(
                     name
                     for name, _ in _pandas_helpers.list_columns_and_indexes(dataframe)
                 )
-                # schema fields not present in the dataframe are not needed
                 job_config.schema = [
-                    field for field in table.schema if field.name in columns_and_indexes
+                    # Field description and policy tags are not needed to
+                    # serialize a data frame.
+                    SchemaField(
```
tswast, line 2297: This is the actual bug fix. Rather than populate all properties of schema field from the table schema, just populate the minimum we need to convert to parquet/CSV and then upload

google/cloud/bigquery/schema.py
shollyman: My default inclination would be for special handling for None values to happen at the places where it's significant, like when calling tables.update.  It's also the case that schema fields can't be manipulated individually, so perhaps I'm simply just not thinking this through properly.

google/cloud/bigquery/client.py
shollyman: We'll need to revisit this for parameterization constraints, but that's a problem for future Tim.

shollyman (approved):

google/cloud/bigquery/schema.py
tswast: I called that out as a possibility in https://github.com/googleapis/python-bigquery/issues/558, but that'd require updating our field mask logic to support sub-fields, which gets into some hairy string parsing (perhaps not all that hairy, as it could be as simple as split on `'.'`, but is definitely a departure from what we've been doing).

Also, it might mean that we'd have to introduce a field mask to our load job methods. Based on the error message we're seeing, it sounds like it's possible to make updates to fields like policy tags from a load job.

tswast:
For posterity, here is the error you get if you include `policyTags` in the schema, even if they aren't actually changed:

```
[1] Reason: 403 POST
https://bigquery.googleapis.com/upload/bigquery/v2/projects/YOUR-PROJECT-ID/jobs?uploadType=resumable:
Access Denied: Taxonomy projects/REDACTED/locations/eu/taxonomies/REDACTED: 
User does not have permission to get taxonomy projects/REDACTED/locations/eu/taxonomies/REDACTED\n'
```

PR #560, opened 2021-03-18
feat: add `ExternalConfig.connection_id` property to connect to external sources

tswast:
In response to internal issue b/182112589.

### DIFF
--- a/google/cloud/bigquery/external_config.py
+++ b/google/cloud/bigquery/external_config.py
@@ -774,7 +774,7 @@ class ExternalConfig(object):
         return self._properties.get("connectionId")
 
     @connection_id.setter
-    def connectionid(self, value):
+    def connection_id(self, value):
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/external_config.py
```diff
@@ -760,6 +760,14 @@ def schema(self):
         prop = self._properties.get("schema", {})
         return [SchemaField.from_api_repr(field) for field in prop.get("fields", [])]
 
+    @property
+    def connection_id(self):
```
tswast, line 764: Will likely need the standard "[Experimental]" disclaimer, as this feature is only available to certain customers in a private Preview at the moment.

shollyman (approved):
Can you file an issue on the backlog so we can keep an eye on this for all languages as it matures?

PR #575, opened 2021-03-24
feat: add configurable max size for the queue holding the result pages streamed over the BQ Storage API

### LINKED ISSUE #561: Avoid running out of memory during slow iteration: set worker_queue maxsize to be 1, and/or make it configurable

When using `to_dataframe_iterable` for a large result set (with nested/repeated records) on a system with a fast upstream connection to BigQuery, but _slow_ downstream processing, Python can use all the memory on the system, and get killed.

I think I have tracked it down to the `worker_queue` that is used to pass data from the worker threads back to the main thread: it does not have a `maxsize`.

https://github.com/googleapis/python-bigquery/blob/cc3394f80934419eb00c2029bb81c92a696e7d88/google/cloud/bigquery/_pandas_helpers.py#L657

This means that if items are not pulled from the queue fast enough, then all the memory on the system can be used

I think a reasonable solution would be to set the `maxsize` to be 1:

```python
worker_queue = queue.Queue(maxsize=1)
```

This would still effectively be a buffer size of "number of threads + 1" pages since each thread would fetch into a variable, and _then_ wait if the queue is full.

However, it being configurable would also be reasonable I think.

I've monkey-patched for now, and it seems to work. But ideally, it wouldn't be necessary

```python
# ...

def _monkey_patch_queue_maxsize_1():
    OriginalQueue = queue.Queue

    class QueueWithMaxsize1(OriginalQueue):
        def __init__(self):
            super().__init__(maxsize=1)

    def _restore():
        queue.Queue = OriginalQueue

    queue.Queue = QueueWithMaxsize1

    return _restore

# ....

query = bqClient.query(sql)
result_rows = query.result()

ensure_original_queue = _monkey_patch_queue_maxsize_1()

pages = result_rows.to_dataframe_iterable(bqStorageClient)

for page in pages:
   ensure_original_queue()
   # ... something slow, even time.sleep can do it
```

plamut:
Closes #561.

This PR limits the size of the internal queue that stores result pages when streaming data over the BQ Storage API. It also makes the limit configurable.

Still need to add a few additional unit tests, but that should be it.

**Note:**
The new parameter is _not_ exposed to the `bigquery` Jupyter cell magic - I presume that's fine? I don't think cell magic needs such fine-grained control, since it's not really meant to fetch huge query results into a Jupyter notebook session where any performance difference could actually matter.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1497,11 +1497,7 @@ class RowIterator(HTTPIterator):
         )
         yield from result_pages
 
-    def _to_arrow_iterable(
-        self,
-        bqstorage_client=None,
-        max_queue_size=_pandas_helpers._DEFAULT_MAX_QUEUE_SIZE,
-    ):
+    def _to_arrow_iterable(self, bqstorage_client=None):
         """Create an iterable of arrow RecordBatches, to process the table as a stream."""
         bqstorage_download = functools.partial(
             _pandas_helpers.download_arrow_bqstorage,
@@ -1510,7 +1506,6 @@ class RowIterator(HTTPIterator):
             bqstorage_client,
             preserve_order=self._preserve_order,
             selected_fields=self._selected_fields,
-            max_queue_size=max_queue_size,
         )
         tabledata_list_download = functools.partial(
             _pandas_helpers.download_arrow_row_iterator, iter(self.pages), self.schema
@@ -1528,7 +1523,6 @@ class RowIterator(HTTPIterator):
         progress_bar_type=None,
         bqstorage_client=None,
         create_bqstorage_client=True,
-        max_queue_size=_pandas_helpers._DEFAULT_MAX_QUEUE_SIZE,
     ):
         """[Beta] Create a class:`pyarrow.Table` by loading all pages of a
         table or query.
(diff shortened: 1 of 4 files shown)


tswast:
Looks like some tests are timing out now. I suspect that 1 is not the right default

plamut:
I'm fine with that, I'll remove the parameter from other methods where it's expected that query results are downloaded in full. Will also check the timeouts and what a better default could be.

google/cloud/bigquery/table.py
```diff
@@ -1642,6 +1646,17 @@ def to_dataframe_iterable(self, bqstorage_client=None, dtypes=None):
...
                 specified. Otherwise, the default pandas behavior is used.
 
+            max_queue_size (Optional[int]):
+                The maximum number of result pages to hold in the internal queue when
+                streaming query results over the BigQuery Storage API. Ignored if
+                Storage API is not used.
+
+                By default, the max queue size is set to the number of BQ Storage streams
+                created by the server. If ``max_queue_size`` is :data:`None`, the queue
+                size is infinite.
```
plamut, line 1656: Just in case somebody _really_ wants the old behavior, I added it as an option.

plamut:
@tswast ping :)

tswast (approved):
Wonderful!

PR #591, opened 2021-04-06
docs: add sample to run DML query

tswast:

Towards internal issue 178085085

### DIFF
--- a/samples/snippets/conftest.py
+++ b/samples/snippets/conftest.py
@@ -12,8 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
+import random
+
 from google.cloud import bigquery
 import pytest
+
+
+def resource_prefix() -> str:
+    timestamp = datetime.datetime.utcnow().strftime("%Y%m%d_%H%M%S")
+    random_string = hex(random.randrange(1000000))[2:]
+    return f"python_bigquery_samples_snippets_{timestamp}_{random_string}"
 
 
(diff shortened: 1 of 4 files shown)


tswast:
Re-assigning based on OOO

tseaver (approved):
LGTM

samples/snippets/conftest.py
```diff
@@ -12,10 +12,35 @@
...
+    yesterday = datetime.datetime.utcnow() - datetime.timedelta(days=1)
+    for dataset in bigquery_client.list_datasets():
+        if (
+            dataset.dataset_id.startswith(RESOURCE_PREFIX)
+            and dataset.created < yesterday
+        ):
+            bigquery_client.delete_dataset(
+                dataset, delete_contents=True, not_found_ok=True
+            )
+
```
tmatsuo, line 42: This cleanup code looks great. Since `dataset` has `created` property so technically there's no need to have timestamp in the resource name, but it's also good when we manually check the resources.

Only concern is this code might fail and the test fails too. It might annoy us. If it annoys us too much, maybe we can consider ignoring failure here.

tmatsuo (approved):

samples/snippets/conftest.py
tswast: The `not_found_ok` will ignore the most common error of 404. Other than that, I've not found the `delete_dataset` operation to be all that flaky with the default retry options.

Issue #556, opened 2021-03-17
ValueError: Protocol message FieldDescriptorProto has no "proto3_optional" field


Tsakunelson:
Hi All,
I am still having this same error, though have all the dependencies installed on a Jupyter Notebook in AWS Sagemaker including:

`Proto-plus = 1.18.0
Protobuf = 3.13.0`

For Google BigQuery 2.12.0

But I get the following error:

```---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-017a44a94d83> in <module>
----> 1 df = gbq.read_gbq('SELECT * FROM `cmg-consumer-analytics.122142903.ga_sessions_20210314` LIMIT 10', project_id='cmg-consumer-analytics')

~/anaconda3/envs/python3/lib/python3.6/site-packages/pandas/io/gbq.py in read_gbq(query, project_id, index_col, col_order, reauth, auth_local_webserver, dialect, location, configuration, credentials, use_bqstorage_api, private_key, verbose, progress_bar_type)
    161     DataFrame.to_gbq : Write a DataFrame to Google BigQuery.
    162     """
--> 163     pandas_gbq = _try_import()
    164 
    165     kwargs: Dict[str, Union[str, bool]] = {}

~/anaconda3/envs/python3/lib/python3.6/site-packages/pandas/io/gbq.py in _try_import()
     15         "See the docs: https://pandas-gbq.readthedocs.io."
     16     )
---> 17     pandas_gbq = import_optional_dependency("pandas_gbq", extra=msg)
     18     return pandas_gbq
     19 

~/anaconda3/envs/python3/lib/python3.6/site-packages/pandas/compat/_optional.py in import_optional_dependency(name, extra, raise_on_missing, on_version)
     87     )
     88     try:
---> 89         module = importlib.import_module(name)
     90     except ImportError:
     91         if raise_on_missing:

~/anaconda3/envs/python3/lib/python3.6/importlib/__init__.py in import_module(name, package)
    124                 break
    125             level += 1
--> 126     return _bootstrap._gcd_import(name[level:], package, level)
    127 
    128 

~/anaconda3/envs/python3/lib/python3.6/importlib/_bootstrap.py in _gcd_import(name, package, level)

~/anaconda3/envs/python3/lib/python3.6/importlib/_bootstrap.py in _find_and_load(name, import_)

[...]

PR #593, opened 2021-04-07
feat: use pyarrow stream compression, if available

### LINKED ISSUE #579: feat: make use of LZ4 in `to_arrow` / `to_dataframe`
https://github.com/googleapis/python-bigquery-storage/pull/166 adds a compression option to Arrow streams. This can greatly speed-up some workloads.

Requires `pyarrow >=1.0.0`.

plamut:
Closes #579.

This PR uses `pyarrow` compression of BQ Storage streams when available, i.e. when a recent enough version of `google-cloud-bigquery-storage` is installed.

Currently blocked on the BQ Storage [release](https://github.com/googleapis/python-bigquery-storage/pull/170) that will add this support, but it's still possible to test this PR locally by changing the noxfile to use a development version of `google-cloud-bigquery-storage`.

**PR checklist**

### DIFF
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -38,12 +38,8 @@ try:
 except ImportError:
     _ARROW_COMPRESSION_SUPPORT = False
 else:
-    import pkg_resources
-
-    # Having BQ Storage available implies that pyarrow is available, too.
-    _ARROW_COMPRESSION_SUPPORT = pkg_resources.get_distribution(
-        "pyarrow"
-    ).parsed_version >= pkg_resources.parse_version("1.0.0")
+    # Having BQ Storage available implies that pyarrow >=1.0.0 is available, too.
+    _ARROW_COMPRESSION_SUPPORT = True
 
 from google.cloud.bigquery import schema
 
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -33,6 +33,18 @@ try:
 except ImportError:  # pragma: NO COVER
     pyarrow = None
 
+try:
+    from google.cloud.bigquery_storage import ArrowSerializationOptions
+except ImportError:
+    _ARROW_COMPRESSION_SUPPORT = False
+else:
+    import pkg_resources
+
+    # Having BQ Storage available implies that pyarrow is available, too.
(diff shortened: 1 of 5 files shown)


google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -33,6 +33,18 @@
...
+    from google.cloud.bigquery_storage import ArrowSerializationOptions
+except ImportError:
+    _ARROW_COMPRESSION_SUPPORT = False
+else:
+    import pkg_resources
+
+    # Having BQ Storage available implies that pyarrow is available, too.
+    _ARROW_COMPRESSION_SUPPORT = pkg_resources.get_distribution(
+        "pyarrow"
+    ).parsed_version >= pkg_resources.parse_version("1.0.0")
```
plamut, line 46: Do we actually need `pyarrow` version check?

The `bqstorage` extra already pins the minimum `pyarrow` version to `1.0.0`, thus I suppose if somebody somehow installs a less recent version, it can be considered an error on the user side?

tswast: Sounds good to me (removing this check). Since all of our extras require `> 1.0.0`, I'm okay with failing.

Counterpoint though is: what error will they get when `pyarrow` is too low? I wonder if we should file an FR to check for minimum versions and throw nicer errors? It might help with some of the issues like https://github.com/googleapis/python-bigquery/issues/556. This would mean keeping minimum versions in sync with 3 locations: `setup.py`, `constraints.txt`, and `version_check.py`, though.

tswast:
> Currently blocked on the BQ Storage release that will add this support

Oops! I'll see what open PRs we have and hopefully cut a release today.

google/cloud/bigquery/_pandas_helpers.py
plamut: The new pip dependency resolver ([announcement](https://discuss.python.org/t/announcement-pip-20-2-release/4863)) is significantly stricter, thus I suppose this will become a no-issue once its use is widespread enough and/or becomes a default. It shouldn't even be possible to install an incompatible `pyarrow` version with it.

Yes, we can open a FR to discuss there if covering this presumably corner-case scenario in the transitional period is worth doing, considering the problem with keeping the min versions in sync.

plamut:
I see that a new version of BQ Storage has been [released](https://github.com/googleapis/python-bigquery-storage/pull/170), good. But I now also see that one of the Python 3.6 unit tests fails, will investigate.

tswast (approved):

PR #594, opened 2021-04-07
docs: update the description of the return value of `_QueryResults.rows()`

kdeggelman:
Updated the description of the return value of `rows` to be more accurate.


plamut (approved):
Looks good, thanks for that!

PR #613, opened 2021-04-15
feat: add type hints for public methods

### LINKED ISSUE #157: feat: add type hints to methods

plamut:
Closes #157.
Supersedes #445.

**PR checklist:**


noxfile.py
```diff
@@ -90,6 +91,14 @@ def unit(session):
     default(session)
 
 
+@nox.session(python="3.8")
+def pytype(session):
```
tswast, line 95: Can we add this to `nox.options.sessions` so that it runs in the `Kokoro` build?

tswast:
Beautiful! I'm OOO until Tuesday, so I can't take a look at this too closely right now, but happy to get it merged assuming we get it running in Kokoro and someone else reviews.

Oh, and can we get `pytype` running in our `pre-commit` config?

plamut:
> Oh, and can we get pytype running in our pre-commit config?

The check is too slow, unfortunately, it takes like ~90 seconds on my machine, which disrupts the workflow too much. Unless, of course, your experience is different and it's my machine that needs some tweaking.

I noticed that, `pytype` does seem to cache the results, but when I added a trivial change (a single blank line in `client.py`), running it once more again took the same as before.

noxfile.py
plamut: Done. From [Kokoro log](https://source.cloud.google.com/results/invocations/fa869575-8829-4345-8cbc-ed10b88abed7/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fpresubmit/log):
```
...
nox > Session blacken was successful.
nox > Running session pytype
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/pytype
nox > python -m pip install pytype==2021.4.9
nox > pytype
Computing dependencies
Analyzing 40 sources with 7 local dependencies
ninja: Entering directory `.pytype'
[1/69] check google.cloud.bigquery.encryption_configuration
...
```

google/cloud/bigquery/client.py
```diff
@@ -1545,12 +1649,18 @@ def delete_table(
...
     def _get_query_results(
-        self, job_id, retry, project=None, timeout_ms=None, location=None, timeout=None,
+        self,
+        job_id: str,
+        retry: retries.Retry,
+        project: str = None,
+        timeout_ms: int = None,
+        location: str = None,
+        timeout: float = None,
     ):
```
tswast, line 1664: Any reason we don't have a return type here?

```suggestion
    ) -> _QueryResults:
```

google/cloud/bigquery/client.py
```diff
@@ -1609,7 +1719,7 @@ def _get_query_results(
         )
         return _QueryResults.from_api_repr(resource)
 
-    def job_from_resource(self, resource):
+    def job_from_resource(self, resource: dict) -> job.UnknownJob:
```
tswast, line 1722: Can we do a union type here?

Alternatively, what do you think about introducing some `abc`-style module for all these different job classes? I know we have the [_AsyncJob](https://github.com/googleapis/python-bigquery/blob/6502a602337ae562652a20b20270949f2c9d5073/google/cloud/bigquery/job/base.py#L131) class, but that's private and also poorly named given that it's not usable in true `async` code.

google/cloud/bigquery/job/base.py
```diff
@@ -633,7 +649,9 @@ def done(self, retry=DEFAULT_RETRY, timeout=None, reload=True):
             self.reload(retry=retry, timeout=timeout)
         return self.state == _DONE_STATE
 
-    def result(self, retry=DEFAULT_RETRY, timeout=None):
+    def result(
+        self, retry: retries.Retry = DEFAULT_RETRY, timeout: float = None
+    ) -> "_AsyncJob":
```
tswast, line 654: Since some job classes do use this implementation without overriding it, I wonder if this is more evidence that we should make `_AsyncJob` public (just `Job`?) / make a separate `abc` type for jobs?

google/cloud/bigquery/job/base.py
```diff
@@ -831,7 +852,7 @@ def _fill_from_default(self, default_job_config):
         return new_job_config
 
     @classmethod
-    def from_api_repr(cls, resource):
+    def from_api_repr(cls, resource: dict) -> "_JobConfig":
```
tswast, line 855: Maybe we need `abc` types for configurations too?

google/cloud/bigquery/job/query.py
```diff
@@ -1265,12 +1275,12 @@ def to_arrow(
...
-        progress_bar_type=None,
-        create_bqstorage_client=True,
-        date_as_object=True,
-    ):
+        bqstorage_client: bigquery_storage.BigQueryReadClient = None,
+        dtypes: Dict[str, Any] = None,
+        progress_bar_type: str = None,
+        create_bqstorage_client: bool = True,
+        date_as_object: bool = True,
+    ) -> Any:
```
tswast, line 1283: Why `Any` and not `"pandas.DataFrame"`?

google/cloud/bigquery/job/query.py
```diff
@@ -1196,10 +1206,10 @@ def result(
...
     def to_arrow(
         self,
-        progress_bar_type=None,
-        bqstorage_client=None,
-        create_bqstorage_client=True,
-    ):
+        progress_bar_type: str = None,
+        bqstorage_client: bigquery_storage.BigQueryReadClient = None,
+        create_bqstorage_client: bool = True,
+    ) -> Any:
```
tswast, line 1212: Why `Any` and not `"pyarrow.Table"`?

google/cloud/bigquery/table.py
```diff
@@ -47,6 +47,10 @@
 from google.cloud.bigquery.external_config import ExternalConfig
 from google.cloud.bigquery.encryption_configuration import EncryptionConfiguration
 
+# Types needed only for Type Hints
+from google.cloud import bigquery_storage  # pytype: disable=import-error
```
tswast, line 51: As much as I think it'd be quite useful to depend on `google-cloud-bigquery-storage` being present, I don't think we wanted to make `google-cloud-bigquery-storage` a required dependency just for this feature.

How do other packages support cross-package types for their "extras"?

tswast (changes requested):
Looking good!

I have a few questions about how we want to handle job classes, but if we make any changes beyond a few more `Unions` such as `abc` types, I think that warrants a separate PR.

Also curious about cross-package types. I could actually probably be convinced to make `google-cloud-bigquery-storage` a required dependency, but I don't want to do so by accident.

tswast:
Looks like we can conditionally import, only if we are currently type checking: https://stackoverflow.com/a/57940134/101923

Not sure if `pytype` respects this or if it's just a `mypy` thing.

google/cloud/bigquery/job/query.py
plamut: Perhaps because of [this](https://github.com/googleapis/python-bigquery/pull/445#discussion_r549215027)? I'll nevertheless check again if a quick workaround can be found.

google/cloud/bigquery/table.py
plamut: It appears pytype supports this as well, will use it.

https://github.com/google/pytype/blob/master/docs/faq.md#how-do-i-write-code-that-is-seen-by-pytype-but-ignored-at-runtime

plamut:
> Looks like we can conditionally import, only if we are currently type checking: https://stackoverflow.com/a/57940134/101923
> 
> Not sure if `pytype` respects this or if it's just a `mypy` thing.

It appears pytype supports this as well, will use it.

https://github.com/google/pytype/blob/master/docs/faq.md#how-do-i-write-code-that-is-seen-by-pytype-but-ignored-at-runtime

google/cloud/bigquery/job/query.py
plamut: Good news - I figured out that if we install all dependencies in the nox session before running the type checks, A LOT of these import errors disappear, and we don't need to make workarounds such as this one or silencing the checker all around the code.

google/cloud/bigquery/client.py
plamut: I kind of dislike "wide" union hints that clutter the code, thus introducing a base class for these type sounds reasonable. But perhaps in a separate PR, since that's an actual code refactoring, and not merely annotations that don't affect the runtime?

google/cloud/bigquery/job/base.py
plamut: If that's not something end users should know about, I'd prefer keeping it private, but I support the renaming idea - "async" is too often automatically associated with `async.io` nowadays.

Ditto for adding an `ABC`, if we decide for that with other class families, such as job configs.

plamut:
Whoa, the very first test that I wrote in this library caught a method signature mismatch (incompatible annotations). :)

Will fix it.

plamut:
I don't know why imports directly from `google.cloud` still cause false import errors, but fortunately there are not many of them. On the other hand, it was possible to get rid of the vast majority of those noisy "ignore import error" directives.

tswast (commented):
Looks great!!!!

I'm holding off on approval because I'd like a unit test nox session implemented that doesn't have any "extras" so we can verify that at least minimal functionality is supported if pandas/pyarrow/google-cloud-bigquery-storage aren't installed https://github.com/googleapis/python-bigquery/issues/621

I trust that `if typing.TYPE_CHECKING:` is working as expected, but I'd like us to verify too

tswast (approved):
LGTM assuming Kokoro passes

tswast:
Marking CLI as yes, as this PR is based on work done for hire last year.

PR #617, opened 2021-04-15
feat: accept job object as argument to `get_job` and `cancel_job`

### LINKED ISSUE #616: get_job, cancel_job should accept Job classes in addition to job_id
If a Job class is passed in, use that location instead of the location from the argument parameters.

tswast:
This allows one to more easily cancel or get updated metadata for an
existing job from the client class. Ensures that project ID and location
are correctly populated.

Fixes #616 🦕

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -2971,7 +2971,11 @@ class TestClient(unittest.TestCase):
         QUERY = "SELECT * from test_dataset:test_table"
         ASYNC_QUERY_DATA = {
             "id": "{}:{}".format(self.PROJECT, JOB_ID),
-            "jobReference": {"projectId": self.PROJECT, "jobId": "query_job"},
+            "jobReference": {
+                "projectId": "job-based-proj",
+                "jobId": "query_job",
+                "location": "us-east1",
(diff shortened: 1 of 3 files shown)


tests/unit/test_client.py
```diff
@@ -2971,7 +2971,11 @@ def test_get_job_hit_w_timeout(self):
         QUERY = "SELECT * from test_dataset:test_table"
         ASYNC_QUERY_DATA = {
             "id": "{}:{}".format(self.PROJECT, JOB_ID),
-            "jobReference": {"projectId": self.PROJECT, "jobId": "query_job"},
+            "jobReference": {
+                "projectId": "job-based-proj",
```
shollyman, line 2975: nit: job-based-proj reads very oddly to me.  maybe some variant of proj-containing-job?  project-owning-job?

shollyman (approved):

tests/unit/test_client.py
tswast: How about `resource-proj`? The distinction I'm trying to communicate is that this project ID is coming from the job argument and not the client default.

PR #330, opened 2020-10-18
docs(samples): add more clustering code snippets

ryanyuan:
Add sample code for creating a clustered table from a query result.
samples/client_query_destination_table_clustered.py
Section: [https://cloud.google.com/bigquery/docs/creating-clustered-tables#creating_a_clustered_table_from_a_query_result](https://cloud.google.com/bigquery/docs/creating-clustered-tables#creating_a_clustered_table_from_a_query_result)

Add sample code for creating a clustered table when you load data.
samples/load_table_clustered.py
Section: [https://cloud.google.com/bigquery/docs/creating-clustered-tables#creating_a_clustered_table_when_you_load_data](https://cloud.google.com/bigquery/docs/creating-clustered-tables#creating_a_clustered_table_when_you_load_data)

Fixes #329  🦕

PR #286, opened 2020-09-30
docs: update clustering field docstrings

ryanyuan:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #285  🦕

PR #622, opened 2021-04-21
fix: unsetting clustering fileds on Table is now possible

### LINKED ISSUE #615: FR: support mutable clustering configuration
BigQuery service released support [mutable clustering configuration](https://cloud.google.com/bigquery/docs/creating-clustered-tables#modifying-cluster-spec) some time ago, but this hasn't propagated to all clients.

For example, in the docs for python library, there's comments

> Clustering fields are immutable after table creation.

Library should ensure:

* Clustering configuration can be modified on a managed table via a table update call
* Documentation reflects this capability, and possibly links to the cloud.google.com content above to describe the behavior when clustering is added/changed.
* Stale comments indicating that clustering requires a partitioning config are removed (you can cluster a table without partitioning it)

plamut:
Closes #615.

Support for mutable clustering configuration already existed, but due to a bug it was not possible to unset the clustering fields altogether. This PR fixes that, and makes some refactoring changes to `table.py`.

**PR checklist:**
    - [x] _"Documentation reflects this capability, and possibly links to the cloud.google.com content above to describe the behavior when clustering is added/changed."_ - DONE in [#330](https://github.com/googleapis/python-bigquery/pull/330/files) (samples using clustering on a non-partitioned table).
    - [x] _"Stale comments indicating that clustering requires a partitioning config are removed (you can cluster a table without partitioning it)"_ - DONE in [#286](https://github.com/googleapis/python-bigquery/pull/286/files)


google/cloud/bigquery/table.py
```diff
@@ -291,16 +284,36 @@ class Table(object):
...
         "partitioning_type": "timePartitioning",
+        "project": ["tableReference", "projectId"],
+        "range_partitioning": "rangePartitioning",
+        "time_partitioning": "timePartitioning",
+        "schema": "schema",
+        "streaming_buffer": "streamingBuffer",
+        "self_link": "selfLink",
+        "table_id": ["tableReference", "tableId"],
         "time_partitioning": "timePartitioning",
+        "type": "type",
```
tswast, line 316: I'm pretty sure we fallback to the passed-in name if we don't find it here, but I'm fine including everything for consistency.

google/cloud/bigquery/table.py
```diff
@@ -316,17 +329,23 @@ def __init__(self, table_ref, schema=None):
     @property
     def project(self):
         """str: Project bound to the table."""
-        return self._properties["tableReference"]["projectId"]
+        return _helpers._get_sub_prop(
+            self._properties, self._PROPERTY_TO_API_FIELD["project"]
```
tswast, line 333: Ah! That makes sense. I bet we could go one level further and do some metaclass magic to generate properties, but metaclasses are one level of magic too far in my opinion. This looks good to me.

google/cloud/bigquery/table.py
```diff
@@ -636,12 +674,13 @@ def clustering_fields(self, value):
...
+        api_field = self._PROPERTY_TO_API_FIELD["clustering_fields"]
+
         if value is not None:
-            prop = self._properties.setdefault("clustering", {})
+            prop = self._properties.setdefault(api_field, {})
             prop["fields"] = value
         else:
-            if "clustering" in self._properties:
-                del self._properties["clustering"]
+            self._properties[api_field] = None
```
tswast, line 683: Let's add a comment to explain why we need to unset the top-level field to `None`. (And thus can't use `_helpers._set_sub_prop` like some of the other properties in this class)

tswast (commented):
Sweet! Just one suggestion, but otherwise looks good. Thanks for adding that system test.

google/cloud/bigquery/table.py
plamut: I always thought a custom dict subtype could be useful here, but never convinced myself that we really need it - even though it would be less arcane than using metaclasses.

Proof of concept:

```py
class ApiProperties(dict):
    def __getitem__(self, key):
        if not isinstance(key, tuple):
            return super().__getitem__(key)
        result = super()
        for item in key:
            result = result.__getitem__(item)
        return result

>>> p = ApiProperties({"foo": {"subfoo": "bar"}})
>>> p
{'foo': {'subfoo': 'bar'}}
>>> p["foo"]["subfoo"]
'bar'
>>> p["foo", "subfoo"]  # also works
'bar'
```

The customized dict would assume that the keys are never tuples (which is true, we just use strings), and it could override `__setitem__()` in a similar fashion. An it could also automatically create subdictionaries if a (sub) key does not exist yet, similarly to `defaultdict`. And more...

google/cloud/bigquery/table.py
plamut: ... and _explicit is better than implicit_. :)

(also, the [routine](https://github.com/googleapis/python-bigquery/blob/0abb56669c097c59fbffce007c702e7a55f2d9c1/google/cloud/bigquery/routine/routine.py#L41-L54) implementation takes the same approach)

google/cloud/bigquery/table.py
tswast: I hadn't thought about subclassing what we use for dictionaries. That would indeed help clean thing up a bit, assuming it's still JSON-serializable.

tswast (approved):

PR #623, opened 2021-04-22
chore: add unit test nox session without extras installed

### LINKED ISSUE #621: add test session to nox without installing any "extras"
https://github.com/googleapis/python-bigquery/pull/613 is making me a bit nervous that we might accidentally introduce a required dependency that we thought was optional. It wouldn't be the first time this has happened (https://github.com/googleapis/python-bigquery/issues/549), so I'd like at least a unit test session that runs without any extras.

plamut:
Closes #621. 🦕

The new check revealed a few cases where not installing extras caused some of the tests to fail, but that was due to test setup, and not the codebase itself.

**PR checklist:**

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -31,6 +31,7 @@ CURRENT_DIRECTORY = pathlib.Path(__file__).parent.absolute()
 
 # 'docfx' is excluded since it only needs to run in 'docs-presubmit'
 nox.options.sessions = [
+    "unit_noextras",
     "unit",
     "system",
     "snippets",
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -90,6 +92,12 @@ def unit(session):
     default(session)
 
 
+@nox.session(python=UNIT_TEST_PYTHON_VERSIONS[-1])
+def unit_noextras(session):
```
plamut, line 96: Presumably using just the latest version is fine?
(the CI will finish faster)

On rare occasions an extra requires a more recent Python version, but I can't recall any cases when an extra would be restricted to _older_ Python versions.

tswast: Sounds good to me. We could also use `DEFAULT_PYTHON_VERSION` here, though that's not always the latest.

tswast (approved):

PR #624, opened 2021-04-22
feat: retry google.auth TransportError by default

### LINKED ISSUE #620: google.auth.exceptions.TransportError is not being retried
Hi, 

Recently i faced an error can you please consider using this exception as a retry one also. Since i have faced this error in one of our production system

https://github.com/googleapis/python-storage/issues/414

plamut:
Closes #620.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/retry.py
+++ b/google/cloud/bigquery/retry.py
@@ -14,6 +14,7 @@
 
 from google.api_core import exceptions
 from google.api_core import retry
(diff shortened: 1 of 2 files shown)


tswast (approved):
LGTM.

Note: I see that https://github.com/googleapis/python-storage/pull/418 tries to be a bit more specific about which TransportErrors it retries, but I don't actually see a problem with applying our retry logic on any low-level transport errors.

PR #625, opened 2021-04-23
chore(revert): revert preventing normalization

dandhlee:
reverts previous commit for preventing normalization of versioning

Reverts #611.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ if "google.cloud" in packages:
 
 setuptools.setup(
     name=name,
-    version=setuptools.sic(version),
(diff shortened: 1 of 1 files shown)


plamut (commented):
Why do we need a revert, did the original change cause some sort of regression?

Looks good otherwise.

tswast:
> Why do we need a revert, did the original change cause some sort of regression?

It forces folks to use a much newer version of pip/setuptools than we actually want to require. The docs team will workaround the version parsing issue in a different way.

tswast (approved):

PR #626, opened 2021-04-26
feat: Extended DB API parameter syntax to optionally provide parameter types

### LINKED ISSUE #609: None or empty array parameter values aren't handled in the dbapi
```
>>> import google.cloud.bigquery.dbapi
>>> conn = google.cloud.bigquery.dbapi.connect()
/home/jim/p/g/python-bigquery/google/cloud/bigquery/client.py:444: UserWarning: Cannot create BigQuery Storage client, the dependency google-cloud-bigquery-storage is not installed.
  warnings.warn(
>>> cursor = conn.cursor()
>>> cursor.execute("select %(x)s", dict(x=None))
  File "/home/jim/p/g/python-bigquery/google/cloud/bigquery/dbapi/_helpers.py", line 270, in with_closed_check
    return method(self, *args, **kwargs)
  File "/home/jim/p/g/python-bigquery/google/cloud/bigquery/dbapi/cursor.py", line 173, in execute
    query_parameters = _helpers.to_query_parameters(parameters)
  File "/home/jim/p/g/python-bigquery/google/cloud/bigquery/dbapi/_helpers.py", line 167, in to_query_parameters
    return to_query_parameters_dict(parameters)
  File "/home/jim/p/g/python-bigquery/google/cloud/bigquery/dbapi/_helpers.py", line 146, in to_query_parameters_dict
    param = scalar_to_query_parameter(value, name=name)
  File "/home/jim/p/g/python-bigquery/google/cloud/bigquery/dbapi/_helpers.py", line 53, in scalar_to_query_parameter
    raise exceptions.ProgrammingError(
google.cloud.bigquery.dbapi.exceptions.ProgrammingError: encountered parameter x with value None of unexpected type
```

The same issue applies to empty array (`[]`) parameters.

jimfulton:

Fixes #609 🦕


google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -445,9 +459,61 @@ def _format_operation(operation, parameters=None):
...
 
     if isinstance(parameters, collections_abc.Mapping):
-        return _format_operation_dict(operation, parameters)
+        return _format_operation_dict(operation, parameters), parameter_types
+
+    return _format_operation_list(operation, parameters), parameter_types
+
+
+def _extract_types(
+    operation, extra_type_sub=re.compile(r"%(\(([^:)]*)(:(\w+))?\))?s").sub,
```
plamut, line 477: Random thought - how to we treat strings like the following?
```
SELECT "%%(name:type)s" as weird_literal;
```

Percent-formatted strings treat double percent signs as a literal percent and would keep the following string intact:
```py
>>> 'SELECT "%%(name)s" as weird_literal;' % {"name": "foo"}
'SELECT "%(name)s" as weird_literal;'
```
Presumably our extension should follow suit for consistency, and would thus benefit from a negative lookbehind assertion?
```py
r"(?<!%)%(\(([^:)]*)(:(\w+))?\))?s"
```

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -445,9 +459,61 @@ def _format_operation(operation, parameters=None):
...
+    For every parameter of the form %(name:type)s, replace with %(name)s and add the
+    item name->type to dict that's returned.
+
+    Returns operation without type information and a dictionary of names and types.
+    """
+    parameter_types = None
+
+    def repl(m):
+        nonlocal parameter_types
+        _, name, _, type_ = m.groups()
```
plamut, line 490: We could simplify this to
```py
name, type = m.groups()
```
if we mark two of the groups in the pattern as non-capturing:
```
r"%(?:\(([^:)]*)(?::(\w+))?\))?s"
```

plamut (commented):
Not a complete review by any means, just glanced over the PR and the regex pattern caught my eye, gave have two improvement suggestions.

google/cloud/bigquery/dbapi/cursor.py
jimfulton: +1

google/cloud/bigquery/dbapi/cursor.py
jimfulton: Cool. Will do.

google/cloud/bigquery/dbapi/cursor.py
jimfulton: done

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -26,9 +26,42 @@
 
 _NUMERIC_SERVER_MIN = decimal.Decimal("-9.9999999999999999999999999999999999999E+28")
 _NUMERIC_SERVER_MAX = decimal.Decimal("9.9999999999999999999999999999999999999E+28")
+_BIGQUERY_SCALAR_TYPES = {
```
tswast, line 29: Any chance we can use https://github.com/googleapis/python-bigquery/blob/4396e70771af6889d3242c37c5ff2e80241023a2/google/cloud/bigquery/enums.py#L219 or https://github.com/googleapis/python-bigquery/blob/4396e70771af6889d3242c37c5ff2e80241023a2/google/cloud/bigquery/enums.py#L145 instead? I worry this will get out-of-sync (already has, missing GEOGRAPHY, which just encodes as a string)

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -26,9 +26,42 @@
...
+            raise exceptions.ProgrammingError(
+                f"The given parameter type, {query_parameter_type},"
+                f" for {name} is not a valid BigQuery scalar type."
+            )
+    else:
+        parameter_type = bigquery_scalar_type(value)
+        if parameter_type is None:
+            raise exceptions.ProgrammingError(
+                f"Encountered parameter {name} with "
+                f"{value_doc}value {value} of unexpected type."
```
tswast, line 59: missing space?

```suggestion
                f"{value_doc} value {value} of unexpected type."
```

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -209,8 +217,14 @@ def executemany(self, operation, seq_of_parameters):
             seq_of_parameters (Union[Sequence[Mapping[str, Any], Sequence[Any]]]):
                 Sequence of many sets of parameter values.
         """
-        for parameters in seq_of_parameters:
-            self.execute(operation, parameters)
+        if seq_of_parameters:
+            formatted_operation, parameter_types = _format_operation(
+                operation, seq_of_parameters[0]
```
tswast, line 222: Could you describe in a comment why we grab just the first one? I guess either we get the types correct on the first go, or we have a problem anyway?

~Also it's a niche case, but it's possible that someone could pass in a generator or something for the sequence of parameters. In which case `[0]` won't work. I think we need to either always loop like we had (and add some logic to only extract the types once) or do some funny business with the low-level iteration function.~

~I probably should have had a test case with this (generator as parameters) when I documented the allowed type as a Sequence.~

Edit: Nevermind, Sequence supports integer-based item access https://docs.python.org/3/glossary.html#term-sequence It's "Iterable" that only supports iteration. We might actually want to support this use case, but probably best to wait until we actually get a customer request for that feature.

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: Yes! I didn't know those existed.

So some questions:

- `SqlParameterScalarTypes` includes some aliases.  
  Is that something we want in this context? (If so, I'll lobby for `INT`. ;))

- `SqlParameterScalarTypes` and `_SQL_SCALAR_TYPES` lack `DECIMAL` and `BIGDECIMAL`. 
   https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
   
   Should I add `DECIMAL` and `BIGDECIMAL`?

google/cloud/bigquery/dbapi/cursor.py
jimfulton: Comment added.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -445,9 +464,61 @@ def _format_operation(operation, parameters=None):
...
 
     if isinstance(parameters, collections_abc.Mapping):
-        return _format_operation_dict(operation, parameters)
+        return _format_operation_dict(operation, parameters), parameter_types
+
+    return _format_operation_list(operation, parameters), parameter_types
+
+
+def _extract_types(
+    operation, extra_type_sub=re.compile(r"(?<!%)%(?:\(([^:)]*)(?::(\w+))?\))?s").sub,
```
plamut, line 482: Meh... when I woke up, I realized that the negative lookbehind fails if there  is a _triple_ precent sign in the string, because that preceding percent is actually escaped by a yet another percent and does not affect the percent that starts our substitution placeholder.

The following test case fails:
```py
(
    "values(%%%(foo:INT64)s, %(bar)s)",
    ("values(%%%(foo)s, %(bar)s)", dict(foo="INT64")),
)
```

The negative lookbehind should actually match an odd number of percent characters before the actual `%(...)` pattern (because if there's even number of them, they do not escape the starting `%` in the potential match).

An even number of `%`s can be matched with the following:
```
[^%]*((%%)+[^%]*)+
```
but that can't be used as a lookbehind, because it's not fixed width, and I'm starting to dislike this rabbit hole. :rabbit: :hole: (is the mini-formatting language even regular?)

We might be better-off if we simply catch all preceding `%` chars and manually check their length in the processing cod. The following worked for me, although I did not test it that thoroughly, and mostly relied on the existing test cases:

```patch
diff --git google/cloud/bigquery/dbapi/cursor.py google/cloud/bigquery/dbapi/cursor.py
index f4ff0a3..1d1fc7f 100644
--- google/cloud/bigquery/dbapi/cursor.py
+++ google/cloud/bigquery/dbapi/cursor.py
@@ -479,7 +479,7 @@ def _format_operation(operation, parameters):
 
 
 def _extract_types(
-    operation, extra_type_sub=re.compile(r"(?<!%)%(?:\(([^:)]*)(?::(\w+))?\))?s").sub,
+    operation, extra_type_sub=re.compile(r"(%*)%(?:\(([^:)]*)(?::(\w+))?\))?s").sub,
 ):
     """Remove type information from parameter placeholders.
 
@@ -492,7 +492,13 @@ def _extract_types(
 
     def repl(m):
         nonlocal parameter_types
-        name, type_ = m.groups()
+        percent_prefix, name, type_ = m.groups()
+        if len(percent_prefix) % 2:
+            # An odd number of % chars precede the match, meaning that the
+            # leading percent in %(...) is actually escaped and should not be treated
+            # as a replacement placeholder.
+            return m.group()
+
         try:
             if name:
                 if not parameter_types:
@@ -510,7 +516,7 @@ def _extract_types(
                     if not isinstance(parameter_types, dict):
                         raise TypeError()
 
-                return f"%({name})s"
+                return f"{percent_prefix}%({name})s"
             else:
                 if parameter_types is None:
                     parameter_types = []
diff --git tests/unit/test_dbapi_cursor.py tests/unit/test_dbapi_cursor.py
index 9f25a24..5afe269 100644
--- tests/unit/test_dbapi_cursor.py
+++ tests/unit/test_dbapi_cursor.py
@@ -787,6 +787,18 @@ class TestCursor(unittest.TestCase):
         ),
         (" %s %()s %(:int64)s ", (" %s %s %s ", [None, None, "int64"])),
         (" %%s %s %()s %(:int64)s ", (" %%s %s %s %s ", [None, None, "int64"])),
+        (
+            "values(%%%(foo:INT64)s, %(bar)s)",
+            ("values(%%%(foo)s, %(bar)s)", dict(foo="INT64")),
+        ),
+        (
+            "values(%%%%(foo:INT64)s, %(bar)s)",
+            ("values(%%%%(foo:INT64)s, %(bar)s)", dict()),
+        ),
+        (
+            "values(%%%%%(foo:INT64)s, %(bar)s)",
+            ("values(%%%%%(foo)s, %(bar)s)", dict(foo="INT64")),
+        ),
     ],
 )
 def test__extract_types(inp, expect):
```

jimfulton: Thanks. I implemented the approach you suggested.

google/cloud/bigquery/dbapi/_helpers.py
tswast: > SqlParameterScalarTypes and _SQL_SCALAR_TYPES lack DECIMAL and BIGDECIMAL.
https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
> Should I add DECIMAL and BIGDECIMAL?

Good catch! Filed https://github.com/googleapis/python-bigquery/issues/633 so that we can follow-up and add those.

> SqlParameterScalarTypes includes some aliases.
> Is that something we want in this context? (If so, I'll lobby for INT. ;))

Those aliases were intended to help folks who first learned BigQuery with "legacy" syntax. https://cloud.google.com/bigquery/data-types I think it's worth allowing the same aliases here in the DB-API.

tswast: On second-thought, I don't know if we want to be adding `DECIMAL` and `BIGDECIMAL` as aliases. Let's leave them out of this implementation for now, and we can decide later if we want more aliases beyond those we have for Legacy SQL support.

jimfulton: But they're documented in the _standard_ sql docs.

tests/unit/test_dbapi_cursor.py
```diff
@@ -734,3 +747,80 @@ def test__format_operation_unescaped_percent_w_list_param(self):
...
+        (
+            "values(%(foo:INT64)s, %(bar)s)",
+            ("values(%(foo)s, %(bar)s)", dict(foo="INT64")),
+        ),
+        (
+            "values('%%(oof:INT64)s, %(foo:INT64)s, %(foo)s)",
+            ("values('%%(oof:INT64)s, %(foo)s, %(foo)s)", dict(foo="INT64")),
+        ),
+        (
+            "values(%(foo:INT64)s, %(foo:INT64)s)",
```
tswast, line 781: Can we add some tests that check our aliases, too? (INTEGER -> INT64, for example)

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -20,15 +20,34 @@
...
 
-def scalar_to_query_parameter(value, name=None):
+def _parameter_type(name, value, query_parameter_type=None, value_doc=""):
+    if query_parameter_type:
+        parameter_type = query_parameter_type.upper()
+        if parameter_type not in _BIGQUERY_SCALAR_TYPES:
+            raise exceptions.ProgrammingError(
+                f"The given parameter type, {query_parameter_type},"
+                f" for {name} is not a valid BigQuery scalar type."
+            )
```
tswast, line 39: We might want to do something like this to make sure our aliases work as expected:

```suggestion
            )
            parameter_type = getattr(enums.SqlParameterScalarTypes, parameter_type)._type
```

tests/unit/test_dbapi_cursor.py
tswast: I think there is a backend alias for `INTEGER`, so it's not strictly necessary, but will be if we want to support aliases like `DECIMAL` in the future.

tswast (approved):
Looks good!

One optional suggestion, but I'm okay with solving the type aliases feature at a future date (if ever)

tests/unit/test_dbapi_cursor.py
jimfulton: That's included in my latest push.

We do need it  if we want to do client-side checks.

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: Oops, yup.

jimfulton: Pushed.

tswast (approved):

plamut (approved):
Looks good, although I mostly focused on the params replacement logic and only glanced over the rest (but didn't see anything obvious to fix).

PR #644, opened 2021-05-03
feat: add with_name() to ScalarQueryParameterType

### LINKED ISSUE #642: feat: add builder / copy method to `google.cloud.bigquery.query.ScalarQueryParameterType` that makes a copy with a new name

When constructing a [StructQueryParameterType](https://github.com/googleapis/python-bigquery/blob/6a48e80bc7d347f381b181f4cf81fef105d0ad0d/google/cloud/bigquery/query.py#L195), it is necessary to pass in types for sub-fields. We have created a [SqlParameterScalarTypes](https://github.com/googleapis/python-bigquery/blob/6a48e80bc7d347f381b181f4cf81fef105d0ad0d/google/cloud/bigquery/enums.py#L219-L238) enum to make this easier, but often times a struct's sub-fields are named. In this case, the enum values are less useful.

It'd be helpful to construct a type for the struct sub-fields by using the enum classes. Perhaps

```
google.cloud.bigquery.enums.SqlParameterScalarTypes.STRING.with_name("first_name")
```

could create a copy of `google.cloud.bigquery.enums.SqlParameterScalarTypes.STRING`, but with the name field populated?

Possible that the `SqlParameterScalarTypes` enum was unnecessary, but it's part of our public interface now.

 **Additional context**

https://github.com/googleapis/python-bigquery/issues/633#issuecomment-830170608

plamut:
Fixes #642.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/query.py
+++ b/google/cloud/bigquery/query.py
@@ -16,7 +16,7 @@
 
 from collections import OrderedDict
 import copy
-from typing import Optional
+from typing import Union
 
 from google.cloud.bigquery.table import _parse_schema_resource
 from google.cloud.bigquery._helpers import _rows_from_json
@@ -120,11 +120,13 @@ class ScalarQueryParameterType(_AbstractQueryParameterType):
         # attributes in the API representation when needed. Here we omit them.
         return {"type": self._type}
 
-    def with_name(self, new_name: Optional[str]):
+    def with_name(self, new_name: Union[str, None]):
         """Return a copy of the instance with ``name`` set to ``new_name``.
 
         Args:
-            name (Optional[str]): The new name of the query parameter type.
+            name (Union[str, None]):
+                The new name of the query parameter type. If ``None``, the existing
+                name is cleared.
 
         Returns:
             google.cloud.bigquery.query.ScalarQueryParameterType:
--- a/google/cloud/bigquery/query.py
+++ b/google/cloud/bigquery/query.py
@@ -16,6 +16,7 @@
 
 from collections import OrderedDict
 import copy
+from typing import Optional
 
 from google.cloud.bigquery.table import _parse_schema_resource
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/query.py
```diff
@@ -119,6 +120,18 @@ def to_api_repr(self):
         # attributes in the API representation when needed. Here we omit them.
         return {"type": self._type}
 
+    def with_name(self, new_name: Optional[str]):
```
jimfulton, line 123: Why make name optional?

jimfulton (commented):
I don't think name should be optional. :)

Also, omitting the name is untested, but easily fixed by making it required.

google/cloud/bigquery/query.py
plamut: We allow explicit `None`, the same as the constructor does (to allow clearing the name). Perhaps using the alternative form `Union[str, None]` would make it less confusing, as the argument itself is not optional.

jimfulton: You're right that the argument is required.  But I still don't see a point in allowing `None`.  Nameless types are readily available in `SqlParameterScalarTypes`.  To expect someone to call `with_name` and, you know, not provide a name seems silly. :)  

I don't think changing `Optional[str]` to `Union[str, None]` is an improvement.

I'm willing to defer.

jimfulton (approved):
Reluctantly.

google/cloud/bigquery/query.py
plamut: I tend to not assume about the users' use cases too much, as several times in the past we have gotten that wrong. :)

I don't feel strongly about either way (supporting or not supporting clearing the name), but since the maintenance burden is more or less the same for both options, supporting `None` seems fine here.

PR #656, opened 2021-05-07
chore: avoid pytype error caused by attrs==21.1.0

### LINKED ISSUE #655: The release attrs==21.1.0 broke pytype check
Running the `pytype` check on `master` fails:
```
$ nox -s pytype
```

```
...
    return attr.evolve(self, *args, **kwargs)
  File "/home/peter/workspace/python-bigquery/.nox/pytype/lib/python3.8/site-packages/attr/_funcs.py", line 347, in evolve
    changes[init_name] = evolve(value, **changes[init_name])
TypeError: evolve() argument after ** must be a mapping, not NamedType
...
```

This happens with the most recent `attrs` release, namely [21.1.0](https://github.com/python-attrs/attrs/releases/tag/21.1.0). The error does not happen in a release before that ([attrs==20.3.0](https://github.com/python-attrs/attrs/releases/tag/20.3.0)).

A possible workaround is to temporarily force installing that version before running `pytype`.

plamut:
**PR checklist:**

Fixes #655.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -103,6 +103,10 @@ def unit_noextras(session):
 @nox.session(python=DEFAULT_PYTHON_VERSION)
 def pytype(session):
     """Run type checks."""
+    # An indirect dependecy attrs==21.1.0 breaks the check, and installing a less
+    # recent version avoids the error until a possibly better fix is found.
+    # https://github.com/googleapis/python-bigquery/issues/655
+    session.install("attrs==20.3.0")
     session.install("-e", ".[all]")
     session.install("ipython")
     session.install(PYTYPE_VERSION)


noxfile.py
```diff
@@ -103,6 +103,10 @@ def unit_noextras(session):
 @nox.session(python=DEFAULT_PYTHON_VERSION)
 def pytype(session):
     """Run type checks."""
+    # An indirect dependecy attrs==21.1.0 breaks the check, and installing a less
+    # recent version avoids the error until a possibly better fix is found.
+    # https://github.com/googleapis/python-bigquery/issues/655
+    session.install("attrs==20.3.0")
```
jimfulton, line 109: Shouldn't this also be set in setup.py? Presumably with `<=`?

jimfulton: OK. I'd love to get a knowledge dump from you on `pytype`. :)

AFAIK, our code doesn't use `attr` directly.

It seems like you should be able to just tell `pytype` to ignore 3rd-party modules/packages, but I don' see a way to do that.  Apparently, `pytype` can use a configuration file, but they aren't documented AFAICT.

It seems it should be possible to exclude `attr` using `-x`, but I haven't figure out how. 

`pytype` is barely documented afaict.

jimfulton (approved):

noxfile.py
plamut: Yeah, that would also be my preference to silence that false positive without monkey-patching the dependency or by silencing all 3rd party modules. I think I'll create an issue to re-visit this workaround in the future.

PR #666, opened 2021-05-17
feat: detect obsolete BQ Storage extra at runtime

### LINKED ISSUE #629: support google-cloud-bigquery-storage v1

I'm filing this in response to internal feedback 186096865

```
Error: module 'google.cloud.bigquery_storage_v1.types' has no attribute 'DataFormat'
```

I believe this is because the customer is using a 1.x version of `google-cloud-bigquery-storage`.

(A) Ideally we'd have some code to support either 1.x or 2.x versions of `google-cloud-bigquery-storage`

* (B) Detect when 1.x is used and raise an appropriate exception asking to upgrade to 2.x
* (C) Detect ... and fallback to tabledata.list / getQueryResults + warn that the package should be upgraded

 **Additional context**

N/A

plamut:
Closes #629.

This PR adds logic to detect obsolete versions of BQ Storage extra and emit a warning or raise an error where applicable.

Traced all BQ Storage-related code and call paths and I think the changes here cover them all. As a bonus, we get a graceful fallback to `tabledata.list` in most cases (a warning is emitted instead of raising an error).

### DIFF
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -42,7 +42,6 @@ else:
     _ARROW_COMPRESSION_SUPPORT = True
 
 from google.cloud.bigquery import schema
-from google.cloud.bigquery._helpers import _verify_bq_storage_version
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -628,8 +627,6 @@ def _download_table_bqstorage(
     # Passing a BQ Storage client in implies that the BigQuery Storage library
     # is available and can be imported.
     from google.cloud import bigquery_storage
-
-    _verify_bq_storage_version()
 
     if "$" in table.table_id:
         raise ValueError(
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -42,6 +42,7 @@ else:
     _ARROW_COMPRESSION_SUPPORT = True
 
 from google.cloud.bigquery import schema
+from google.cloud.bigquery._helpers import _verify_bq_storage_version
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -627,6 +628,8 @@ def _download_table_bqstorage(
     # Passing a BQ Storage client in implies that the BigQuery Storage library
     # is available and can be imported.
     from google.cloud import bigquery_storage
+
+    _verify_bq_storage_version()
 
     if "$" in table.table_id:
         raise ValueError(
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -54,12 +54,14 @@ from google.cloud.bigquery._helpers import _del_sub_prop
 from google.cloud.bigquery._helpers import _get_sub_prop
 from google.cloud.bigquery._helpers import _record_field_to_json
 from google.cloud.bigquery._helpers import _str_or_none
+from google.cloud.bigquery._helpers import _verify_bq_storage_version
 from google.cloud.bigquery._helpers import _verify_job_config_type
 from google.cloud.bigquery._http import Connection
 from google.cloud.bigquery import _pandas_helpers
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetListItem
 from google.cloud.bigquery.dataset import DatasetReference
+from google.cloud.bigquery.exceptions import LegacyBigQueryStorageError
 from google.cloud.bigquery.opentelemetry_tracing import create_span
 from google.cloud.bigquery import job
 from google.cloud.bigquery.job import (
@@ -462,6 +464,12 @@ class Client(ClientWithProject):
                 "Cannot create BigQuery Storage client, the dependency "
                 "google-cloud-bigquery-storage is not installed."
             )
+            return None
+
+        try:
+            _verify_bq_storage_version()
+        except LegacyBigQueryStorageError as exc:
+            warnings.warn(str(exc))
             return None
 
         return bigquery_storage.BigQueryReadClient(credentials=self._credentials)
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -50,6 +50,13 @@ import google.cloud._helpers
 from google.cloud import exceptions  # pytype: disable=import-error
 from google.cloud.client import ClientWithProject  # pytype: disable=import-error
 
+try:
+    from google.cloud.bigquery_storage_v1.services.big_query_read.client import (
+        DEFAULT_CLIENT_INFO as DEFAULT_BQSTORAGE_CLIENT_INFO,
+    )
+except ImportError:
+    DEFAULT_BQSTORAGE_CLIENT_INFO = None
+
 from google.cloud.bigquery._helpers import _del_sub_prop
 from google.cloud.bigquery._helpers import _get_sub_prop
 from google.cloud.bigquery._helpers import _record_field_to_json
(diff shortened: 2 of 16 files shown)


plamut:
Will fix the coverage error son, one line not hit by the tests.

google/cloud/bigquery/client.py
```diff
@@ -464,6 +466,12 @@ def _create_bqstorage_client(self):
             )
             return None
 
+        try:
+            _verify_bq_storage_version()
```
jimfulton, line 470: Why not do this check in `Connection.__init__` if the `bqstorage_client` is not None after calling this?

You could even set it to None if the check fails.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -280,6 +281,8 @@ def _bqstorage_fetch(self, bqstorage_client):
         # bigquery_storage can indeed be imported here without errors.
         from google.cloud import bigquery_storage
 
+        _verify_bq_storage_version()
```
jimfulton, line 284: If we do the check in `Connection.__init__`, why do we need to do it here?  Are we afraid that someone would set that attribute?

Why is it fatal here but not in `Connection.__init__`?

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -628,6 +629,8 @@ def _download_table_bqstorage(
     # is available and can be imported.
     from google.cloud import bigquery_storage
 
+    _verify_bq_storage_version()
```
jimfulton, line 632: Why is this fatal here, but a warning elsewhere?

google/cloud/bigquery/client.py
jimfulton: I see this is also called in `table.py`, so you have the check in one place.  Maybe pass the given value here and do the create-if-none check here and return None if the check fails even if a non-None value is passed in.

google/cloud/bigquery/magics/magics.py
```diff
@@ -777,6 +777,12 @@ def _make_bqstorage_client(use_bqstorage_api, credentials, client_options):
         )
         raise customized_error from err
 
+    try:
+        bigquery._helpers._verify_bq_storage_version()
+    except bigquery.LegacyBigQueryStorageError as exc:
+        warnings.warn(str(exc))
+        return None
+
```
jimfulton, line 785: Would it make sense to replace this and the creation logic on line 794 with a call to `_create_bqstorage_client`?

jimfulton (commented):
I think we can be a little DRYer if we reuse `_create_bqstorage_client` more after adding some arguments to it.

google/cloud/bigquery/dbapi/cursor.py
plamut: Somebody could create a `Cursor` instance directly, i.e. without going through `Connection.cursor()`:

```py
conn = dbapi.Connection()
cursor = dbapi.Cursor(conn)
```

Not saying this is necessarily a recommended pattern, but `Cursor` is directly exposed in `google.cloud.bigquery.dbapi` after all, thus has its own check.

google/cloud/bigquery/magics/magics.py
plamut: This magics module already contains a bit too much cruft, thus removing some sounds good. :+1:

google/cloud/bigquery/_pandas_helpers.py
plamut: Similar reasoning as below, the helper's job is to download data specifically using the BQ Storage API. If it is called with an obsolete `bqstorage_client`, the actual error most likely is somewhere higher up.

Now, the check _might_ be unnecessary if the higher level code is water proof, but that might not be the case.

google/cloud/bigquery/client.py
plamut: Sorry, not following, you mean `Connection.__init__` from `dbapi`? That is not necessarily used here, the client methods can hit this path without using `dbapi`.

plamut: > I see this is also called in `table.py`, so you have the check in one place. Maybe pass the given value here and do the create-if-none check here and return `None` if the check fails even if a non-None value is passed in.

I'll have another look at the convoluted call graph that I drew. :) Might actually work, I'll consider it.

plamut:
> I think we can be a little DRYer if we reuse _create_bqstorage_client more after adding some arguments to it.

Sounds reasonable, I'll probably give it a shot. :+1:

google/cloud/bigquery/dbapi/cursor.py
jimfulton: > Somebody could create a `Cursor` instance directly, i.e. without going through `Connection.cursor()`:
> 
> ```python
> conn = dbapi.Connection()
> cursor = dbapi.Cursor(conn)
> ```

Sure, but the `Connection` constructor will still do the check.  The cursor will still get the storage client from the connection.

I may be misunderstanding you. :)

plamut: Or maybe it's too late here already. :)
Yeah, you're right, `Cursor` takes the client that's stored on the `Connection` instance, will simplify. And I even think there's still one non-graceful code path in `dbpai` (read; no fallback on the REST client), will double check tomorrow.

jimfulton:
I'm proposing that `_create_bqstorage_client` becomes something like:

```
    def _create_bqstorage_client(self, bqstorage_client=None, client_info=None, client_options=None):
        """Create a BigQuery Storage API client using this client's credentials.

        If a client cannot be created due to missing dependencies, raise a
        warning and return ``None``.

        Returns:
            Optional[google.cloud.bigquery_storage.BigQueryReadClient]:
                A BigQuery Storage API client.
        """
        try:
        except ImportError:
            warnings.warn(
                "Cannot create BigQuery Storage client, the dependency "
                "google-cloud-bigquery-storage is not installed."
            )
            return None

        try:
            _verify_bq_storage_version()
        except LegacyBigQueryStorageError as exc:
            warnings.warn(str(exc))
            return None

        if bqstorage_client is None:
            bqstorage_client = bigquery_storage.BigQueryReadClient(
                credentials=self._credentials,
                client_info=client_info,
                client_options=client_options,
                )

        return bqstorage_client
```

and maybe gets renamed :) and that it's called _even_ when the caller is given a client.

If the library is out of date, the caller will end up with `None` even if given a value.

google/cloud/bigquery/dbapi/cursor.py
jimfulton: > > Why is it fatal here but not in `Connection.__init__`?
> 
> This method specifically tries to fetch data over the BQ Storage API, and if it is given an outdated client, it's the caller's fault somewhere higher up the call chain. The method can't guess how to recover, thus at least throws a more informative exception.
> 
> (otherwise an `AttributeError` would be raised a few lines below at `bigquery_storage.types.DataFormat.ARROW`)

See my straw man I just posted for `_create_bqstorage_client`.  If all `bqstorage_client` assignments go through that, then you can never end up with an invalid client and thus never get the `AttributeError` mentioned above.

From your other comments, I think you can see where I'm going. :)

plamut:
I checked the call chains and if I didn't miss anything, the only place left where a bad BQ Storage client could leak down is `dbapi.Connection`, which I sealed. Some of the lower-level checks turned out to be redundant, thus removed them:
- In `dbapi.Cursor._bqstorage_fetch()`
- In `_pandas_helpers._download_table_bqstorage()`

Refactored `Client._make_bqstorage_client` as suggested and de-duplicated logic in `magics` and `dbapi`.

Apart from the client's factory, the other central place where we check if BQ Storage client is compatible is `table._validate_bqstorage()` (and issues a warning if it isn't). The top level `table` methods use this check and set `bqstorage_client` to `None`, if necessary, before sending it down the call chain to lower level utility functions.

google/cloud/bigquery/magics/magics.py
```diff
@@ -762,12 +762,12 @@ def _split_args_line(line):
     return params_option_value, rest_of_args
 
 
-def _make_bqstorage_client(use_bqstorage_api, credentials, client_options):
+def _make_bqstorage_client(client, use_bqstorage_api, client_options):
```
jimfulton, line 765: Why the credentials drop?

jimfulton (approved):
LGTM

PR #672, opened 2021-05-20
fix(tests): invalid path to strptime()

### LINKED ISSUE #671: AttributeError in `resource_name_to_date()` samples fixture
See this samples [test run](https://source.cloud.google.com/results/invocations/e5c424d5-84a6-4505-ae44-3bc70fa94e44/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fsamples%2Fpython3.6%2Fpresubmit/log). A bug in [this fixture](https://github.com/googleapis/python-bigquery/blob/bd7dbdae5c972b16bafc53c67911eeaa3255a880/samples/snippets/conftest.py#L33-L36) made all of them fail.

> AttributeError: module 'datetime' has no attribute 'strptime'

I wonder how this got in in the first place?

plamut:
Fixes #671.

**PR checklist:**

### DIFF
--- a/samples/snippets/conftest.py
+++ b/samples/snippets/conftest.py
@@ -33,7 +33,7 @@ def resource_prefix() -> str:
 def resource_name_to_date(resource_name: str):
     start_date = len(RESOURCE_PREFIX) + 1
     date_string = resource_name[start_date : start_date + RESOURCE_DATE_LENGTH]
-    return datetime.strptime(date_string, RESOURCE_DATE_FORMAT)
+    return datetime.datetime.strptime(date_string, RESOURCE_DATE_FORMAT)
 
 
 @pytest.fixture(scope="session", autouse=True)


jimfulton (approved):
I was wondering why we didn't hit this before, but I guess this method is only called during cleanup that isn't usually needed.

Maybe we should write a test for this.

kurtisvg (approved):
The periodic sample tests for this repo have had 37 straight days. So it's probably not (just) a glitch in the pipeline, it's that no one is watching the results.

plamut:
> So it's probably not (just) a glitch in the pipeline, it's that no one is watching the results.

Where can those be seen, or is that an internal thing?
(as an external maintainer I can only the Kokoro checks here on GitHub)

PR #673, opened 2021-05-20
feat: Support parameterized NUMERIC, BIGNUMERIC, STRING, and BYTES types

### LINKED ISSUE #668: Parameterized Types
Now:

    Strings and Bytes now have `maxLength`.
    Numeric(/DECIMAL) now have `precision` and `scale`.

These should be reflected in:

- `SchemaField`
- deserialization
- serialization? 
- anything else?

jimfulton:

Fixes #668 🦕

Some notes:

- I verified that the server:
  - enforces STRING maxLen and NUMERIC precision.
  - Enforces scale by truncating values.
- `select *` loses type-parameter information. :( but `client.list_rows` doesn't.
- I was thinking of trying to make sure that the Decimal objects created in Python matched the NUMERIC parameters, however
  - Because `SELECT` loses type information, it's mostly unavailable. IDK how often people use `list_rows`.
  - Because BQ sends data to the client as strings, this mostly just works, however:
    - The strings sent to the client lack trailing zeros. So, for example, given a scale of `2` and a value `123`, 
      BQ sends `123` rather than `123.00`, so the Python Decimal doesn't have the right number of trailing digits.  
     This is easy to fix, but incurs some performance overhead.  IDK if we want to do this. If we do, we can do it 
     in a follow-on PR.


google/cloud/bigquery/schema.py
```diff
@@ -67,6 +67,15 @@ class SchemaField(object):
 
         policy_tags (Optional[PolicyTagList]): The policy tag list for the field.
 
+        precision (Optional[int]):
+            Precison (number of digits) of fields with NUMERIC type.
```
plamut, line 71: (nit) Also for BIGNUMERIC, thus maybe reword to (BIG)NUMERIC?

google/cloud/bigquery/schema.py
```diff
@@ -77,6 +86,9 @@ def __init__(
         description=_DEFAULT_VALUE,
         fields=(),
         policy_tags=None,
+        precision=_DEFAULT_VALUE,
+        scale=_DEFAULT_VALUE,
+        maxLength=_DEFAULT_VALUE,
```
plamut, line 91: Didn't the liner complain here about a camelCase name?

google/cloud/bigquery/schema.py
```diff
@@ -86,9 +98,22 @@ def __init__(
...
+        if scale is not _DEFAULT_VALUE:
+            self._properties["scale"] = scale
+        if maxLength is not _DEFAULT_VALUE:
+            self._properties["maxLength"] = maxLength
         self._fields = tuple(fields)
         self._policy_tags = policy_tags
 
+    @staticmethod
+    def __get_int(f, name):
+        v = f.get(name, _DEFAULT_VALUE)
```
plamut, line 112: It might not be immediately clear what `f` means, let's to rename to `field`? Or `field_dict`?

I actually expected the linter to complain here ([naming style guide](https://google.github.io/styleguide/pyguide.html#s3.16-naming)), but I guess that was on a different project. :)

tests/system/test_client.py
```diff
@@ -2173,6 +2173,35 @@ def test_list_rows_page_size(self):
...
+            "create table {} ({})".format(
+                table_id, ", ".join(" ".join(f) for f in fields)
+            )
+        ).result()
+        table = client.get_table(table_id)
+        self.to_delete.insert(0, table)
+        table_id2 = table_id + "2"
+        client.create_table(Table(f"{client.project}.{table_id2}", table.schema))
+        table2 = client.get_table(table_id2)
+        self.to_delete.insert(0, table2)
```
plamut, line 2201: Just above this line, `client.create_table()` makes an API request, and if it fails `table2` will not be added to the `to_delete` list. It seems we need this one line up?

plamut (commented):
Looks good, just a few minor remarks that were supposed (?) to be caught by the linters.

The only more significant one is about the (non) cleanup in system tests.

google/cloud/bigquery/schema.py
jimfulton: Apparently not. :)

Which is a good thing, since I have no control over this name.  This is the parameter name defined by BQ.

plamut: Yeah, the field names in API response use camel case, thus there exist several `_PROPERTY_TO_API_FIELD` mappings in various classes. These mappings are used to get obtain the actual key to use when looking up the properties in API dicts.

Unless I'm missing something, it should be possible to rename the `maxLength` kwarg to `max_length`.

google/cloud/bigquery/schema.py
jimfulton: done

google/cloud/bigquery/schema.py
jimfulton: done

tests/system/test_client.py
jimfulton: Done.

Note that the dataset is deleted, so this is fairly moot.

google/cloud/bigquery/schema.py
jimfulton: Interesting.  I'm worried about users reading about `maxLength` somewhere, but I'll change this.  It should only affect `from_api_repr` and `to_api_repr` (at least after I removed some duplicate logic).

tests/system/test_client.py
plamut: Right... so the worst it could happen is that a table existed for a bit longer than necessary. Tunnel vision I'd say.

plamut (approved):
LGTM!

PR #678, opened 2021-05-24
test: add column ACLs test with real policy tag

### LINKED ISSUE #614: minor: add integration test for column acls feature
Once the datacatalog v1 endpoint releases a version with support for the PolicyTagManager client, let's add an integration test to better exercise the column ACL feature.

[This commit](https://github.com/googleapis/googleapis/commit/91eee3d039fbdbadee008393504900287bbc6f43) to proto comments should unblock the datacatalog release.

plamut:
Closes #614.

This is the test, PTAL. I could not test it locally, though, the backend still responds with UNIMPLEMENTED error. Am I missing anything?

**PR checklist:**

### DIFF
--- a/tests/system/test_client.py
+++ b/tests/system/test_client.py
@@ -383,9 +383,9 @@ class TestBigQuery(unittest.TestCase):
 
     def test_create_table_with_real_custom_policy(self):
         from google.cloud.bigquery.schema import PolicyTagList
-        from google.cloud.datacatalog_v1beta1 import PolicyTagManagerClient
-        from google.cloud.datacatalog_v1beta1.types import PolicyTag
-        from google.cloud.datacatalog_v1beta1.types import Taxonomy
+        from google.cloud.datacatalog_v1 import PolicyTagManagerClient
+        from google.cloud.datacatalog_v1.types import PolicyTag
+        from google.cloud.datacatalog_v1.types import Taxonomy
 
         policy_tag_client = PolicyTagManagerClient()
 
--- a/tests/system/test_client.py
+++ b/tests/system/test_client.py
@@ -68,6 +68,7 @@ from google.cloud.bigquery.table import Table
 from google.cloud._helpers import UTC
 from google.cloud.bigquery import dbapi, enums
 from google.cloud import storage
+from google.cloud.datacatalog_v1 import types as datacatalog_types
 
 from test_utils.retry import RetryErrors
 from test_utils.retry import RetryInstanceState
@@ -183,6 +184,10 @@ class TestBigQuery(unittest.TestCase):
                 retry_in_use(Config.CLIENT.delete_dataset)(doomed, delete_contents=True)
             elif isinstance(doomed, (Table, bigquery.TableReference)):
                 retry_in_use(Config.CLIENT.delete_table)(doomed)
+            elif isinstance(
+                doomed, (datacatalog_types.PolicyTag, datacatalog_types.Taxonomy)
+            ):
+                pass  # TODO: add logic
             else:
                 doomed.delete()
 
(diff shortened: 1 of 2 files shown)


tseaver:
Does the policy tag feature require adding the CI project to a list of allowed projects for the feature?

The second failure (in `TestBigQuery.test_dbapi_connection_does_not_leak_sockets`) is actually a "negative" leak (3 connections open at the start, 1 open at the end).  I'm assuming GC cleaned up after the policy tag test failed, but while the leak test was running.

shollyman (approved):
Thanks for putting this together.

plamut:
> Does the policy tag feature require adding the CI project to a list of allowed projects for the feature?

If I understood correctly, then no, but apparently there's still some missing piece to make this work. @shollyman  ?

> The second failure (in `TestBigQuery.test_dbapi_connection_does_not_leak_sockets`) is actually a "negative" leak (3 connections open at the start, 1 open at the end). I'm assuming GC cleaned up after the policy tag test failed, but while the leak test was running.

It's the first time I'm seeing this, but this explanation sounds plausible. It's slightly worrying that the test is not completely isolated, but it to date it hasn't caused problems. One option would be to change the [assertion](https://github.com/googleapis/python-bigquery/blob/27e4d34a257e2110a3928931c7ea58b2607e85d0/tests/system/test_client.py#L1497) to "less or equal", although that could also be masking the tests interference.

I'd say we keep an eye on it, but don't try to fix it, unless the flakiness becomes problematic.

shollyman:
I seem to be able to use the client without issue.

```
$ pip freeze
cachetools==4.2.2
certifi==2021.5.30
chardet==4.0.0
google-api-core==1.30.0
google-auth==1.31.0
google-cloud-datacatalog==3.2.1
googleapis-common-protos==1.53.0
grpc-google-iam-v1==0.12.3
grpcio==1.38.0
idna==2.10
libcst==0.3.19
mypy-extensions==0.4.3
packaging==20.9
proto-plus==1.18.1
protobuf==3.17.3
pyasn1==0.4.8
pyasn1-modules==0.2.8
pyparsing==2.4.7
pytz==2021.1
PyYAML==5.4.1
requests==2.25.1
rsa==4.7.2
six==1.16.0
typing-extensions==3.10.0.0
typing-inspect==0.7.1
urllib3==1.26.5
```

sample code:
```
from google.cloud.datacatalog_v1 import PolicyTagManagerClient
from google.cloud.datacatalog_v1.types import Taxonomy

def main():

  client = PolicyTagManagerClient()


  new_taxonomy = Taxonomy(
                          display_name="Custom test taxonomy",
                          description="This taxonomy is ony used for a test.",
                          activated_policy_types=[Taxonomy.PolicyType.FINE_GRAINED_ACCESS_CONTROL])

  createresp = client.create_taxonomy(parent="projects/shollyman-demo-test/locations/us", taxonomy=new_taxonomy)

  resp = client.list_taxonomies(parent="projects/shollyman-demo-test/locations/us")

  for t in resp:
    print("{} -> {}".format(t.display_name, t.name))


if __name__ == "__main__":
  main()
```


invocation:
```
$ python catalogtest.py
Custom test taxonomy -> projects/shollyman-demo-test/locations/us/taxonomies/408338391262190617
Sample Taxonomy -> projects/shollyman-demo-test/locations/us/taxonomies/6484488580374782575
test2 -> projects/shollyman-demo-test/locations/us/taxonomies/6247855699646749172
testing -> projects/shollyman-demo-test/locations/us/taxonomies/3363827134841403621
```

plamut:
Hm... and yet it fails on CI (and locally). :confused:  I'll try to check where the env differences might be when I re-visit this.

shollyman:
and apparently I mangled Peter's previous comment somehow instead of quoting it in my own.

plamut:
I see the full project path is used for the `parent` argument in the working sample, i.e. `"projects/shollyman-demo-test/locations/us"`, while the test here uses just the project id.

If I change that to a full project path, I start getting a permission denied error, but that actually seems OK, as my test service account does not have any DataCatalog roles granted. After fixing that, things actually started working.

Is there any way to make the backend error message more informative?

tests/system/test_client.py
```diff
@@ -183,6 +184,10 @@ def _still_in_use(bad_request):
                 retry_in_use(Config.CLIENT.delete_dataset)(doomed, delete_contents=True)
             elif isinstance(doomed, (Table, bigquery.TableReference)):
                 retry_in_use(Config.CLIENT.delete_table)(doomed)
+            elif isinstance(
+                doomed, (datacatalog_types.PolicyTag, datacatalog_types.Taxonomy)
+            ):
+                pass  # TODO: add logic
```
plamut, line 190: Will add this remaining bit tomorrow, but this should work now.

shollyman (approved):
Thanks for sticking with this.

PR #698, opened 2021-06-14
feat: add max_results parameter to some of the QueryJob methods

### LINKED ISSUE #296: Add max_results option to QueryJob.to_dataframe and QueryJob.to_arrow methods

plamut:
Closes #296.

It is now possible to cap the number of result rows returned when invoking `to_dataframe()` or `to_arrow()` method on a `QueryJob` instance.

**PR checklist:**:

### DIFF
--- a/google/cloud/bigquery/_tqdm_helpers.py
+++ b/google/cloud/bigquery/_tqdm_helpers.py
@@ -17,7 +17,7 @@
 import concurrent.futures
 import time
 import typing
-from typing import Any, Optional
+from typing import Optional
 import warnings
 
 try:
@@ -27,6 +27,7 @@ except ImportError:  # pragma: NO COVER
 
 if typing.TYPE_CHECKING:  # pragma: NO COVER
     from google.cloud.bigquery import QueryJob
+    from google.cloud.bigquery.table import RowIterator
 
 _NO_TQDM_ERROR = (
     "A progress bar was requested, but there was an error loading the tqdm "
@@ -62,7 +63,7 @@ def wait_for_query(
     query_job: "QueryJob",
     progress_bar_type: Optional[str] = None,
     max_results: Optional[int] = None,
-) -> Any:
+) -> "RowIterator":
     """Return query result and display a progress bar while the query running, if tqdm is installed.
 
     Args:
--- a/google/cloud/bigquery/_tqdm_helpers.py
+++ b/google/cloud/bigquery/_tqdm_helpers.py
@@ -16,12 +16,19 @@
 
 import concurrent.futures
 import time
+import typing
+from typing import Optional, Union
 import warnings
 
 try:
     import tqdm
 except ImportError:  # pragma: NO COVER
     tqdm = None
+
+if typing.TYPE_CHECKING:  # pragma: NO COVER
+    from google.cloud.bigquery import QueryJob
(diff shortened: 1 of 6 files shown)


google/cloud/bigquery/_tqdm_helpers.py
```diff
@@ -53,16 +58,34 @@ def get_progress_bar(progress_bar_type, description, total, unit):
     return None
 
 
-def wait_for_query(query_job, progress_bar_type=None):
-    """Return query result and display a progress bar while the query running, if tqdm is installed."""
+def wait_for_query(
+    query_job: "QueryJob",
+    progress_bar_type: Optional[str] = None,
+    max_results: Optional[int] = None,
+) -> Any:
```
plamut, line 65: The return type annotation was set to `Any`, because when set to `Union["RowIterator", "_EmptyRowIterator"]`, the `pytype` check failed with the following error:

```
File "/tmpfs/src/github/python-bigquery/google/cloud/bigquery/_tqdm_helpers.py", line 63, in <module>: Function dict.__setitem__ was called with the wrong arguments [wrong-arg-types]
         Expected: (self: Dict[str, Any], ...)
  Actually passed: (self: Dict[str, Union[Type[Optional[int]], str]], ...)
```

This is a bit odd, because line 63 used to be the line where `wait_for_query()` was defined, i.e. `def wait_for_query(`. It's not clear to me what dicts `pytype` complained about. :confused:

tseaver: Maybe try blowing away the `.pytype/` directory?

plamut: Did that, same result. And the local error was consistent with what failed on Kokoro, too.

tseaver: Is there a base class for `RowIterator` / `_EmptyRowIterator` which could be used?  Ideally, the caller shouldn't care about the "empty" iterator:  it should just behave normally.

plamut: There isn't. `RowIterator` inherits from `HTTPIterator`, while `_EmptyRowIterator` directly subclasses object.

I agree that users shouldn't care, however. After all both are the same thing, in principle, except that one's result set is empty. I'll see if changing the `_EmptyRowIterator`'s base class works.

plamut: **Update:** Making `_EmptyRowIterator` a subclass of `RowIterator` (with some plumbing) seems to work, setting `RowIterator` as the return type now passes the check.

I'll see if there's anything else to override in the parent iterator class, but this approach seems promising.

plamut:
The `docfx` failure is unrelated, the word is that forcing a re-run should resolve it.

tseaver (approved):

dandhlee:
all is good!

PR #703, opened 2021-06-18
feat: enable unsetting policy tags on schema fields

### LINKED ISSUE #558: add ability to unset policy tags (and description?) in schema fields

I noticed in https://github.com/googleapis/python-bigquery/pull/557 that we don't send the policy tags property if none are set in SchemaField. I believe this tells the backend not to make any updates to the policy tags property. This means it's not possible to _unset_ policy tags from Python.

(Note: I haven't verified this, so some system tests / samples should be written for this use case)

Similar to what I did for `description` in https://github.com/googleapis/python-bigquery/pull/557, we should disambiguate between unset policy tags and empty / none for policy tags.

We could make the update fields property more fine-grained, such as `schema.policytags` in order to force a sub-property to be included. This feels overly complicated, though.

 **Additional context**

N/A

plamut:
Closes #558.

This PR enables unsetting policy tags on schema fields.

(FWIW, it it already possible to unset the description, this PR just includes that detail in one of the system tests)

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/schema.py
+++ b/google/cloud/bigquery/schema.py
@@ -15,6 +15,7 @@
 """Schemas for BigQuery tables / queries."""
 
 import collections
+from typing import Optional
 
 from google.cloud.bigquery_v2 import types
 
@@ -106,11 +107,25 @@ class SchemaField(object):
             self._properties["maxLength"] = max_length
         self._fields = tuple(fields)
 
-        if policy_tags is None:
-            is_record = field_type is not None and field_type.upper() in _STRUCT_TYPES
-            self._policy_tags = None if is_record else PolicyTagList()
-        else:
-            self._policy_tags = policy_tags
+        self._policy_tags = self._determine_policy_tags(field_type, policy_tags)
+
+    @staticmethod
+    def _determine_policy_tags(
+        field_type: str, given_policy_tags: Optional["PolicyTagList"]
+    ) -> Optional["PolicyTagList"]:
+        """Return the given policy tags, or their suitable representation if `None`.
+
+        Args:
+            field_type: The type of the schema field.
+            given_policy_tags: The policy tags to maybe ajdust.
+        """
+        if given_policy_tags is not None:
+            return given_policy_tags
+
+        if field_type is not None and field_type.upper() in _STRUCT_TYPES:
+            return None
+
+        return PolicyTagList()
 
     @staticmethod
     def __get_int(api_repr, name):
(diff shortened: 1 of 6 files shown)


tests/system/test_client.py
```diff
@@ -653,6 +653,56 @@ def test_update_table_schema(self):
...
+            mode=old_field.mode,
+            description=None,
+            fields=old_field.fields,
+            policy_tags=None,
+        )
+
+        table.schema = new_schema
+        updated_table = Config.CLIENT.update_table(table, ["schema"])
+
+        self.assertFalse(updated_table.schema[1].description)  # Empty string or None.
```
plamut, line 703: I noticed there's a difference between when if description is not set (or set to `None`), or when set to an empty string. Does this matter?

In any case, it is possible to override the existing description, thus no extra code changes necessary for this.

plamut:
After the first submission I noticed that non-leaf schema fields should not have policy tags set (even if empty), the server complains, thus adjustment had to be made. But I opted not to raise any errors if a RECORD `SchemaField` is instantiated with non-empty policy tags, as we generally avoid duplicating the backend's validation logic.

plamut:
All tests passed, except for the snippets (500 internal error).

google/cloud/bigquery/schema.py
```diff
@@ -105,7 +105,12 @@ def __init__(
         if max_length is not _DEFAULT_VALUE:
             self._properties["maxLength"] = max_length
         self._fields = tuple(fields)
-        self._policy_tags = policy_tags
+
+        if policy_tags is None:
+            is_record = field_type is not None and field_type.upper() in _STRUCT_TYPES
+            self._policy_tags = None if is_record else PolicyTagList()
+        else:
+            self._policy_tags = policy_tags
```
tseaver, line 113: Can we move this logic out to a helper function, rather than replicating it in `from_api_repr` below?

tseaver (changes requested):

tseaver (approved):

tswast:
I wish I had reviewed this.

This change caused https://github.com/googleapis/python-bigquery-pandas/issues/387 and possibly reverted the fix introduced in https://github.com/googleapis/python-bigquery/pull/557.

In load jobs especially, we want to omit the `policyTags` key if not explicitly set/explicitly unset.

tests/system/test_client.py
tswast: Yes, this matters. In fact, it's what requested in https://github.com/googleapis/python-bigquery/issues/558#issue-834938931

> Similar to what I did for description in #557, we should disambiguate between unset policy tags and empty / none for policy tags.

PR #715, opened 2021-06-22
chore: pin sphinx plugin version to working one

dandhlee:
There's an issue with the current `gcp-sphinx-docfx-yaml` version failing for the following text:

```
        """Union[int, None]: Expiration time in milliseconds for a partition.

        If :attr:`partition_expiration` is set and :attr:`type_` is
        not set, :attr:`type_` will default to
        :attr:`~google.cloud.bigquery.table.TimePartitioningType.DAY`.
        """
```
which gets converted to
```
Union[int, None]: Expiration time in milliseconds for a partition.


If <xref:partition_expiration> is set and <xref:type_> is
not set, <xref:type_> will default to
<xref:google.cloud.bigquery.table.TimePartitioningType.DAY>.
```
which fails on the plugin because it thinks `<xref:type_>` should be a token it needs to process. The docstring is not malformed, the plugin needs additional work.

I'll have a fix for this by the end of the day, but should you need to submit PRs then please merge this one to have yourself unblocked!

Unblocks #698 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -300,7 +300,7 @@ def docfx(session):
 
     session.install("-e", ".")
     session.install(
-        "sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml"
+        "sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml==0.2.0"
     )
 
     shutil.rmtree(os.path.join("docs", "_build"), ignore_errors=True)


plamut:
Thanks for this!

As mentioned in the chat, it's fine for now and we can wait for the proper fix. But if the latter gets significantly delayed for some reason, we'll use this one as a workaround.

dandhlee:
https://github.com/googleapis/sphinx-docfx-yaml/pull/55 is ready for review, once that's submitted and released it should be good! Hopefully it won't take too long.

jimfulton (approved):

PR #718, opened 2021-06-23
feat: Support passing struct data to the DB API

### LINKED ISSUE #717: Support bound struct parameters with explicit types in db api interface
Allow passing struct data to queries if explicit type information is provided in the placeholder.

For example, given a table created with:

```
create table mydb.mytable (person struct<name string, children array<struct<name string, bdate date>>>)
```
You can insert data with:

```
insert into mydb.mytable (person) values (%(v:struct<name string, children array<struct<name string, bdate date>>>)s)
```
passing a value like:

```
dict(name='par',
                children=[
                    dict(name='ch1', bdate=datetime.date(2021, 1, 1)),
                    dict(name='ch2', bdate=datetime.date(2021, 1, 2)),
                    ])
```
We have to pass the type explicitly because struct data are dictionaries and may not preserve ordering reliably.

Even though, dictionaries are now ordered, the ordering is likely to be fragile.

```
>>> d = dict()
>>> d['a'] = 1
>>> d['b'] = 2
>>> d
{'a': 1, 'b': 2}
>>> del d['a']
>>> d['a'] = 4
>>> d
{'b': 2, 'a': 4}
```
¯\_(ツ)_/¯

jimfulton:
- [ x] Appropriate docs were updated (if necessary)

Fixes #717  🦕

### DIFF
--- a/docs/dbapi.rst
+++ b/docs/dbapi.rst
@@ -37,7 +37,14 @@ colon, as in::
 
   insert into people (name, income) values (%(name:string)s, %(income:numeric)s)
 
-For unnamed parameters, use the named syntax with a type, but now
+For unnamed parameters, use the named syntax with a type, but no
 name, as in::
 
   insert into people (name, income) values (%(:string)s, %(:numeric)s)
+
+Providing type information is the *only* way to pass `struct` data::
+
+  cursor.execute(
+    "insert into points (point) values (%(:struct<x float64, y float64>)s)",
+    [{"x": 10, "y": 20}],
+    )
--- a/docs/dbapi.rst
+++ b/docs/dbapi.rst
@@ -25,7 +25,7 @@ and using named parameters::
 Providing explicit type information
 -----------------------------------
 
-BigQuery requires type information for parameters.  The The BigQuery
+BigQuery requires type information for parameters.  The BigQuery
 DB-API can usually determine parameter types for parameters based on
 provided values.  Sometimes, however, types can't be determined (for
 example when `None` is passed) or are determined incorrectly (for
--- a/google/cloud/bigquery/dbapi/_helpers.py
+++ b/google/cloud/bigquery/dbapi/_helpers.py
@@ -148,16 +148,35 @@ parse_struct_field = re.compile(
 ).match
 
 
-def split_struct_fields(fields):
+def _split_struct_fields(fields):
+    # Split a string of struct fields.  They're defined by commas, but
+    # we have to avoid splitting on commas interbal to fields.  For
+    # example:
+    # name string, children array<struct<name string, bdate date>>
+    #
+    # only has 2 top-level fields.
     fields = fields.split(",")
+    fields = list(reversed(fields)) # in the off chance that there are very many
     while fields:
-        field = fields.pop(0)
+        field = fields.pop()
         while fields and field.count("<") != field.count(">"):
-            field += "," + fields.pop(0)
+            field += "," + fields.pop()
         yield field
 
 
 def complex_query_parameter_type(name: typing.Optional[str], type_: str, base: str):
+    """Construct a parameter type (`StructQueryParameterType`) for a complex type
+
+    or a non-complex type that's part of a complex type.
+
+    Examples:
+
+    array<struct<x float64, y float64>>
+
+    struct<name string, children array<struct<name string, bdate date>>>
+
+    This is used for computing array types.
+    """
     type_ = type_.strip()
     if "<" not in type_:
         # Scalar
@@ -186,7 +205,7 @@ def complex_query_parameter_type(name: typing.Optional[str], type_: str, base: s
         )
     else:
         fields = []
-        for field_string in split_struct_fields(sub):
+        for field_string in _split_struct_fields(sub):
             field_string = field_string.strip()
             m = parse_struct_field(field_string)
             if not m:
@@ -206,6 +225,13 @@ def complex_query_parameter(
     Construct a query parameter for a complex type (array or struct record)
 
     or for a subtype, which may not be complex
+
+    Examples:
+
+    array<struct<x float64, y float64>>
+
+    struct<name string, children array<struct<name string, bdate date>>>
+
     """
     type_ = type_.strip()
     base = base or type_
@@ -251,7 +277,7 @@ def complex_query_parameter(
         if not isinstance(value, collections_abc.Mapping):
             raise exceptions.ProgrammingError(f"Non-mapping value for type {type_}")
         value_keys = set(value)
-        for field_string in split_struct_fields(sub):
+        for field_string in _split_struct_fields(sub):
             field_string = field_string.strip()
             m = parse_struct_field(field_string)
             if not m:
--- a/google/cloud/bigquery/dbapi/_helpers.py
+++ b/google/cloud/bigquery/dbapi/_helpers.py
@@ -157,9 +157,7 @@ def split_struct_fields(fields):
         yield field
 
 
-def complex_query_parameter_type(
-    name: typing.Optional[str], type_: str, base: str
-):
+def complex_query_parameter_type(name: typing.Optional[str], type_: str, base: str):
     type_ = type_.strip()
     if "<" not in type_:
         # Scalar
--- a/google/cloud/bigquery/dbapi/_helpers.py
+++ b/google/cloud/bigquery/dbapi/_helpers.py
@@ -18,9 +18,10 @@ import datetime
 import decimal
 import functools
 import numbers
+import re
 
 from google.cloud import bigquery
-from google.cloud.bigquery import table, enums
+from google.cloud.bigquery import table, enums, query
 from google.cloud.bigquery.dbapi import exceptions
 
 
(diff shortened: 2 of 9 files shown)


jimfulton:
This also, as a side effect makes the pandas system tests run faster by changing the scope of the `dataset_id` fixture to session.

jimfulton:
A non-trivial part of this is allowing type parameters, as in `string(10)`, which it then strips off.

This allows users to specify type parameters matching the table definitions.  But we have to strip them off because they aren't allowed past table creation.  I'm not sure the benefit of this is worth the extra complexity.

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -113,6 +114,135 @@ def array_to_query_parameter(value, name=None, query_parameter_type=None):
...
+    ([A-Z0-9<> ,]+)  # Field type
+    $""", re.VERBOSE | re.IGNORECASE).match
+
+
+def split_struct_fields(fields):
+    fields = fields.split(',')
+    while fields:
+        field = fields.pop(0)
+        while fields and field.count('<') != field.count('>'):
+            field += ',' + fields.pop(0)
```
plamut: How many fields can we normally expect? Enough for `pop(0)`, which is an `O(N)` operation, to be noticeable?

We might want to process a reversed list of of fields instead, and pop from the right end, or use a `deque`?

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -113,6 +114,135 @@ def array_to_query_parameter(value, name=None, query_parameter_type=None):
...
+    \s*
+    <([A-Z0-9<> ,]+)>      # Subtype(s)
+    \s*$
+    """,
+    re.IGNORECASE | re.VERBOSE,
+    ).match
+parse_struct_field = re.compile(
+    r"""
+    (?:(\w+)\s+)    # field name
+    ([A-Z0-9<> ,]+)  # Field type
```
plamut: This also matches invalid struct fields such as `name str<>ing` - is that fine and we rely on the higher-level logic to detect and handle these cases?

plamut:
Quick comment - it would be helpful to reason about the code if helpers had an example or two in their docstrings, explaining the format of a typical (sub)string they process. You know, just to give readers some context. :)

plamut (commented):
Just two quick comments for a start, will have another look next time with a fresh head. :)

(some examples in helpers' docstrings would also be appreciated)

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: The number of fields is the number of fields in a struct, which is usually small.

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: Yup. This is just used to pull out the name. The type is then parsed further by calling the same function recursively.

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: I added  `list(reversed(fields)).

jimfulton:
> Quick comment - it would be helpful to reason about the code if helpers had an example or two in their docstrings, explaining the format of a typical (sub)string they process. You know, just to give readers some context. :)

Good idea! Done.

google/cloud/bigquery/dbapi/cursor.py
```diff
@@ -483,7 +483,33 @@ def _format_operation(operation, parameters):
...
+              (?=[, >)])        # Must be followed by ,>) or space
+              [a-zA-Z0-9<>, ]*  # Optional non-paren chars
+            )*                # Can be zero or more of parens and following text
+          )             # end of type group
+        )?            # close type clause ":type"
+        \))?          # End of optional name and/or type
+
+        s             # End of replacement
+        """,
+        re.VERBOSE,
```
plamut, line 511: This makes it more understandable, thanks!

plamut (commented):
Spotted a few minor things, but otherwise this looks good!

See if we could reduce code duplication with a reasonable effort, but we can do it separately if it proves to be too complex for now.

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -113,6 +129,176 @@ def array_to_query_parameter(value, name=None, query_parameter_type=None):
...
+        raise exceptions.ProgrammingError(f"Invalid parameter type, {type_}")
+    tname, sub = m.group(1, 2)
+    tname = tname.upper()
+    sub = sub.strip()
+    if tname == "ARRAY":
+        return query.ArrayQueryParameterType(
+            complex_query_parameter_type(None, sub, base), name=name
+        )
+    else:
+        fields = []
```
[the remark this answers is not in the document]
jimfulton: The `go` linter complains about it. I don't agree. I prefer the `else` because it makes the decision logic more explicit.

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -129,7 +315,9 @@ def to_query_parameters_list(parameters, parameter_types):
     result = []
 
     for value, type_ in zip(parameters, parameter_types):
-        if isinstance(value, collections_abc.Mapping):
+        if type_ is not None and "<" in type_:
+            param = complex_query_parameter(None, value, type_)
+        elif isinstance(value, collections_abc.Mapping):
```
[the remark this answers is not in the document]
jimfulton: done

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -113,6 +129,176 @@ def array_to_query_parameter(value, name=None, query_parameter_type=None):
...
+                )
+            field_name, field_type = m.groups()
+            fields.append(complex_query_parameter_type(field_name, field_type, base))
+
+        return query.StructQueryParameterType(*fields, name=name)
+
+
+def complex_query_parameter(
+    name: typing.Optional[str], value, type_: str, base: typing.Optional[str] = None
+):
```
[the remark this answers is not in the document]
jimfulton: done

docs/dbapi.rst
```diff
@@ -37,7 +37,14 @@ colon, as in::
 
   insert into people (name, income) values (%(name:string)s, %(income:numeric)s)
 
-For unnamed parameters, use the named syntax with a type, but now
+For unnamed parameters, use the named syntax with a type, but no
```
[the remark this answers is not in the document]
jimfulton: done

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -302,6 +302,24 @@ def complex_query_parameter(
...
+def _dispatch_parameter(type_, value, name=None):
+    if type_ is not None and "<" in type_:
+        param = complex_query_parameter(name, value, type_)
+    elif isinstance(value, collections_abc.Mapping):
+        raise NotImplementedError(
+            f"STRUCT-like parameter values are not supported"
+            f"{' (parameter ' + name + ')' if name else ''},"
+            f" unless an explicit type is give in the parameter placeholder"
+            f" (e.g. '%({name if name else ''}:struct<...>)s')."
+            )
```
plamut: Much more informative, great!

google/cloud/bigquery/dbapi/_helpers.py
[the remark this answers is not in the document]
plamut: Fair enough. I prefer a more flat function with multiple sequential exit points, but I guess this boils down to personal preferences.

plamut (approved):
LGTM, thanks for the refactoring.

I'll wait with merging a bit, just in case @shollyman has something to add.

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -113,6 +129,197 @@ def array_to_query_parameter(value, name=None, query_parameter_type=None):
...
+    parse_struct_field=re.compile(
+        r"""
+        (?:(\w+)\s+)    # field name
+        ([A-Z0-9<> ,()]+)  # Field type
+        $""",
+        re.VERBOSE | re.IGNORECASE,
+    ).match,
+):
+    # Split a string of struct fields.  They're defined by commas, but
+    # we have to avoid splitting on commas interbal to fields.  For
```
tswast, line 144:
```suggestion
    # we have to avoid splitting on commas internal to fields.  For
```

google/cloud/bigquery/dbapi/_helpers.py
```diff
@@ -113,6 +129,197 @@ def array_to_query_parameter(value, name=None, query_parameter_type=None):
...
+        if name:
+            type_ = type_.with_name(name)
+        return SCALAR, type_
+
+    m = complex_query_parameter_parse(type_)
+    if not m:
+        raise exceptions.ProgrammingError(f"Invalid parameter type, {type_}")
+    tname, sub = m.group(1, 2)
+    if tname.upper() == "ARRAY":
+        sub_type = complex_query_parameter_type(None, sub, base)
```
tswast, line 204: This allows for arrays of scalars, right? Looks like it from looking at the code, but then I'm a bit confused why we need the scalar logic in lines 183-197.

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: Thanks. :)

google/cloud/bigquery/dbapi/_helpers.py
jimfulton: We need the code on line 183-197 _because_ we support arrays and structs of scalars.  Basically, we allow arrays of anything (other than arrays). To handle that, we recurse to handle whatever's inside the `<>`s, which might be a simple declaration like `INT64`.

plamut:
Last call - anybody else wants to add anything her? If not, I'll merge it soon. :slightly_smiling_face:

tswast (approved):

PR #734, opened 2021-07-01
feat: make it easier to disable best-effort deduplication with streaming inserts

### LINKED ISSUE #720: `Client.insert_rows_json()`: add option to disable best-effort deduplication
Currently, the `Client.insert_rows_json()` method for streaming inserts always inserts an `insertId` unique identifier for each row provided.
This row identifier can be user-provided; if the user doesn't provide any identifiers, the library automatically fills the row IDs by using UUID4.

Here's the code:
```python
        for index, row in enumerate(json_rows):
            info = {"json": row}
            if row_ids is not None:
                info["insertId"] = row_ids[index]
            else:
                info["insertId"] = str(uuid.uuid4())
            rows_info.append(info)
```

However, insert IDs are entirely optional, and there are actually valid use cases _not_ to use them. From the [BigQuery documentation](https://cloud.google.com/bigquery/streaming-data-into-bigquery#disabling_best_effort_de-duplication):

> You can disable best effort de-duplication by not populating the insertId field for each row inserted. When you do not populate insertId, you get higher streaming ingest quotas in certain regions. This is the recommended way to get higher streaming ingest quota limits.

The BigQuery Python client library provides no way of omitting the `insertId`s. it would be nice to have a parameter for that.

plamut:
Closes #720.

As a bonus, it is now also possible to provide any iterable (not just sequences) for `row_ids`, avoiding the need to create a possibly large temporary list/tuple. And even though we do accumulate all items, we still save the memory that would otherwise be occupied by the list itself.

**PR checklist:**


tswast (approved):
Looks great!

plamut:
Thanks!

PR #735, opened 2021-07-02
feat: add support for decimal target types

### LINKED ISSUE #726: Support decimalTargetTypes
This FR is for supporting decimalTargetTypes.  This feature gives users the option to import Avro/Parquet/ORC Decimal to GoogleSQL BIGNUMERIC type, in load jobs and inline/persistent external tables.

Additional supporting information to be shared with implementors separately.

plamut:
Closes #726.

This PR adds support for decimal target types.

(still need to add the same to external config)

**PR checklist:**:


google/cloud/bigquery/external_config.py
```diff
@@ -693,6 +694,28 @@ def compression(self):
...
+        """Possible SQL data types to which the source decimal values are converted.
+
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#ExternalDataConfiguration.FIELDS.decimal_target_types
+
+        .. versionadded:: 2.21.0
+        """
+        prop = self._properties.get("decimalTargetTypes")
+        if prop is not None:
+            prop = frozenset(prop)
```
tseaver, line 708: The enum docstrings added above suggest that order is significant.  However, `frozenset` is unordered:  if the ordering is signifcant, then we should be preserving it -- maybe return a tuple to preserve the order?

plamut: As per [docs](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#ExternalDataConfiguration.FIELDS.decimal_target_types), the order is not significant, which why a (frozen) set is returned.

> The order of the types in this field is ignored.

I'll re-read the enum docstrings and make this more clear, if it's currently ambiguous.

tseaver: So we are storing it as a list in the setter for compatibility with how the back-end sends it to us?

plamut: It's primarily because sets are not JSON serializable out of the box - `json.dumps()` uses `JSONEncoder` by default, but the latter doesn't know how to serialize sets.

plamut: @shollyman The `DecimalTargetType` enum [docs](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#DecimalTargetType) can be misleading if not knowing the entire context. That _"Depending on the order"_ bit might be understood as if the order matters when providing a list of target decimal types, thus I removed it from the Python class docstrings.

Should the generated docs follow suit? Or perhaps modify the wording?

shollyman: Yeah, order absolutely matters here.  The feedback about the docs is good, but I'm going to punt that to @stephaniewang526 to drive.

stephaniewang526: Hi Peter, yes I can see how the doc is a little confusing... Order matters very much in this feature please.

plamut: "disco doc" - gotta love that lingo. :heart: :confetti_ball: :rofl:

stephaniewang526: Upon checking w myzhong@, the order actually does not matter here. It would always go for NUMERIC if it's possible even if user specified [\"BIGNUMERIC\", \"NUMERIC\"].

plamut: OK thanks, we'll keep the target types in a `frozenset` then. We should probably reword the generated enum docs, though.

tseaver (approved):

google/cloud/bigquery/external_config.py
stephaniewang526: b/193235792 created to track the doc update. cl/383854927 created to update the doc. Pending review from BQ techwriters.

google/cloud/bigquery/enums.py
```diff
@@ -49,6 +49,24 @@ class Compression(object):
     """Specifies no compression."""
 
 
+class DecimalTargetType:
```
tswast, line 52: I notice we are missing docs for a bunch of enums. I filed https://github.com/googleapis/python-bigquery/issues/744 to switch the docs over to module-based docs for this module.

tswast (approved):

PR #736, opened 2021-07-02
feat: add `LoadJobConfig.projection_fields` to select DATASTORE_BACKUP fields

tswast:
For https://issuetracker.google.com/191814927


google/cloud/bigquery/job/load.py
```diff
@@ -277,6 +277,27 @@ def null_marker(self):
...
+        error is returned in the job result.
+
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationLoad.FIELDS.projection_fields
+        """
+        return self._get_sub_prop("projectionFields")
+
+    @projection_fields.setter
+    def projection_fields(self, value):
+        self._set_sub_prop("projectionFields", value)
```
plamut, line 299: Sanity check - if `value` is `None`, should we omit the `projectionFields` key altogether? Or do we still want it to be present and explicitly set to `None`?

(I'm never quite sure when this makes a difference on the backend and when not)

google/cloud/bigquery/job/load.py
```diff
@@ -277,6 +277,27 @@ def null_marker(self):
     def null_marker(self, value):
         self._set_sub_prop("nullMarker", value)
 
+    @property
+    def projection_fields(self):
+        """Optional[List[str]]: If
```
plamut, line 282: (nit) Let's annotate the types with type annotation syntax, we started doing that for all new code.

plamut (commented):
Looks good, I just have two remarks/questions.

google/cloud/bigquery/job/load.py
tswast: I don't think it should make a difference in this case since jobs aren't mutable. In general, I prefer to send explicit `None` if we got one, since that allows things like "unsetting" values.

google/cloud/bigquery/job/load.py
tswast: Added! Also, TIL that pytype doesn't really work on macOS. I hit https://github.com/google/pytype/issues/661 when testing locally.

google/cloud/bigquery/job/load.py
plamut: Makes sense. :+1:

google/cloud/bigquery/job/load.py
```diff
@@ -278,7 +279,7 @@ def null_marker(self, value):
         self._set_sub_prop("nullMarker", value)
 
     @property
-    def projection_fields(self):
+    def projection_fields(self) -> Optional[List[str]]:
```
plamut, line 282: Thanks, but let's annotate the setter, too (the `value` param).

plamut (commented):
We should annotate the setter, too. And resolve a merge conflict now that a few other PRs have been merged.

**Edit:** Resolved it myself, it was trivial (merging the `typing` imports).

plamut (approved):

PR #740, opened 2021-07-08
feat: add support for table snapshots

### LINKED ISSUE #730: Implement Table Snapshots
Feature request for tracking implementation of Table Snapshots functionality in BigQuery client.  Table Snapshot feature allows customers to generate a persistent copy of an existing table as it were at a given point in time within the same project.

Additional information has been provided to implementors separately.

plamut:
Closes #730.

This PR adds support for handling table snapshots, as per specs.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1306,10 +1306,9 @@ class SnapshotDefinition:
 
         self.snapshot_time = None
         if "snapshotTime" in resource:
-            dt = google.cloud._helpers._rfc3339_to_datetime(resource["snapshotTime"])
-            # The helper returns a timezone *aware* datetime object (with UTC tzinfo),
-            # we need to make it naive.
-            self.snapshot_time = dt.replace(tzinfo=None)
+            self.snapshot_time = google.cloud._helpers._rfc3339_to_datetime(
+                resource["snapshotTime"]
+            )
 
 
 class Row(object):
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -321,6 +321,7 @@ class Table(object):
         "range_partitioning": "rangePartitioning",
         "time_partitioning": "timePartitioning",
         "schema": "schema",
+        "snapshot_definition": "snapshotDefinition",
         "streaming_buffer": "streamingBuffer",
         "self_link": "selfLink",
         "table_id": ["tableReference", "tableId"],
@@ -910,6 +911,19 @@ class Table(object):
             self._PROPERTY_TO_API_FIELD["external_data_configuration"]
         ] = api_repr
 
+    @property
+    def snapshot_definition(self) -> Optional["SnapshotDefinition"]:
+        """Information about the snapshot. This value is set via snapshot creation.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.snapshot_definition
+        """
+        snapshot_info = self._properties.get(
+            self._PROPERTY_TO_API_FIELD["snapshot_definition"]
+        )
+        if snapshot_info is not None:
+            snapshot_info = SnapshotDefinition(snapshot_info)
+        return snapshot_info
+
     @classmethod
     def from_string(cls, full_table_id: str) -> "Table":
         """Construct a table from fully-qualified table ID.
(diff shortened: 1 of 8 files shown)


google/cloud/bigquery/table.py
```diff
@@ -1274,6 +1288,30 @@ def __init__(self, resource):
...
+        if "baseTableReference" in resource:
+            self.base_table_reference = TableReference.from_api_repr(
+                resource["baseTableReference"]
+            )
+
+        self.snapshot_time = None
+        if "snapshotTime" in resource:
+            dt = google.cloud._helpers._rfc3339_to_datetime(resource["snapshotTime"])
+            # The helper returns a timezone *aware* datetime object (with UTC tzinfo),
+            # we need to make it naive.
```
tseaver, line 1311: Why would we prefer a naive time here?  If the back-end is returning a Zulu-point-in-time, what benefit is there in discarding that information?

tests/system/test_client.py
```diff
@@ -394,7 +394,7 @@ def test_create_table_with_real_custom_policy(self):
         taxonomy_parent = f"projects/{Config.CLIENT.project}/locations/us"
 
         new_taxonomy = datacatalog_types.Taxonomy(
-            display_name="Custom test taxonomy",
+            display_name=unique_resource_id("Custom test taxonomy"),
```
tseaver, line 397: Given that the back-end generates a unique ID, why mangle the human-readable version?

tests/unit/job/test_copy.py
```diff
@@ -70,6 +76,20 @@ def test_to_api_repr_with_encryption_none(self):
             resource, {"copy": {"destinationEncryptionConfiguration": None}}
         )
 
+    def test_operation_type_unspecified(self):
+        from google.cloud.bigquery.job import OperationType
+
+        config = self._make_one()
+        self.assertEqual(
+            config.operation_type, OperationType.OPERATION_TYPE_UNSPECIFIED
+        )
```
tseaver, line 85: This part seems like it belongs in a (not-yet-there) `test_ctor_defaults` testcase, along with assertions about other default properties.

tests/unit/job/test_copy.py
```diff
@@ -70,6 +76,20 @@ def test_to_api_repr_with_encryption_none(self):
...
+
+        # Setting it to None is the same as setting it to OPERATION_TYPE_UNSPECIFIED.
+        config.operation_type = None
+        self.assertEqual(
+            config.operation_type, OperationType.OPERATION_TYPE_UNSPECIFIED
+        )
```
tseaver, line 91: This test would be clearer if we constructed the instance initially with a non-default value, and then did the assignment to `None`.  We likely also need a testcase for the setter passing a non-`None` value.

tseaver (changes requested):

google/cloud/bigquery/table.py
plamut: If a timezone is not explicitly specified, the default time zone, UTC, is used for timestamps. But yeah, that's implicit, I'll convert it back to a tz-aware object.

tests/system/test_client.py
plamut: It turned out that this isn't the case, and `display_name` is used as an identifier - trying to create another taxonomy with the same display name results in a 409 Already Exists.

When such taxonomy was not cleaned up in one of the test runs (for some mysterious reason), the subsequent runs suddenly started failing on `master` (actually, on all branches).

tests/unit/job/test_copy.py
plamut: Makes sense, will make the tests more specific.

tests/system/test_client.py
tseaver: OK.  In that case, that is a pretty surpising use of `unique_resource_id`, which takes an argument, `delimiter` (not `prefix`).  It can in theory appear twice in the generated string.  A more usual idiom would be:
```suggestion
            display_name="Custom test taxonomy" + unique_resource_id(),
```

tseaver: (BTW, the back-end behavior is totally bogus).

plamut: Yeah, I was surprised too, did not expect the _display_name_ to affect the resource identity. :confused:

tseaver:
@googlebot  I fixed it.

tseaver:
@plamut Ugh, sorry about the CLA mess.  Just edit me out of the history.

plamut:
Oh dear, the CLA bot again :/

@tswast Can you please help?

tseaver (approved):

tswast (commented):
Looks good, thanks. Could you also add the new classes to https://github.com/googleapis/python-bigquery/blob/master/docs/reference.rst ?

tests/system/test_client.py
```diff
@@ -394,7 +394,7 @@ def test_create_table_with_real_custom_policy(self):
         taxonomy_parent = f"projects/{Config.CLIENT.project}/locations/us"
 
         new_taxonomy = datacatalog_types.Taxonomy(
-            display_name="Custom test taxonomy",
+            display_name="Custom test taxonomy" + unique_resource_id(),
```
shollyman, line 397: I'm guessing testing flaked?

shollyman (approved):

tests/system/test_client.py
plamut: One of the test runs did not clean after itself (for some unknown reason), meaning that all subsequent test runs started failing because of a static resource ID (display name in this case, which is kind of odd).

plamut:
> Looks good, thanks. Could you also add the new classes to https://github.com/googleapis/python-bigquery/blob/master/docs/reference.rst ?

I thought they will be added automatically? Will do it, thanks for catching that.

Issue #747, opened 2021-07-09
make `google-cloud-bigquery-storage>=2.6.0` the minimum version


tswast:
2.6.0 makes `read_session` optional to `rows()` (used in pandas connector).

BLOCKED: wait for a few versions of `google-cloud-bigquery-storage` before bumping the minimum.

(TODO: I still want to figure out how long it takes people to upgrade naturally so that we can set some data-driven policies around deps upgrades)

PR #748, opened 2021-07-09
refactor: omit `read_session` with latest google-cloud-bigquery-storage

### LINKED ISSUE #742: don't pass `read_session` to BQ Storage API `rows` from `to_dataframe` and `to_arrow`
https://github.com/googleapis/python-bigquery-storage/pull/228 makes `read_session` optional and quietly deprecates it.

If a new enough `google-cloud-bigquery-storage` client is installed, I'd like to avoid passing the `read_session` to `rows`, so that we can more loudly deprecate it (https://github.com/googleapis/python-bigquery-storage/issues/229)

tswast:
`read_session` is unnecessary as of `google-cloud-bigquery-storage>=2.6.0`.
This will allow us to more loudly deprecate the use of `rows(read_session)`.  https://github.com/googleapis/google-cloud-python/issues/14281

Rather than require 2.6.0, version switches will allow us to keep our
requirements range wider. Will want to give this version some time to bake
before making it the minimum version. https://github.com/googleapis/python-bigquery/issues/747

Fixes googleapis/python-bigquery-storage#742  🦕


google/cloud/bigquery/_helpers.py
```diff
@@ -67,6 +71,9 @@ def _verify_bq_storage_version():
         )
         raise LegacyBigQueryStorageError(msg)
 
+    if installed_version >= _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION:
+        _IS_BQ_STORAGE_READ_SESSION_OPTIONAL = True
```
plamut, line 75: This could potentially be tricky, because the value of the `_IS_BQ_STORAGE_READ_SESSION_OPTIONAL` "constant" depends on whether `_verify_bq_storage_version()` has been called or not.

It would IMO be more straightforward if we performed version detection at import time and already make the final decision on the value of `_IS_BQ_STORAGE_READ_SESSION_OPTIONAL` then.

tswast: We don't import `google.cloud.bigquery_storage` in `_download_table_bqstorage_stream`. Since that function gets a client and other objects passed in to it.

How about we make sure we call `_verify_bq_storage_version()` from `_download_table_bqstorage_stream()` and optimize to not call the version parsing too often?

plamut: The overhead of parsing a version is most likely negligible compared to the request/response times, I was more concerned that changing a module "constant" might represent a landmine in the "right" constellation of the planets.

Perhaps just adding a comment that this private "constant" is not a constant and can change when calling `_verify_bq_storage_version()` is enough. You know, just to warn the readers about it.

tswast: Good point. I think there's a way to do this without "constants" that aren't actually constant. I think maybe I'll create a little class to hold our version comparisons.

google/cloud/bigquery/_helpers.py
```diff
@@ -26,7 +26,7 @@
 from google.cloud._helpers import _RFC3339_MICROS
 from google.cloud._helpers import _RFC3339_NO_FRACTION
 from google.cloud._helpers import _to_bytes
-import pkg_resources
+import packaging.version
```
tswast, line 29: I wanted to add type annotations, but it turns out `pkg_resources` is just a thin wrapper around `packaging`.

https://github.com/pypa/setuptools/blob/a4dbe3457d89cf67ee3aa571fdb149e6eb544e88/pkg_resources/__init__.py/#L112

google/cloud/bigquery/_helpers.py
```diff
@@ -41,31 +42,53 @@
...
-_MIN_BQ_STORAGE_VERSION = pkg_resources.parse_version("2.0.0")
+_MIN_BQ_STORAGE_VERSION = packaging.version.Version("2.0.0")
+_BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION = packaging.version.Version("2.6.0")
 
 
-def _verify_bq_storage_version():
-    """Verify that a recent enough version of BigQuery Storage extra is installed.
+class BQStorageVersions:
+    def __init__(self):
+        self._installed_version = None
```
plamut, line 51: This version checker, sir, is classy (pun intended). Well done!

plamut (dismissed):
Looks good and clean!

Consider maybe adding a few unit tests directly testing the `BQStorageVersions` class, although it does seem that the class is tested indirectly through other tests (the coverage does not complain).

**Edit:** I mean, the coverage check does complain, but not about `_helpers.BQStorageVersions`.

google/cloud/bigquery/_helpers.py
```diff
@@ -41,31 +42,53 @@
...
+        return self.installed_version >= _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION
+
+    def verify_version(self):
+        """Verify that a recent enough version of BigQuery Storage extra is installed.
+
+        The function assumes that google-cloud-bigquery-storage extra is installed, and
+        should thus be used in places where this assumption holds.
+
+        Because `pip` can install an outdated version of this extra despite the constraints
+        in setup.py, the the calling code can use this helper to verify the version
```
plamut, line 77: Oops, I now noticed you copied my typo, let's remove it.

```suggestion
        in `setup.py`, the calling code can use this helper to verify the version
```

tswast:
Turns out the coverage error was due to my use of a `lambda` that never executed. I'll remove those and add a few tests for the underlying class in my next commit.

plamut (approved):

PR #750, opened 2021-07-11
fix: use pandas function to check for NaN

### LINKED ISSUE #729: Bug: `dataframe_to_json_generator` doesn't support `pandas.NA` type
#### Environment details

  - OS type and version: Ubuntu 20.04.2
  - Python version: `3.9.4`
  - pip version: `21.1.2`
  - `google-cloud-bigquery` version: `2.20.0`

#### Steps to reproduce

  1. Run the code example below

#### Code example

```python
import pandas

df = pandas.DataFrame({
    "series_a": [1, 2, pandas.NA]
})

json_iter = bigquery._pandas_helpers.dataframe_to_json_generator(df)
for row in json_iter:
    print(row)
```

#### Stack trace
```
{'series_a': 1}
{'series_a': 2}
  File "/home/christian/code/bug_example.py", line 11, in <module>
    for row in json_iter:
  File "/home/christian/miniconda3/envs/data-services-prod/lib/python3.9/site-packages/google/cloud/bigquery/_pandas_helpers.py", line 783, in dataframe_to_json_generator
    if value != value:
  File "pandas/_libs/missing.pyx", line 360, in pandas._libs.missing.NAType.__bool__
TypeError: boolean value of NA is ambiguous
```

#### Suggested fix

Starting with pandas 1.0, an experimental pandas.NA value (singleton) is available to represent scalar missing values as
opposed to numpy.nan. Comparing the variable with itself (`value != value`) results in a `TypeError` as the `pandas.NA` value [doesn't support type-casting to boolean](https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#na-in-a-boolean-context).

I am planning to make a PR that switches the syntax `value != value` on [_pandas_helpers.py#L783](https://github.com/googleapis/python-bigquery/blob/ddd9ce714ebc496e22bc122e89f3bbee36ea15be/google/cloud/bigquery/_pandas_helpers.py#L783) to use the `pandas.isna` function but wanted to check if there is a better solution before I submit a patch?

LinuxChristian:
Starting with pandas 1.0, an experimental `pandas.NA` value (singleton) is available to represent scalar missing values as
opposed to `numpy.nan`. Comparing the variable with itself results in a `pandas.NA` value that doesn't support type-casting
to boolean. Using the build-in `pandas.isna` function handles all pandas supported NaN values (`None`, `pd.NA`, `np.nan`, `NaT`).

Fixes #729

### DIFF
--- a/tests/unit/test__pandas_helpers.py
+++ b/tests/unit/test__pandas_helpers.py
@@ -19,6 +19,7 @@ import functools
 import operator
 import queue
 import warnings
+import pkg_resources
 
 import mock
 
@@ -46,6 +47,14 @@ try:
     from google.cloud import bigquery_storage
 except ImportError:  # pragma: NO COVER
     bigquery_storage = None
+
+PANDAS_MINIUM_VERSION = pkg_resources.parse_version("1.0.0")
+
+if pandas is not None:
+    PANDAS_INSTALLED_VERSION = pkg_resources.get_distribution("pandas").parsed_version
+else:
+    # Set to less than MIN version.
+    PANDAS_INSTALLED_VERSION = pkg_resources.parse_version("0.0.0")
 
 
 skip_if_no_bignumeric = pytest.mark.skipif(
@@ -734,7 +743,10 @@ def test_list_columns_and_indexes_with_named_index_same_as_column_name(
     assert columns_and_indexes == expected
 
 
-@pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
+@pytest.mark.skipIf(
+    pandas is None or PANDAS_INSTALLED_VERSION < PANDAS_MINIUM_VERSION,
+    reason="Requires `pandas version >= 1.0.0` which introduces pandas.NA",
+)
 def test_dataframe_to_json_generator(module_under_test):
     utcnow = datetime.datetime.utcnow()
     df_data = collections.OrderedDict(
--- a/tests/unit/test__pandas_helpers.py
+++ b/tests/unit/test__pandas_helpers.py
@@ -771,8 +771,7 @@ def test_dataframe_to_json_generator(module_under_test):
         {"a_series": 3, "b_series": 0.3, "d_series": utcnow, "e_series": True},
         {"a_series": 4, "b_series": 0.4, "c_series": "d"},
     ]
-    for row, expect in zip(rows, expected):
-        assert row == expect
+    assert list(rows) == expected
 
 
 @pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
--- a/tests/unit/test__pandas_helpers.py
+++ b/tests/unit/test__pandas_helpers.py
@@ -735,6 +735,35 @@ def test_list_columns_and_indexes_with_named_index_same_as_column_name(
 
 
 @pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
+def test_dataframe_to_json_generator(module_under_test):
+    utcnow = datetime.datetime.utcnow()
+    df_data = collections.OrderedDict(
+        [
+            ("a_series", [pandas.NA, 2, 3, 4]),
+            ("b_series", [0.1, float("NaN"), 0.3, 0.4]),
+            ("c_series", ["a", "b", pandas.NA, "d"]),
+            ("d_series", [utcnow, utcnow, utcnow, pandas.NaT]),
+            ("e_series", [True, False, True, None]),
+        ]
+    )
+    dataframe = pandas.DataFrame(
+        df_data, index=pandas.Index([4, 5, 6, 7], name="a_index")
+    )
+
+    dataframe = dataframe.astype({"a_series": pandas.Int64Dtype()})
+
+    rows = module_under_test.dataframe_to_json_generator(dataframe)
+    expected = [
+        {"b_series": 0.1, "c_series": "a", "d_series": utcnow, "e_series": True},
+        {"a_series": 2, "c_series": "b", "d_series": utcnow, "e_series": False},
+        {"a_series": 3, "b_series": 0.3, "d_series": utcnow, "e_series": True},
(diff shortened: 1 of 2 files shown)


tests/unit/test__pandas_helpers.py
```diff
@@ -734,6 +734,35 @@ def test_list_columns_and_indexes_with_named_index_same_as_column_name(
...
 
 
+@pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
+def test_dataframe_to_json_generator(module_under_test):
+    utcnow = datetime.datetime.utcnow()
+    df_data = collections.OrderedDict(
+        [
+            ("a_series", [pandas.NA, 2, 3, 4]),
+            ("b_series", [0.1, float("NaN"), 0.3, 0.4]),
+            ("c_series", ["a", "b", pandas.NA, "d"]),
```
plamut, line 744: If I understood correctly, `pandas.NA` was introduced in pandas 1.0?

In `setup.py` pandas is currently pinned to `>=0.23.0`, meaning that this test will fail. We should modify the `skipif` condition to also skip if too old a version is installed.

tests/unit/test__pandas_helpers.py
```diff
@@ -734,6 +734,35 @@ def test_list_columns_and_indexes_with_named_index_same_as_column_name(
...
+    dataframe = dataframe.astype({"a_series": pandas.Int64Dtype()})
+
+    rows = module_under_test.dataframe_to_json_generator(dataframe)
+    expected = [
+        {"b_series": 0.1, "c_series": "a", "d_series": utcnow, "e_series": True},
+        {"a_series": 2, "c_series": "b", "d_series": utcnow, "e_series": False},
+        {"a_series": 3, "b_series": 0.3, "d_series": utcnow, "e_series": True},
+        {"a_series": 4, "b_series": 0.4, "c_series": "d"},
+    ]
+    for row, expect in zip(rows, expected):
```
plamut, line 762: In theory, this could miss the case when the number of rows does not match, because `zip()` stops when the shortest iterable is exhausted. One way around it is replacing it with `zip_longest()`:

```suggestion
    for row, expect in itertools.zip_longest(rows, expected, fillvalue=object()):
```

Or even better, why not just `assert list(rows) == expected`? It will do both the row-wise comparisons and checking if the lengths of the lists match.

plamut (changes requested):
Thanks for the fix and the test, it generally looks good! Just made two improvement suggestions.

plamut:
FWIW, the `prerelease-deps` failure is flakiness, while Python 3.6 unit tests failure is legitimate - we use the oldest supported depenency versions there, and the new test failed with an attribute error:

```
 AttributeError: module 'pandas' has no attribute 'NA'
```

tests/unit/test__pandas_helpers.py
LinuxChristian: I agree, `assert list(rows) == expected` is a lot better.

tests/unit/test__pandas_helpers.py
LinuxChristian: Very good point. I have copied the `PANDAS_..._VERSION` syntax from `test_client` and updated the syntax as you suggested.

`pytest.importorskip` would condense the logic a bit but I can see you use `skipIf` everywhere else so consistency is perhaps best here.

```python
def test_dataframe_to_json_generator(module_under_test):
    pytest.importorskip(
        "pandas",
        minversion="1.0.0",
        reason="Requires `pandas version >= 1.0.0` which introduces pandas.NA"
    )
```

LinuxChristian:
I added one commit per comment so it's easier to review. If you would like me to squash a final version let me know.

tests/unit/test__pandas_helpers.py
```diff
@@ -734,7 +743,10 @@ def test_list_columns_and_indexes_with_named_index_same_as_column_name(
     assert columns_and_indexes == expected
 
 
-@pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
+@pytest.mark.skipIf(
+    pandas is None or PANDAS_INSTALLED_VERSION < PANDAS_MINIUM_VERSION,
+    reason="Requires `pandas version >= 1.0.0` which introduces pandas.NA",
```
plamut, line 748: Neat to also include the reason _why_ a minimum version is needed. :+1:

tests/unit/test__pandas_helpers.py
plamut: Currently both are used. I actually like `impororskip` more, but just copy/pasted the first existing example I found. :slightly_smiling_face: 

We should make conditionally skipping the tests internally consistent, but that's a thing for a separate PR.

plamut (approved):
LGTM, thanks for the quick update!

Nothing to worry about squashing the commits, we do that when the PR is merged.

plamut:
It appears that the test was not skipped?

```
    def test_dataframe_to_json_generator(module_under_test):
        utcnow = datetime.datetime.utcnow()
        df_data = collections.OrderedDict(
            [
>               ("a_series", [pandas.NA, 2, 3, 4]),
                ("b_series", [0.1, float("NaN"), 0.3, 0.4]),
                ("c_series", ["a", "b", pandas.NA, "d"]),
                ("d_series", [utcnow, utcnow, utcnow, pandas.NaT]),
                ("e_series", [True, False, True, None]),
            ]
        )
E       AttributeError: 'NoneType' object has no attribute 'NA'
```

Also reproducible locally.

tests/unit/test__pandas_helpers.py
```diff
@@ -734,6 +743,37 @@ def test_list_columns_and_indexes_with_named_index_same_as_column_name(
     assert columns_and_indexes == expected
 
 
+@pytest.mark.skipIf(
```
plamut, line 746: OK now I see it, I hate this gotcha. :)

```suggestion
@pytest.mark.skipif(
```

PR #751, opened 2021-07-12
fix: avoid possible job already exists error

### LINKED ISSUE #738: Job already exists
Python: 3.9 Cloud function runtime

Lib:
google-cloud-bigquery==2.15.0
google-auth==1.30.0
google-api-python-client==2.7.0

`
"/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 3149, in query query_job._begin(retry=retry, timeout=timeout) File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/bigquery/job/query.py", line 1044, in _begin super(QueryJob, self)._begin(client=client, retry=retry, timeout=timeout) File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/bigquery/job/base.py", line 461, in _begin api_response = client._call_api( File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 685, in _call_api return call() File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/api_core/retry.py", line 285, in retry_wrapped_func return retry_target( File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/api_core/retry.py", line 188, in retry_target return target() File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/_http.py", line 483, in api_request raise exceptions.from_http_response(response) google.api_core.exceptions.Conflict: 409 POST https://bigquery.googleapis.com/bigquery/v2/projects/<project_id>/jobs?prettyPrint=false: Already Exists: Job <project>:US.<Job_id>(job ID: <Job_id>) `

I got the following error in one of production environment, its first time i have encountered this error.

plamut:
Fixes #738.

If job create request fails, a query job might still have started successfully. This PR handles this edge case and returns such
query job one can be found.

Based on the [similar fix](https://github.com/googleapis/java-bigquery/blob/e755470ad3699eee8816f55cdec4dd995cf85d52/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java#L385-L401) in the Java client.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -3225,11 +3225,10 @@ class Client(ClientWithProject):
 
         try:
             query_job._begin(retry=retry, timeout=timeout)
-        except core_exceptions.GoogleAPICallError as create_exc:
-            # Even if create job request fails, it is still possible that the job has
-            # actually been successfully created. We check this by trying to fetch it,
-            # but only if we created a random job ID ourselves (otherwise we might
-            # mistakenly fetch a job created by someone else).
+        except core_exceptions.Conflict as create_exc:
+            # The thought is if someone is providing their own job IDs and they get
+            # their job ID generation wrong, this could end up returning results for
+            # the wrong query. We thus only try to recover if job ID was not given.
             if job_id_given:
                 raise create_exc
 
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -3190,6 +3190,7 @@ class Client(ClientWithProject):
                 If ``job_config`` is not an instance of :class:`~google.cloud.bigquery.job.QueryJobConfig`
                 class.
         """
+        job_id_given = job_id is not None
         job_id = _make_job_id(job_id, job_id_prefix)
 
         if project is None:
@@ -3221,9 +3222,31 @@ class Client(ClientWithProject):
 
         job_ref = job._JobReference(job_id, project=project, location=location)
         query_job = job.QueryJob(job_ref, query, client=self, job_config=job_config)
-        query_job._begin(retry=retry, timeout=timeout)
 
-        return query_job
+        try:
+            query_job._begin(retry=retry, timeout=timeout)
+        except core_exceptions.GoogleAPICallError as create_exc:
+            # Even if create job request fails, it is still possible that the job has
+            # actually been successfully created. We check this by trying to fetch it,
+            # but only if we created a random job ID ourselves (otherwise we might
+            # mistakenly fetch a job created by someone else).
+            if job_id_given:
+                raise create_exc
+
+            try:
+                query_job = self.get_job(
+                    job_id,
+                    project=project,
+                    location=location,
+                    retry=retry,
+                    timeout=timeout,
+                )
+            except core_exceptions.GoogleAPIError:  # (includes RetryError)
+                raise create_exc
+            else:
+                return query_job
+        else:
+            return query_job
 
     def insert_rows(
         self,
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -4617,53 +4617,69 @@ class TestClient(unittest.TestCase):
             },
         )
 
-    def test_query_job_rpc_fail_w_job_id_given(self):
-        from google.api_core.exceptions import GoogleAPICallError
+    def test_query_job_rpc_fail_w_random_error(self):
+        from google.api_core.exceptions import Unknown
         from google.cloud.bigquery.job import QueryJob
 
         creds = _make_credentials()
         http = object()
         client = self._make_one(project=self.PROJECT, credentials=creds, _http=http)
 
-        job_create_error = GoogleAPICallError("Dubious oops.")
+        job_create_error = Unknown("Not sure what went wrong.")
         job_begin_patcher = mock.patch.object(
             QueryJob, "_begin", side_effect=job_create_error
         )
         with job_begin_patcher:
-            with pytest.raises(GoogleAPICallError, match="Dubious oops."):
+            with pytest.raises(Unknown, match="Not sure what went wrong."):
                 client.query("SELECT 1;", job_id="123")
 
-    def test_query_job_rpc_fail_w_random_id_job_create_indeed_failed(self):
-        from google.api_core.exceptions import GoogleAPICallError
-        from google.api_core.exceptions import NotFound
+    def test_query_job_rpc_fail_w_conflict_job_id_given(self):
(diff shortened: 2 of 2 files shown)


google/cloud/bigquery/client.py
```diff
@@ -3190,6 +3190,7 @@ def query(
                 If ``job_config`` is not an instance of :class:`~google.cloud.bigquery.job.QueryJobConfig`
                 class.
         """
+        job_id_given = job_id is not None
```
tseaver, line 3193: I'm not sure why it matters whether the user has passed in the `job_id`:  if we call `QueryJob._begin`, and the back-end returns `409 Conflict`, ISTM we should handle it the same either way.

google/cloud/bigquery/client.py
```diff
@@ -3221,9 +3222,31 @@ def query(
 
         job_ref = job._JobReference(job_id, project=project, location=location)
         query_job = job.QueryJob(job_ref, query, client=self, job_config=job_config)
-        query_job._begin(retry=retry, timeout=timeout)
 
-        return query_job
+        try:
+            query_job._begin(retry=retry, timeout=timeout)
+        except core_exceptions.GoogleAPICallError as create_exc:
```
tseaver, line 3228: Wouldn't it make sense only to catch `core_exceptions.Conflict` here?  Other errors presumably can't involve an already-created job.

plamut:
@tseaver I don't know the exact mechanics on the backend, this fix is mostly based on a similar [fix](https://github.com/googleapis/java-bigquery/blob/e755470ad3699eee8816f55cdec4dd995cf85d52/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/BigQueryImpl.java#L385-L401) in the Java client.

@tswast Can you chime in?

google/cloud/bigquery/client.py
plamut: If I understood correctly, we want to catch any exception here, because even if one occurs, a job might have nevertheless been created. Admittedly, that's weird to me.

google/cloud/bigquery/client.py
tswast: In Java the thought was if someone is providing their own job IDs and they get their job ID generation wrong, this could end up returning results for the wrong query.

google/cloud/bigquery/client.py
plamut: Makes sense, I changed the logic to only try to recover from Conflict errors.

plamut:
The docs check failure does not seem to be related:
```
gpg: keyserver receive failed: No name
```

google/cloud/bigquery/client.py
tseaver: OK.

tseaver (approved):

tseaver:
https://github.com/googleapis/synthtool/pull/1155 landed here in #762.  I'm not sure why the config isn't making you merge with `master` to pick that fix up, however.

tswast (approved):

google/cloud/bigquery/client.py
```diff
@@ -3221,9 +3222,30 @@ def query(
...
+            query_job._begin(retry=retry, timeout=timeout)
+        except core_exceptions.Conflict as create_exc:
+            # The thought is if someone is providing their own job IDs and they get
+            # their job ID generation wrong, this could end up returning results for
+            # the wrong query. We thus only try to recover if job ID was not given.
+            if job_id_given:
+                raise create_exc
+
+            try:
+                query_job = self.get_job(
```
andrewjw, line 3236: Hi, there is a slight problem with this change - self.get_job has a different return type to this function. It can return LoadJob, etc as well as the QueryJob we're expecting so the actual return type doesn't match what is declared for this function.

I don't understand the situations that could result in this code being called, but presumably in reality this would always be a QueryJob? Unfortunately this is causing me problems when running pylint over some code that calls this, because it thinks the function can return LoadJob, and that has a different set of members to QueryJob.

Many thanks,
Andrew Wilkinson

plamut: Indeed, in this context `self.get_job()` returns a `QueryJob`, because `job_id` is the same ID that was used a few lines above when constructing a new query job (and then starting it).

This project uses `pytype` for static type checks and it did not complain, but apparently `pylint` could not deduce the same and reported a false issue.

Could you tell `pylint` to ignore return type in that specific line where `query()` is called? IMHO that justifiable, because `pylint` is wrong there.

andrewjw: Having looked into this a bit further I agree that pylint is wrong. It's a bit of a pain to have disable this check every time we call query, but I think this is a sign that pylint is aging and not keeping up with modern Python's type syntax.

Sorry for the noise.

Cheers,
Andrew

plamut: No worries, it was a perfectly valid comment.

Ideally, `pylint` would allow ignoring particular warnings for lines matching a regex, but I'm not sure if that's currently supported? It would make disabling those false positives much cleaner compared to spamming the `# pylint: disable=...` comments all around the code.

andrewjw: Sadly the error isn't raised on the call to `query`, but when you try and use the return value. In my case this is accessing `num_dml_affected_rows`, which only exists on `QueryJob`, and not on `LoadJob`. Even if it did support disabling errors using a regex, I'm not sure it would be practical to create one.

It's been bugging me why this wouldn't be picked up by the type checker. I think I've tracked it down to the fact that `LoadJob`, `QueryJob`, etc all derive from `_AsyncJob`, which in turn derives from `google.api_core.future.polling.PollingFuture`. The problem is that `google.api_core.future.polling.PollingFuture` is not typable, so it gets turns into an `Any` type, which makes all the job types equivalent and therefore doesn't generate an error. When testing with mypy you have to add `# type: ignore` to the `PollingFuture` import line explicitly, but I guess `pytype` is more forgiving.

I've create the attached file demonstrating the problem (annoyingly github won't let me attach the file as a .py). As currently written it'll generate an error in both mypy and pytype, but swap the comments on lines 5 and 6 and the error goes away.

Anyway, I have a reasonable workaround, so if you want to leave this that's absolutely fine. If in future the `python-api-core` library adds typing then I expect this to break though. Adding an `isinstance(query_job, job.QueryJob` will resolve the issue.

Cheers,
Andrew
[invalid_return_union.txt](https://github.com/googleapis/python-bigquery/files/6888519/invalid_return_union.txt)

PR #758, opened 2021-07-14
feat: add support for more detailed DML stats

### LINKED ISSUE #749: Add support for additional DML statistics from MERGE statements
This feature request covers the additional of new information returned from MERGE statements on the total number of rows affected.  There are 3 types of information:
- Total number of records updated
- Total number of records inserted
- Total number of records deleted

Additional design and detail to be shared with implementers separately.

plamut:
Closes #749.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -114,6 +114,38 @@ def _to_api_repr_table_defs(value):
     return {k: ExternalConfig.to_api_repr(v) for k, v in value.items()}
 
 
+class DmlStats(typing.NamedTuple):
+    """Detailed statistics for DML statements.
+
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/DmlStats
+    """
+
+    inserted_row_count: int = 0
+    """Number of inserted rows. Populated by DML INSERT and MERGE statements."""
+
+    deleted_row_count: int = 0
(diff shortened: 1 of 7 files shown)


google/cloud/bigquery/job/query.py
```diff
@@ -114,6 +114,38 @@ def _to_api_repr_table_defs(value):
...
+    """Number of deleted rows. populated by DML DELETE, MERGE and TRUNCATE statements.
+    """
+
+    updated_row_count: int = 0
+    """Number of updated rows. Populated by DML UPDATE and MERGE statements."""
+
+    @classmethod
+    def from_api_repr(cls, stats: Optional[Dict[str, str]]) -> Optional["DmlStats"]:
+        if stats is None:
+            return None
```
tseaver, line 136: It feels icky to do the test for `None` here, rather than in the property below.  Is that pattern common in BQ?

plamut: I checked and while this pattern [does exist](https://github.com/googleapis/python-bigquery/blob/5437d443c3e89cf0458771662c9ef2d2f2f8e4d8/google/cloud/bigquery/schema.py#L448-L470), it's very rare. I'll move the check to where the `DmlStats` property actually gets serialized.

tseaver (approved):

PR #203, opened 2020-07-30
chore: copy all samples into samples/snippets

busunkim96:
TODO:

- [ ] pin existing samples to commit hash
- [ ] merge this PR
- [ ] update samples to use master again

PR #766, opened 2021-07-14
refactor: blacken samples

### LINKED ISSUE #765: blacken samples using templated noxfile
This repository has three historical samples locations (see https://github.com/googleapis/python-bigquery/issues/790 for the issue to migrate them all to the most recent layout, supported by templates), but this issue only refers to the samples in (3). The samples in (1, 2) should remain blackened by the existing mechanism:

1. `docs/snippets.py`
2. `samples/*.py` + `samples/tests/*.py`
3. `samples/snippets/*.py`, `samples/geography/*.py`, `samples/magics/*.py`, and possibly more in the future, see: https://github.com/googleapis/python-bigquery/issues/1352.

For this issue, I anticipate the following updates:

1. Update the blacken session in noxfile.py to only blacken the samples in legacy directories. That is https://github.com/googleapis/python-bigquery/blob/34a3f5cf34d4a08889fe1407f4ad6ce3c9d93838/noxfile.py#L28 should be updated to `BLACK_PATHS = ("docs", "google", "samples/*.py", "samples/tests/*.py", "tests", "noxfile.py", "setup.py")` or similar.
2. Update owlbot config to make sure the code samples in `samples/snippets|geography|etc` stay blackened. We still want to run `nox -s blacken` in the root directory but then we also want to run `nox -s blacken` in each samples directory that has a `noxfile.py`. See the implementation in python-bigquery-datatransfer for reference: https://github.com/googleapis/python-bigquery-datatransfer/blob/822223fc02b1e2ff6f52b834c51f4bf46924e2e8/owlbot.py#L60-L61

tswast:

Fixes #765 🦕


tswast:
I noticed some conflicts with `pre-commit` and the version of black used in samples sessions. Mailed https://github.com/googleapis/synthtool/pull/1158 to address that.

tseaver (approved):

plamut:
> I noticed some conflicts with `pre-commit` and the version of black used in samples sessions.

Aha, so that's why the lint check failed, one of the snippets was still re-formatted.

Ignoring that, the PR looks good.

plamut (approved):

tswast:
Hmm... Turns out these were already being blackened by the root noxfile. Twice, actually, due to `docs` and `samples` included here: https://github.com/googleapis/python-bigquery/blob/c45a7380871af3dfbd3c45524cb606c60e1a01d1/noxfile.py#L26

Since we've still got a mess of files in the root `samples/` dir like https://github.com/googleapis/python-bigquery/blob/master/samples/add_empty_column.py

I think it'd be best to wait on this until we do something like https://github.com/googleapis/python-bigquery/pull/203 (We'll have to do the move in two phases to avoid breaking cloud.google.com links)

Issue #733, opened 2021-06-30
tests/system/test_pandas.py::test_insert_rows_from_dataframe is flaky


jimfulton:
In CI, this test has failed at least twice in the last day.

```
>       assert len(row_tuples) == len(expected)
E       AssertionError: assert 0 == 6
E        +  where 0 = len([])
E        +  and   6 = len([(1.11, True, 'my string', 10), (2.22, False, 'another string', 20), (3.33, False, 'another string', 30), (4.44, True, 'another string', 40), (5.55, False, 'another string', 50), (6.66, True, None, 60)])

tests/system/test_pandas.py:696: AssertionError
```

Perhaps this due to eventual consistency and we need to poll until we get 6 results.

plamut:
Interestingly, I haven't observed this in the last 2+ years. As per docs, streamed data should be available within a [few seconds](https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataavailability), but something might have changed so that even this is sometimes too late.

@shollyman Is there a reasonable upper bound on those "few seconds"? Just to know how long does it make sense to keep polling in the test.

plamut:
Also linking a [comment](https://github.com/googleapis/python-bigquery/pull/728#issuecomment-871571356) by @tseaver for reference.

tswast:
Queries should be reading from the streaming buffer, so I'm a bit surprised this started flaking.

tswast:
I'm seeing the "within a few seconds" language now. I think that's a bit different from how it used to be. Polling until we get 6 results (max a few times) seems the right option.

Issue #589, opened 2021-04-06
Provide support for synchronous queries through the v2/projects/{projectId}/queries endpoint


kdeggelman:
**Is your feature request related to a problem? Please describe.**
We are part of the BI Engine preview so we're trying to minimize the latency between our python application and BigQuery. The current approach of creating an asynchronous job and then waiting for it to complete adds significant latency and hurts the value proposition of BI Engine.
 **Describe the solution you'd like**
We'd like a way to utilize the `v2/projects/{projectId}/queries` endpoint to execute a query and wait for the response.
 **Describe alternatives you've considered**
We could call the REST API directly. The main drawbacks are 1) authentication with a service account 2) the existing python client does a nice job of creating `Row` objects with the results.
I've also used the following snippet:
```python
bq_client = bigquery.Client(project=project_id)
synchronous_api_response = bq_client._connection.api_request(
    "POST", 
    f"/projects/{project_id}/queries",
    data={"query": query, "useLegacySql": False},
)
```

plamut:
@kdeggelman I'm not sure it would be feasible to change how the said [endpoint](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query) behaves on the backend. Even if POST-ing to it directly, it just creates a new job to run the actual query, and the code would still need to [fetch query results](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) using the job ID obtained from the first call's response.

As an alternative, have you maybe considered using the faster [BigQuery Storage API](https://cloud.google.com/bigquery/docs/reference/storage)? It's more performant than the REST API, allows streaming the rows over multiple streams, etc.

Could that be something that can help with your use case?

kdeggelman:
Thanks for the help @plamut!

> and the code would still need to fetch query results using the job ID obtained from the first call's response.

The doc you linked to for [`jobs.query`](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query) says _"Runs a BigQuery SQL query synchronously and returns query results if the query completes within a specified timeout."_ which is the behavior that I've observed. I am able to grab the `rows` straight from the [response body](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#response-body) of that endpoint without hitting `getQueryResults`.

[...]

PR #776, opened 2021-07-16
deps!: BigQuery Storage and pyarrow are required dependencies

### LINKED ISSUE #757: pull google-cloud-bigquery-storage and pyarrow into required dependencies
https://github.com/googleapis/python-bigquery-storage/pull/224/files#r667241381

We'll still want to keep `bqstorage` around in "extras" so that `pip install google-cloud-bigquery[bqstorage]` still succeeds, but the `[bqstorage` part will be a no-op.

plamut:
Closes #757.

This might already be a final version, but I still need to check a few additional things:
 - ~Remove logic for "ensure client"? Or maybe not, since we still allow tabledata.list?~
   We still need that.
 - ~Remove min version check for BQ Storage, since now it is guaranteed `>=2.0.0` will be installed (no more `pip` quirks with extras).~
   Done.
 - ~Remove checks for GAPIC? It's now a required dependency.~
   Not applicable, the minimum GAPIC version was never checked at runtime.

**PR checklist:**:


plamut:
@tswast It might be easier to review each commit separately, I tried to make them independent from each other. Also, double check if the assumption stated in the last commit's message really holds - if not, we can revert that bit.

setup.py
```diff
@@ -32,24 +32,18 @@
     "grpcio >= 1.38.1, < 2.0dev",  # https://github.com/googleapis/python-bigquery/issues/695
     "google-api-core[grpc] >= 1.29.0, < 3.0.0dev",
     "proto-plus >= 1.10.0",
+    "google-cloud-bigquery-storage >= 2.0.0, <3.0.0dev",
     "google-cloud-core >= 1.4.1, < 3.0dev",
     "google-resumable-media >= 0.6.0, < 3.0dev",
     "packaging >= 14.3",
     "protobuf >= 3.12.0",
+    "pyarrow >= 1.0.0, < 5.0dev",
```
tseaver, line 40: Making these previously-optional dependencies now required feels like it might be a `semver: major` change.

plamut: IIRC @tswast said in a chat that it shouldn't be a problem for most users, and those who cannot install these dependencies for any reason can still pin to a lower (non-major version). But reading this as I write, we might actually want to re-consider this.

plamut: There has been a [debate](https://github.com/googleapis/python-bigquery/issues/757#issuecomment-879908013) on this on the issue itself. Also copying Tim's input:

> The gist is that the problematic dependencies are already required transitively anyway
With pyarrow being the exception
It's probably worth the bump for pyarrow if we're being conservative about it. Esp since we do know of environments where adding it will break

tswast: Let's go with the major version bump. I'd rather people adopt this change a little more slowly than break someone who has a working 2.x branch without pyarrow

google/cloud/bigquery/exceptions.py
```diff
@@ -18,4 +18,8 @@ class BigQueryError(Exception):
 
 
 class LegacyBigQueryStorageError(BigQueryError):
-    """Raised when too old a version of BigQuery Storage extra is detected at runtime."""
+    """Raised when too old a version of BigQuery Storage extra is detected at runtime.
+
+    .. deprecated:: 2.2.0
+       Not used anymore and will be removed in the future.
```
plamut, line 24: Note to self - this should have been 2.22.0, but it's not obsolete, since we've just made that [release](https://github.com/googleapis/python-bigquery/pull/771#issuecomment-882856588). Needs to be adjusted once we decide what the next version will be.

setup.py
```diff
@@ -32,24 +32,18 @@
     "grpcio >= 1.38.1, < 2.0dev",  # https://github.com/googleapis/python-bigquery/issues/695
     "google-api-core[grpc] >= 1.29.0, < 3.0.0dev",
     "proto-plus >= 1.10.0",
+    "google-cloud-bigquery-storage >= 2.0.0, <3.0.0dev",
```
plamut, line 35: This should still allow 1.x, as per https://github.com/googleapis/python-bigquery/pull/784/files

It also means we probably need to add the minimum version check back. Or wait for https://github.com/googleapis/google-cloud-python/issues/10566 to be resolved.

google/cloud/bigquery/exceptions.py
plamut: Done. Removed the custom exception class altogether, because we'll do a major version bump, thus a breaking change is acceptable.

setup.py
plamut: Roger.

Is there perhaps anything else that we want to clean up and would be a breaking change? (possibly in a separate PR)
A major version bump is a good opportunity for that.

tswast: https://github.com/googleapis/python-bigquery/issues/319 is still on my mind. Perhaps easier now that proto-plus added more case options?

tswast: Another breaking change we might want to do: https://issuetracker.google.com/144712110 -- Better type conversions in the pandas connector.

busunkim96:
Hi,

Stopping by to follow up on https://github.com/googleapis/google-cloud-python/issues/10566#issuecomment-883215649. That issue is aimed specifically at `google-cloud-core`, `google-api-core`, and `google-auth` libraries. Are you anticipating having to require >=2.x.x for any of those libraries to support this change?

plamut:
@busunkim96 I edited that [comment](https://github.com/googleapis/python-bigquery/pull/776#discussion_r672860955). :)

I don't think so, we are just making a few extras required now, nothing to do with core dependencies (wrote that comment too fast).

setup.py
plamut: And yet another one from today - maybe bumping the minimum `pyarrow` version, as the older ones suffer from a [bug](https://github.com/googleapis/python-bigquery/issues/781#issuecomment-883497414) (load data from dataframe can upload incorrect data, yikes!)

tswast: That is a nasty bug. Thankfully `pyarrow` didn't follow [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) this time and kept support for Python 3.6 in the 4.x packages. I'd be okay with bumping the minimum for that.

tests/unit/test__pandas_helpers.py
```diff
@@ -29,13 +29,9 @@
     import pandas.testing
 except ImportError:  # pragma: NO COVER
     pandas = None
-try:
```
tswast, line 32: I notice a test failure:

```
________________ test_dataframe_to_parquet_dict_sequence_schema ________________

module_under_test = <module 'google.cloud.bigquery._pandas_helpers' from '/tmpfs/src/github/python-bigquery/google/cloud/bigquery/_pandas_helpers.py'>

    @pytest.mark.skipif(isinstance(pyarrow, mock.Mock), reason="Requires `pyarrow`")
    def test_dataframe_to_parquet_dict_sequence_schema(module_under_test):
        dict_schema = [
            {"name": "field01", "type": "STRING", "mode": "REQUIRED"},
            {"name": "field02", "type": "BOOL", "mode": "NULLABLE"},
        ]

>       dataframe = pandas.DataFrame(
            {"field01": [u"hello", u"world"], "field02": [True, False]}
        )
E       AttributeError: 'NoneType' object has no attribute 'DataFrame'

tests/unit/test__pandas_helpers.py:1252: AttributeError
```

Can we use `importorskip` here for pandas?

tests/unit/test__pandas_helpers.py
```diff
@@ -1093,33 +1077,6 @@ def test_dataframe_to_parquet_compression_method(module_under_test):
...
-    with no_pyarrow_patch, warnings.catch_warnings(record=True) as warned:
-        detected_schema = module_under_test.dataframe_to_bq_schema(
-            dataframe, bq_schema=[]
-        )
-
-    assert detected_schema is None
-
-    # a warning should also be issued
-    expected_warnings = [
-        warning for warning in warned if "could not determine" in str(warning).lower()
```
tswast, line 1116: Looks like the warning for this isn't issued anymore. I wonder if it's possible to get a dataframe without detectable schema if pyarrow is installed?

tswast:
I'm pretty sure it would be possible to delete even more code such as https://github.com/googleapis/python-bigquery/blob/d1cbc3817a4b93f61356bd14ba51fb176e5d0269/google/cloud/bigquery/_pandas_helpers.py#L542-L554 if we always have a BigQuery Storage API client, but there are a few cases where it can't be used.

For example: `list_rows(table).to_dataframe()` won't be able to use BQ Storage API from free tier. Also, when we implement https://github.com/googleapis/python-bigquery/issues/589 we'll need to use the REST API for that (at least in the single page of results case)

plamut:
As a follow-up to https://github.com/googleapis/python-bigquery/issues/781 I also bumped the minimum `pyarrow` version.

setup.py
plamut: And yet another - if this gets delayed significantly (read: months), we might want to consider dropping Python 3.6 support as well. But Python 3.6 [end of life](https://www.python.org/dev/peps/pep-0494/#id15) will only happen in December this year.

tseaver:
Failing systest is another instance of #733.

tests/unit/test__pandas_helpers.py
plamut: This particular test covered the [case](https://github.com/googleapis/python-bigquery/blob/e58702967d572e83b4c774278818302594a511b7/google/cloud/bigquery/_pandas_helpers.py#L346-L352) when we couldn't even start detecting the schema with `pyarrow` because the later is missing. This is not possible anymore, as `pyarrow` is now required, thus that particular warning has been removed.

However, we still call [augment_schema()](https://github.com/googleapis/python-bigquery/blob/e58702967d572e83b4c774278818302594a511b7/google/cloud/bigquery/_pandas_helpers.py#L354-L411) immediately after the deleted lines. That function can also issue a warning on its own if schema detection fails for some fields, but those paths are covered by other tests.

tests/unit/test__pandas_helpers.py
plamut: Done.

google/cloud/bigquery/client.py
```diff
@@ -2576,16 +2572,6 @@ def load_table_from_dataframe(
...
             if job_config.source_format == job.SourceFormat.PARQUET:
-                if _PYARROW_VERSION in _PYARROW_BAD_VERSIONS:
-                    msg = (
-                        "Loading dataframe data in PARQUET format with pyarrow "
-                        f"{_PYARROW_VERSION} can result in data corruption. It is "
-                        "therefore *strongly* advised to use a different pyarrow "
-                        "version or a different source format. "
-                        "See: https://github.com/googleapis/python-bigquery/issues/781"
-                    )
-                    warnings.warn(msg, category=RuntimeWarning)
```
plamut, line 2626: This warning was removed because the just-fixed bug will not come into play once we require `pyarrow > 3.0.0`.

tswast (approved):
This is awesome!

FYI: I pointed this PR at the `v3` branch. Should be safe to merge.

PR #780, opened 2021-07-19
Add "timeout" to retryable reasons

### LINKED ISSUE #779: Streaming inserts are not retried when a timeout occurs
Expected behaviour: streaming inserts are retried on timeout

quentin-sommer:
Fixes #779

ℹ️  This modifies the default retry used for all the API calls. Should I instead make a retry object dedicated to the streaming inserts?

### DIFF
--- a/google/cloud/bigquery/retry.py
+++ b/google/cloud/bigquery/retry.py
@@ -19,7 +19,7 @@ import requests.exceptions
 
 
 _RETRYABLE_REASONS = frozenset(
-    ["rateLimitExceeded", "backendError", "internalError", "badGateway"]
+    ["rateLimitExceeded", "backendError", "internalError", "badGateway", "timeout"]
 )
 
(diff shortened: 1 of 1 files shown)


tswast:
I'd like to see a stacktrace of the failure to confirm this is the right way to solve this.

One thing we might be able to do to reproduce is after we implement https://github.com/googleapis/python-bigquery/issues/897 we can try setting the X-Server-Timeout header to a very small value.

tswast:
In internal issue 195337762, we did some testing with the `X-Server-Timeout` header and these were actually reported as `"reason": "backendError"`. Without examples of where `"reason": "timeout"` appears, I don't think we should accept this PR.

From https://cloud.google.com/bigquery/docs/error-messages, the "timeout" reason happens when "The job timed out.", which to me if anything indicates this should be added to job retry options, but without an example I don't think we should do that either.

PR #786, opened 2021-07-20
fix!: use nullable `Int64` and `boolean` dtypes in `to_dataframe`

### LINKED ISSUE #793: use pandas Int64 type by default to avoid precision loss
The Int64 dtype has been around long enough now that I'd be comfortable with using it by default. In https://github.com/googleapis/python-bigquery/pull/786 I notice that `float64` dtype is used for nullable integers even when this would cause loss-of-precision.

tswast:
To override this behavior, specify the types for the desired columns with the
`dtype` argument.

BREAKING CHANGE: uses Int64 type by default to avoid loss-of-precision in results with large integer values

Fixes https://issuetracker.google.com/144712110 🦕
Fixes #793


tswast:
I'll take a closer look at https://github.com/googleapis/python-bigquery/pull/776 before finishing this one, as it might mean fewer code paths to cover. I think the BQ Storage API will always be used for `to_dataframe` after that PR.

tswast:
I did a little bit of experimentation to see what the intermediate `pyarrow.Table` types are for both REST and BQ Storage API on all scalar types. They do align, so that's good. Also, floating points appear to be handled correctly, even with null values in the table.

It appears https://issuetracker.google.com/144712110 was fixed for FLOAT columns in https://github.com/googleapis/python-bigquery/pull/314 as of google-cloud-bigquery >= 2.2.0 (That was technically a breaking change [oops])

I might still keep this open so that we can have some explicit tests for different data types. Also, we're relying on PyArrow -> Pandas to pick the right data types, so maybe there's some dtype defaults we can help with still.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -217,6 +222,26 @@ def bq_to_arrow_schema(bq_schema):
...
+    includes Int64, which has pandas.NA values and does not result in
+    loss-of-precision.
+
+    Returns:
+        Dict[str, str]: mapping from column names to dtypes
+    """
+    dtypes = {}
+    for bq_field in bq_schema:
+        if bq_field.mode.upper() not in {"NULLABLE", "REQUIRED"}:
+            continue
```
tswast, line 238: TODO(tswast): Need unit tests for this line.

tswast: Done in https://github.com/googleapis/python-bigquery/pull/786/commits/8f90c511d9e97c0341352b71ee4499d886c9c78d

tswast:
Re: system test failure:

```
_____________ TestBigQuery.test_load_avro_from_uri_then_dump_table _____________
.
.
.
E   google.api_core.exceptions.RetryError: Deadline of 120.0s exceeded while calling functools.partial(<bound method PollingFuture._done_or_raise of <google.cloud.bigquery.job.load.LoadJob object at 0x7ff6eae07f40>>), last exception:
```

Didn't we increase the default deadline to 10 minutes? Maybe v3 branch needs a sync?

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -217,6 +222,26 @@ def bq_to_arrow_schema(bq_schema):
...
 
+def bq_schema_to_nullsafe_pandas_dtypes(bq_schema: Sequence[schema.SchemaField]):
+    """Return the default dtypes to use for columns in a BigQuery schema.
+
+    Only returns default dtypes which are safe to have NULL values. This
+    includes Int64, which has pandas.NA values and does not result in
+    loss-of-precision.
+
+    Returns:
+        Dict[str, str]: mapping from column names to dtypes
```
plamut, line 233: (nit) Can be expressed as the annotation of the function return type.

plamut (approved):
Two nits, but not essential, looks good.

tests/system/test_pandas.py
```diff
@@ -795,3 +795,71 @@ def test_list_rows_max_results_w_bqstorage(bigquery_client):
         dataframe = row_iterator.to_dataframe(bqstorage_client=bqstorage_client)
 
     assert len(dataframe.index) == 100
+
+
+@pytest.mark.parametrize(
+    ("max_results",), ((None,), (10,),)  # Use BQ Storage API.  # Use REST API.
+)
+def test_list_rows_nullable_scalars_dtypes(bigquery_client, scalars_table, max_results):
+    df = bigquery_client.list_rows(
```
tswast, line 804: Note to self: I'll need to exclude the INTERVAL column next time we sync with master

PR #787, opened 2021-07-21
fix: issue a warning if buggy pyarrow is detected

### LINKED ISSUE #781: load_table_from_dataframe produces incorrect results when used in list of dict
#### Environment details

  - OS type and version: Debian 10 (dataproc image 2.0-debian10)
  - Python version: `python --version`: Python 3.8.10
  - pip version: `pip --version`: pip 21.1.2
  - `google-cloud-bigquery` version: `pip show google-cloud-bigquery`: google-cloud-bigquery==2.6.2, pyarrow==2.0.0

#### Steps to reproduce

  1. Create a big dataframe (1000 lines) with a column containing a list (at least length 6) of identically structured dictionaries 
  2. Create a bq client and use load_table_from_dataframe to create a table in bigquery
  3. Check the resulting table in bigquery. Structures seem to switch values with other instances in the list. (eg should have [STRUCT('w0' AS name, 0.1 AS value),STRUCT('h1' AS name, 1.2 AS value)] but have [STRUCT('h1' AS name, 0.1 AS value),STRUCT('w0' AS name, 1.2 AS value)]. The big problem is not the order, is that the integrity of information of each structure is not kept (eg. 'w0' should be 0.1, not 1.2). 

#### Code example

```python
# create df with a list of dictionaries
# In this example, the dict structure is {"name": str, "value":float}. name is a letter + int, and value are increasingly big floats
data = [[[{'name':'whyist'[i]+str(i), 'value':np.random.random()*10**i} for i in range(6)]] for n in range(1000)]
df = pd.DataFrame(data, columns=['vals'])

# load
project = 'myproject'
bq_client = bigquery.Client(project=project)
bq_client.load_table_from_dataframe(
    dataframe=df,
    destination='tmp.test_bug',
    job_config = job_config
)

# Checking in bigquery, 
```
At least for this example, the 'value' attribute is transcribed in the correct order (first item has the smallest value, and it increases). The 'name' value was sampled with possibility of repetition. All table lines have the same 'name' values in the same order, and it can change if the code is reexecuted.

plamut:
Fixes #781.

Some `pyarrow` versions can cause issue when loading data from dataframe. This commit detects if such pyarrow version is installed and warns the user.

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -27,6 +27,7 @@ import itertools
 import json
 import math
 import os
+import packaging.version
 import tempfile
 from typing import Any, BinaryIO, Dict, Iterable, Optional, Sequence, Tuple, Union
 import uuid
@@ -34,6 +35,8 @@ import warnings
 
 try:
     import pyarrow
+
+    _PYARROW_VERSION = packaging.version.parse(pyarrow.__version__)
 except ImportError:  # pragma: NO COVER
     pyarrow = None
 
@@ -117,6 +120,9 @@ _LIST_ROWS_FROM_QUERY_RESULTS_FIELDS = "jobReference,totalRows,pageToken,rows"
 # connection timeout before data can be downloaded.
 # https://github.com/googleapis/python-bigquery/issues/438
 _MIN_GET_QUERY_RESULTS_TIMEOUT = 120
+
+# https://github.com/googleapis/python-bigquery/issues/781#issuecomment-883497414
+_PYARROW_BAD_VERSIONS = frozenset([packaging.version.Version("2.0.0")])
 
 
 class Project(object):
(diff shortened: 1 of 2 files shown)


plamut:
@tswast Alternatively we could ban `pyarrow==2.0.0`, since `1.0.x` seems to work just fine, although that might be more disturbing. Besides, the bug seems to occur in more specific cases, such as having a large enough `DataFrame` with arrays of records, but seems to work fine otherwise (at least we haven't heard about anyone complaining to date).

tswast (approved):
This approach seems good to me. I agree that it must be a pretty narrow set of conditions where this matters. If someone is stuck on pyarrow 2.0, hopefully this will at least make them aware.

google/cloud/bigquery/client.py
```diff
@@ -27,13 +27,16 @@
...
 from typing import Any, BinaryIO, Dict, Iterable, Optional, Sequence, Tuple, Union
 import uuid
 import warnings
 
 try:
     import pyarrow
+
+    _PYARROW_VERSION = packaging.version.parse(pyarrow.__version__)
 except ImportError:  # pragma: NO COVER
     pyarrow = None
```
tseaver, line 41: Potentially leaves `_PYARROW_VERSION` undefined.  Maybe:
```suggestion
    pyarrow = None
    _PYARROW_VERSION = None
```

plamut: It's only used in the path where `pyarrow` is definitely available (with a prior check), thus left it out. Although that implicit fact can change...

On the other hand, we are already making `pyarrow` required, thus it probably shouldn't be an issue in the transitional period, but a valid remark nevertheless.

PR #791, opened 2021-07-21
test: Stop creating extra datasets

### LINKED ISSUE #716: test: System tests create and destroy way too many datasets
Every system tests creates and destroys a temporary dataset in it's setUp and tearDown.

- Some system tests don't use the dataset created.
- Some system tests create and destroy their own datasets, seemingly oblivious that temporary datasets were already created for them.

This makes system tests take far longer than they should. ~10 minutes

Proposed fix:
- Don't create datasets by default.  Make tests elect to create them.
- Move toward pytest fixtures to elect creating datasets.
  - Use session fixtures to avoid  creating a dataset for each test.

jimfulton:
This saves ~12% in execution time and unecessary BQ churn.

Fixes #716  🦕

This mostly just stops creating extra datasets.  Only 3 tests used tests used the dataset created in `setUp`.  Those 3 tests now share the dataset created using the dataset_id fixture.  This saves ~12% off test-execution time.

More could be saved by converting most of the remaining tests in `test_client` to pytest and using the dataset_id fixture. That should be done in follow-on PRs.

### DIFF
--- a/tests/system/test_client.py
+++ b/tests/system/test_client.py
@@ -154,7 +154,6 @@ class Config(object):
 
     CLIENT: Optional[bigquery.Client] = None
     CURSOR = None
-    DATASET = None
 
 
 def setUpModule():
@@ -164,9 +163,7 @@ def setUpModule():
 
 class TestBigQuery(unittest.TestCase):
     def setUp(self):
-        Config.DATASET = _make_dataset_id("bq_system_tests")
-        dataset = Config.CLIENT.create_dataset(Config.DATASET)
-        self.to_delete = [dataset]
+        self.to_delete = []
 
     def tearDown(self):
         policy_tag_client = PolicyTagManagerClient()
@@ -1622,20 +1619,6 @@ class TestBigQuery(unittest.TestCase):
         row_tuples = [r.values() for r in rows]
         self.assertEqual(row_tuples, [(5, "foo"), (6, "bar"), (7, "baz")])
 
-    def test_dbapi_create_view(self):
-
-        query = """
-        CREATE VIEW {}.dbapi_create_view
-        AS SELECT name, SUM(number) AS total
-        FROM `bigquery-public-data.usa_names.usa_1910_2013`
-        GROUP BY name;
-        """.format(
-            Config.DATASET
-        )
-
-        Config.CURSOR.execute(query)
-        self.assertEqual(Config.CURSOR.rowcount, 0, "expected 0 rows")
-
     @unittest.skipIf(
         bigquery_storage is None, "Requires `google-cloud-bigquery-storage`"
     )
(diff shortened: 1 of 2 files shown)


tswast:
I see that we already have a session-scope `dataset_id` fixture in https://github.com/googleapis/python-bigquery/blob/e58702967d572e83b4c774278818302594a511b7/tests/system/conftest.py#L35

Maybe there's a way we can use that with our mix of `unittest` and `pytest` style tests?

tswast:
Maybe we could even update that `conftest.py` file to use the new "prefixer" module?

```python
import test_utils.prefixer

from . import helpers


prefixer = test_utils.prefixer.Prefixer("python-bigquery", "tests/system")


@pytest.fixture(scope="session", autouse=True)
def cleanup_datasets(bigquery_client: bigquery.Client):
    for dataset in bigquery_client.list_datasets():
        if prefixer.should_cleanup(dataset.dataset_id):
            bigquery_client.delete_dataset(
                dataset, delete_contents=True, not_found_ok=True
            )

...

@pytest.fixture(scope="session")
def dataset_id(bigquery_client: bigquery.Client, project_id: str):
    dataset_id = prefixer.create_prefix()
    full_dataset_id = f"{project_id}.{dataset_id}"
    dataset = bigquery.Dataset(full_dataset_id)
    bigquery_client.create_dataset(dataset)
    yield dataset_id
    bigquery_client.delete_dataset(dataset, delete_contents=True, not_found_ok=True)

```

jimfulton:
> I see that we already have a session-scope `dataset_id` fixture in
> 
> https://github.com/googleapis/python-bigquery/blob/e58702967d572e83b4c774278818302594a511b7/tests/system/conftest.py#L35

I'm using that.

> 
> Maybe there's a way we can use that with our mix of `unittest` and `pytest` style tests?

You can't use non-auto-use fixtures in unittest tests. I tried.  That's why I moved those three tests to pytest tests.

jimfulton:
> Maybe we could even update that `conftest.py` file to use the new "prefixer" module?

Will do.

jimfulton:
> > Maybe we could even update that `conftest.py` file to use the new "prefixer" module?
> 
> Will do.

done

tests/system/test_client.py
```diff
@@ -2476,104 +2459,6 @@ def test_list_rows_page_size(self):
...
-            destination=source_table_path,
-            job_config=copy_config,
-        )
-        copy_job.result()
-
-        rows_iter = client.list_rows(source_table_path)
-        rows = sorted(row.values() for row in rows_iter)
-        assert rows == [(1, "one"), (2, "two")]
-
     def temp_dataset(self, dataset_id, location=None):
```
tseaver, line 2462: This test uses the shared `dataset_id` fixture, but wants to create the dataset itself -- should it be updated to use a different one?

jimfulton: I'm confused.  Are you referring to `test_table_snapshots`?  It was using the dataset created in `setUp`. I don't see it creating a dataset. Am I blind? :)

tseaver: Sorry, I was referring to the `temp_dataset` helper (my eye read that as `test_dataset`).  It *is* used by `test_create_dataset`, `test_update_dataset`, etc.  Because it *isn't* a test, it isn't using the `dataset_id` fixture, but an actual passed-in `dataset_id` parameter.

I don't know whether all those tests actually need to be using separate datasets, but could imagine that at least some could share.

jimfulton: A number of the other tests could and should be refactored to use the shared dataset.  In the interest of not making this PR too complex, I opted to save updating the other tests for a future PR.

jimfulton: This PR does the minimum to get rid of the mostly-unused dataset creation in `setUp`.

tswast (approved):

PR #802, opened 2021-07-23
fix: retry ChunkedEncodingError by default

### LINKED ISSUE #801: ChunkedEncodingError is not retried when fetching data with list_rows()
Original issue: https://github.com/googleapis/python-bigquery-storage/issues/242

A user reported that they saw an error in production when fetching table data with `Client.list_rows()`. That method uses the [default retry object](https://github.com/googleapis/python-bigquery/blob/7e0e2bafc4c3f98a4246100f504fd78a01a28e7d/google/cloud/bigquery/retry.py#L49), which currently does not consider `requests.exceptions.ChunkedEncodingError` retryable.

(it does retry `requests.exceptions.ConnectionError`, but `ChunkedEncodingError` is not a subclass of that.

plamut:
Fixes #801.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/retry.py
+++ b/google/cloud/bigquery/retry.py
@@ -27,6 +27,7 @@ _UNSTRUCTURED_RETRYABLE_TYPES = (
     exceptions.TooManyRequests,
     exceptions.InternalServerError,
     exceptions.BadGateway,
+    requests.exceptions.ChunkedEncodingError,
     requests.exceptions.ConnectionError,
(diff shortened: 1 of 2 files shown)


jimfulton (approved):

tseaver (approved):

tswast (approved):

tswast:
I've filed https://github.com/googleapis/python-api-core/issues/235 to hopefully get this fixed in libraries that don't define their own default retry.

Issue #812, opened 2021-07-26
`to_dataframe` fails with `Decimal type with precision 77 does not fit` when REST API is used to download extreme BIGNUMERIC values


tswast:
Discovered in https://github.com/googleapis/python-bigquery/pull/786, but I think it's unrelated to the dtype changes I'm making there.

#### Steps to reproduce

  1. Create a table with max / min NUMERIC values.

```
{"bool_col": true, "bytes_col": "abcd", "date_col": "9999-12-31", "datetime_col": "9999-12-31 23:59:59.999999", "geography_col": "POINT(-122.0838511 37.3860517)", "int64_col": "9223372036854775807", "numeric_col": "9.9999999999999999999999999999999999999E+28", "bignumeric_col": "5.7896044618658097711785492504343953926634992332820282019728792003956564819967E+38", "float64_col": "+inf", "string_col": "Hello, World", "time_col": "23:59:59.99999", "timestamp_col": "9999-12-31T23:59:59.999999Z"}
{"bool_col": false, "bytes_col": "abcd", "date_col": "0001-01-01", "datetime_col": "0001-01-01 00:00:00", "geography_col": "POINT(-122.0838511 37.3860517)", "int64_col": "-9223372036854775808", "numeric_col": "-9.9999999999999999999999999999999999999E+28", "bignumeric_col": "-5.7896044618658097711785492504343953926634992332820282019728792003956564819968E+38", "float64_col": "-inf", "string_col": "Hello, World", "time_col": "00:00:00", "timestamp_col": "0001-01-01T00:00:00.000000Z"}
{"bool_col": false, "bytes_col": "", "date_col": "1970-01-01", "datetime_col": "1970-01-01 00:00:00", "geography_col": "POINT(0 0)", "int64_col": "0", "numeric_col": "0.0", "bignumeric_col": "0.0", "float64_col": 0.0, "string_col": "", "time_col": "12:00:00", "timestamp_col": "1970-01-01T00:00:00.000000Z"}
{"bool_col": null, "bytes_col": null, "date_col": null, "datetime_col": null, "geography_col": null, "int64_col": null, "numeric_col": null, "bignumeric_col": null, "float64_col": null, "string_col": null, "time_col": null, "timestamp_col": null}
```

  2. Download table with `list_rows`, but set `max_results=10` so that REST API is used.
  3. Call `to_dataframe`

#### Stack trace

```______________ test_list_rows_nullable_scalars_extreme_dtypes[10] ______________

[...]

PR #813, opened 2021-07-26
tests: add system tests for `to_arrow` with extreme values

tswast:

Towards #812  🦕


google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -93,6 +93,8 @@ def pyarrow_numeric():
 
 
 def pyarrow_bignumeric():
+    # 77th digit is partial.
+    # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_type
```
plamut, line 97:
```suggestion
    # https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#decimal_types
```

The anchor in the URL seems incorrect, it just positions you at the top of the page.

tests/data/scalars_extreme.jsonl
```diff
@@ -0,0 +1,4 @@
+{"bool_col": true, "bytes_col": "abcd", "date_col": "9999-12-31", "datetime_col": "9999-12-31 23:59:59.999999", "geography_col": "POINT(-122.0838511 37.3860517)", "int64_col": "9223372036854775807", "numeric_col": "9.9999999999999999999999999999999999999E+28", "bignumeric_col": "9.999999999999999999999999999999999999999999999999999999999999999999999999999E+37", "float64_col": "+inf", "string_col": "Hello, World", "time_col": "23:59:59.99999", "timestamp_col": "9999-12-31T23:59:59.999999Z"}
+{"bool_col": false, "bytes_col": "abcd", "date_col": "0001-01-01", "datetime_col": "0001-01-01 00:00:00", "geography_col": "POINT(-122.0838511 37.3860517)", "int64_col": "-9223372036854775808", "numeric_col": "-9.9999999999999999999999999999999999999E+28", "bignumeric_col": "-9.999999999999999999999999999999999999999999999999999999999999999999999999999E+37", "float64_col": "-inf", "string_col": "Hello, World", "time_col": "00:00:00", "timestamp_col": "0001-01-01T00:00:00.000000Z"}
```
plamut, line 2: Is this POINT really extreme? I would expect to test at least one of the poles and some point just east of Fiji - if it matters?

plamut (commented):
Looks good, and data files are neat - cleaner (IMO at least) than a large wall of parametrized test inputs.

Made two remarks, but that's it.

tests/data/scalars_extreme.jsonl
tswast: Good idea! I've updated it to use the north and south pole. Something along the international date line would be good too, but I'd also like to have a null value. Maybe I'll add another row.

plamut: > Something along the international date line would be good too

Yeah, that's the "near Fiji" point. :)

plamut (approved):
This now looks _extremely_ good. :upside_down_face:

PR #815, opened 2021-07-27
fix: no longer raise a warning in `to_dataframe` if `max_results` set

plamut:
Fixes #806.

That warning should only be used when BQ Storage client is explicitly passed in to `RowIterator` methods when `max_results`
value is also set.

**PR checklist:**
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

PR #822, opened 2021-07-27
chore: sync v3 with main branch

tswast:


tswast:
I wonder if squashing is going to make resolving commits harder in the future? Ugh.

tswast:
Maybe we keep this PR open and periodically add merge commits so that we can resolve conflicts incrementally?

tseaver:
Except for the fact that you just added branch protection to `v3`, it would be simpler to just rebase the `v3` PR against current `master` periodically and force-push it.

plamut:
+1 for merging/rebasing `master` changes periodically, since the merge conflicts will be smaller and easier to handle.

plamut:
Also, I'm not entirely comfortable with syncing changes from `master` without a prior review, since (semantically) incompatible changes can sneak in without warning in the form of a merge conflict ([comment](https://github.com/googleapis/python-bigquery/pull/815/files#r678045054)).

With that in mind it might actually be better to bring in the changes from `master` through merges in a from of PRs that need to be reviewed first. I'll open one such PR shortly.

tests/unit/test_table.py
```diff
@@ -2436,6 +2479,61 @@ def test_to_dataframe_iterable_w_bqstorage(self):
         # Don't close the client if it was passed in.
         bqstorage_client._transport.grpc_channel.close.assert_not_called()
 
+    @unittest.skipIf(pandas is None, "Requires `pandas`")
+    @unittest.skipIf(
+        bigquery_storage is None, "Requires `google-cloud-bigquery-storage`"
+    )
+    @unittest.skipIf(pyarrow is None, "Requires `pyarrow`")
```
plamut, line 2486: This is what I was talking about, i.e. "semantic conflicts". We should be careful to not let these through, since they do not necessarily create loud merge conflicts.

It can get tricky :/

plamut:
> Maybe we keep this PR open and periodically add merge commits so that we can resolve conflicts incrementally?

After resolving several semantic conflicts myself in the related PR, I think it will be easier to _not_ keep a PR open for too long, but instead incrementally accumulate the changes from `master` in the `v3` branch. This will allow any maintainer to open (smaller) PRs against the common `v3` branch, say, after their new non-trivial contributions have been merged into `master`.

For example, I knew I had to pay attention to [redundant skipIf](https://github.com/googleapis/python-bigquery/pull/822#discussion_r678068319) decorators, because it was my code, but might have overlooked semantic conflicts in other people's contributions. The authors probably know best how to bring their changes to the `v3` branch.

tests/unit/test_table.py
tswast: Good catch!

PR #829, opened 2021-07-28
test: refactor `list_rows` tests and add test for scalars

tswast:
I'd like to system test the (to-be-implemented) type conversions for the `INTERVAL` data type, and the "scalars" tests we recently created seemed a good fit for that.

Summary:

* Adds `rowindex` and `interval_col` columns to the "scalars" test data.
* Moves `list_rows()` tests from `test_client.py` to `test_list_rows.py`.
* Excludes `interval_col` from `test_arrow.py` tests. Neither the BigQuery Storage API nor `pyarrow` support INTERVAL data yet.
* Adds system tests to check that all scalar data types are decoded as expected by `list_rows()`.
*  Updates the bytes value in `scalars.jsonl` to actually be a Base64-encoded value, as required by BigQuery.

Towards googleapis/google-cloud-python#15740 🦕


tswast:
P.S. I know I'll want to test the DB-API with the same scalars, though actually since we use the BQ Storage API in the DB-API connector we might not be able to implement the full DB-API support for INTERVAL yet.

tswast:
Test failure is `test_insert_rows_from_dataframe` (https://github.com/googleapis/python-bigquery/issues/733) again. I'm still surprised that it's acting that way, as the streaming buffer should be available for queries. I'll see what I can find out.

tests/data/scalars_extreme.jsonl
```diff
@@ -1,5 +1,5 @@
...
-{"bool_col": false, "bytes_col": "", "date_col": "1970-01-01", "datetime_col": "1970-01-01 00:00:00", "geography_col": "POINT(0 0)", "int64_col": "0", "numeric_col": "0.0", "bignumeric_col": "0.0", "float64_col": 0.0, "string_col": "", "time_col": "12:00:00", "timestamp_col": "1970-01-01T00:00:00.000000Z"}
-{"bool_col": null, "bytes_col": null, "date_col": null, "datetime_col": null, "geography_col": null, "int64_col": null, "numeric_col": null, "bignumeric_col": null, "float64_col": null, "string_col": null, "time_col": null, "timestamp_col": null}
+{"bool_col": true, "bytes_col": "DQo=\n", "date_col": "9999-12-31", "datetime_col": "9999-12-31 23:59:59.999999", "geography_col": "POINT(-135.0000 90.0000)", "int64_col": "9223372036854775807", "interval_col": "P-10000Y0M-3660000DT-87840000H0M0S", "numeric_col": "9.9999999999999999999999999999999999999E+28", "bignumeric_col": "9.999999999999999999999999999999999999999999999999999999999999999999999999999E+37", "float64_col": "+inf", "rowindex": 0, "string_col": "Hello, World", "time_col": "23:59:59.999999", "timestamp_col": "9999-12-31T23:59:59.999999Z"}
```
plamut, line 1: Good catch of a missing nine digit. :+1:

tests/data/scalars_schema.json
```diff
@@ -1,33 +1,32 @@
...
-    "name": "datetime_col",
-    "type": "DATETIME"
+    "name": "date_col",
+    "type": "DATE"
   },
   {
     "mode": "NULLABLE",
-    "name": "bignumeric_col",
-    "type": "BIGNUMERIC"
+    "name": "datetime_col", "type": "DATETIME"
```
plamut, line 24: Didn't black complain about these two dict items on the same line? :thinking:

tests/system/test_arrow.py
```diff
@@ -31,17 +37,35 @@
...
+    scalars_extreme_table: str,
+    max_results: Optional[int],
+    scalars_table_name: str,
 ):
     table_id = scalars_table
     if scalars_table_name == "scalars_extreme_table":
         table_id = scalars_extreme_table
+
+    # Avoid INTERVAL columns until they are supported by the BigQuery Storage
+    # API and pyarrow.
```
plamut, line 51: I would prefix this reminder with a "TODO:", more chance of actually finding it later when searching for these kind of things.

tests/system/test_list_rows.py
```diff
@@ -0,0 +1,120 @@
...
+    page = next(pages)
+    assert page.num_items == num_last_page
+
+
+def test_list_rows_scalars(bigquery_client: bigquery.Client, scalars_table: str):
+    rows = list(
+        sorted(
+            bigquery_client.list_rows(scalars_table), key=lambda row: row["rowindex"]
+        )
+    )
```
plamut, line 61: `sorted()` already returns a list, no need for explicitly "listifying" it.

tests/system/test_list_rows.py
```diff
@@ -0,0 +1,120 @@
...
+
+def test_list_rows_scalars_extreme(
+    bigquery_client: bigquery.Client, scalars_extreme_table: str
+):
+    rows = list(
+        sorted(
+            bigquery_client.list_rows(scalars_extreme_table),
+            key=lambda row: row["rowindex"],
+        )
+    )
```
plamut, line 94: Same here, `sorted()` already returns a list.

tests/system/test_list_rows.py
```diff
@@ -0,0 +1,120 @@
...
+    assert row["bytes_col"] == b"\r\n"
+    assert row["date_col"] == datetime.date(9999, 12, 31)
+    assert row["datetime_col"] == datetime.datetime(9999, 12, 31, 23, 59, 59, 999999)
+    assert row["geography_col"] == "POINT(-135 90)"
+    assert row["int64_col"] == 9223372036854775807
+    assert row["numeric_col"] == decimal.Decimal(
+        "9.9999999999999999999999999999999999999E+28"
+    )
+    assert row["bignumeric_col"] == decimal.Decimal(
+        "9.999999999999999999999999999999999999999999999999999999999999999999999999999E+37"
```
plamut, line 106: (maybe) It's hard to see how many digits are here (and if it's really an extreme value) - maybe rewrite this to make this more clear?

https://github.com/googleapis/python-bigquery/blob/c44d45bc0481aeef2e39ba3392666125bdd2715d/tests/system/test_client.py#L1814

Something like this could work:
```py
f"9.{'9'*75}E+37"
```
It's not a standard way to write down a number, of course, thus decide for yourself if it makes sense.

plamut (commented):
A few minor remarks, but looks good overall.

tests/data/scalars_schema.json
tswast: I don't think we have a linter for JSON files configured, but good catch.

tests/system/test_list_rows.py
tswast: D'oh! I wonder why I learned it as an iterable.

tests/data/scalars_schema.json
plamut: Ah... it's a JSON file, overlooked that when scrolling up and down. :upside_down_face:

plamut (approved):

Issue #707, opened 2021-06-15
Retry did not occurred


nitishxp:
Hi, 

I got the following error in one of my Cloud function
```
, in reraise raise self.last_attempt.result()
File "/opt/python3.8/lib/python3.8/concurrent/futures/_base.py", line 437, in result return self.__get_result()
File "/opt/python3.8/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result raise self._exception
File "/layers/google.python.pip/pip/lib/python3.8/site-packages/tenacity/__init__.py", line 426, in __call__ result = fn(*args, **kwargs)
File "/workspace/cf/functions.py", line 602, in execute_query return self.gbq.query(query).result()
File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/bigquery/job/query.py", line 1172, in result super(QueryJob, self).result(retry=retry, timeout=timeout)
File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/bigquery/job/base.py", line 679, in result return super(_AsyncJob, self).result(timeout=timeout, **kwargs)
File "/layers/google.python.pip/pip/lib/python3.8/site-packages/google/api_core/future/polling.py", line 134, in result raise
  self._exception google.api_core.exceptions.InternalServerError: 500 Job failed because streaming data from one of the tables referenced in the query is temporarily unavailable. Retrying the job may solve the problem
```

tseaver:
@plamut, @tswast PTAL.

plamut:
@nitishxp By default, `InternalServerError` is is listed among the [retriable errors](https://github.com/googleapis/python-bigquery/blob/18bb443c7acd0a75dcb57d9aebe38b2d734ff8c7/google/cloud/bigquery/retry.py#L25-L32). Ditto if the exception instance contains a reason sent by the backend, `"internalError"` is one of them.

Does the error occur often, or was it just a one-time event_

Is your code perhaps using a custom Retry object? (a sample would be great) And which package versions do you have installed? Is there anything else in the logs that might lead to the eventual error posted here?

Any additional info would be useful, thanks!

nitishxp:
@plamut It was a one time exception that occurred in Cloud Function.

Yes I am also using tenacity retry module.

Below are installed packages 

google-cloud-bigquery==2.15.0
google-auth==1.30.0
google-cloud-pubsub==2.2.0
google-cloud-scheduler==2.1.0
google-cloud-storage==1.38.0
[...]

PR #837, opened 2021-07-29
feat: retry failed query jobs in `result()`

### LINKED ISSUE #539: feat: add ability to retry query job automatically if the job fails
#### Environment details

  - OS type and version: Mac (10.15.7) and Python Base Docker Image
  - Python version: `Python 3.7.8`
  - Poetry version: `Poetry version 1.0.10`
  - `google-cloud-bigquery`:  `1.28.0              Google BigQuery API client library`

#### Issue

I am using the `retry` parameter in `biquery.Client.query` but it isn't retrying my request.  

#### Code example

Setup code: 

```python
import structlog
from google.api_core import retry, exceptions
from google.api_core.retry import if_exception_type, Retry

from common.logging.setup_logging import set_logger_config

set_logger_config() # Enables debug level logging for retry

project = "myproject"
source_table = f"{project}.mydataset.TABLE_NOT_REALLY_HERE"
statement = f"SELECT * FROM `{source_table}`"
client = bigquery.Client(project=project)
```

Now if you run: 

```python
retry_policy = retry.Retry(predicate=if_exception_type(exceptions.NotFound))
query_job = client.query(statement, retry=retry_policy)
result = list(query_job.result(retry=retry_policy))
print(result)
```

Only one request is made: 

```
  File "/Users/.../lib/python3.7/site-packages/google/cloud/bigquery/job.py", line 3230, in result
    super(QueryJob, self).result(retry=retry, timeout=timeout)
  File "/Users/.../lib/python3.7/site-packages/google/cloud/bigquery/job.py", line 835, in result
    return super(_AsyncJob, self).result(timeout=timeout)
  File "/Users/.../lib/python3.7/site-packages/google/api_core/future/polling.py", line 134, in result
    raise self._exception
google.api_core.exceptions.NotFound: 404 Not found: Table <TABLE_NAME> was not found in location US

(job ID: 5aeb5865-7bde-4795-ae80-75410f8b9282)

                                              -----Query Job SQL Follows-----                                              

    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |
   1:SELECT *
   2:                FROM `<TABLE_NAME>`
   3:                
    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |    .    |
python-BaseException

Process finished with exit code 1
```

#### Workaround

If instead, I run: 

```python
@Retry(predicate=if_exception_type(exceptions.NotFound))
def do():
    query_job = client.query(statement)
    result = list(query_job.result())
    print(result)
do()
```

This works as expected and retries the request: 

```
{"logger": "google.api_core.retry", "timestamp": "2021-03-02T19:07:48.000317Z", "severity": "debug", "message": "Retrying due to 404 Not found: Table <table_name> was not found in location US\n\n(job ID: <REMOVED>, sleeping 1.0s ..."}
{"logger": "google.api_core.retry", "timestamp": "2021-03-02T19:07:49.363842Z", "severity": "debug", "message": "Retrying due to 404 Not found: Table <table_name> was not found in location US\n\n(job ID: <REMOVED>, sleeping 3.2s ..."}
{"logger": "google.api_core.retry", "timestamp": "2021-03-02T19:07:52.741042Z", "severity": "debug", "message": "Retrying due to 404 Not found: Table <table_name> was not found in location US\n\n(job ID: <REMOVED>, sleeping 2.8s ..."}
{"logger": "google.api_core.retry", "timestamp": "2021-03-02T19:07:55.969272Z", "severity": "debug", "message": "Retrying due to 404 Not found: Table <table_name> was not found in location US\n\n(job ID: <REMOVED>, sleeping 9.2s ..."}
{"logger": "google.api_core.retry", "timestamp": "2021-03-02T19:08:05.493544Z", "severity": "debug", "message": "Retrying due to 404 Not found: Table <table_name> was not found in location US\n\n(job ID: <REMOVED>, sleeping 4.2s ..."}
{"logger": "google.api_core.retry", "timestamp": "2021-03-02T19:08:09.960265Z", "severity": "debug", "message": "Retrying due to 404 Not found: Table <table_name> was not found in location US\n\n(job ID: <REMOVED>, sleeping 44.0s ..."}
```

jimfulton:

Fixes #539  🦕

Previously, we only retried failed API requests. Now, we retry failed jobs (according to the predicate of the `Retry` object passed to `job.result()`).


tests/system/test_query_retry.py
```diff
@@ -0,0 +1,46 @@
+import contextlib
```
tswast, line 1: license headers :-)

google/cloud/bigquery/client.py
```diff
@@ -3233,34 +3233,53 @@ def query(
...
                 return query_job
-        else:
-            return query_job
+
+        future = do_query()
+        # The future might be in a failed state now, but if it's
+        # unrecoverable, we'll find out when we ask for it's result, at which
+        # point, we may retry.
+        if not job_id_given:
+            future.retry_do_query = do_query  # in case we have to retry later
```
tswast, line 3280: Is the type checker happy with this?

google/cloud/bigquery/client.py
```diff
@@ -3233,34 +3233,53 @@ def query(
...
-            if job_id_given:
-                raise create_exc
+        def do_query():
+            # Make a copy now, so that original doesn't get changed by the process
+            # below and to facilitate retry
+            job_config = copy.deepcopy(job_config_save)
+
+            job_id = _make_job_id(job_id_save, job_id_prefix)
+            job_ref = job._JobReference(job_id, project=project, location=location)
+            query_job = job.QueryJob(job_ref, query, client=self, job_config=job_config)
```
tswast, line 3249: It's a bit surprising to me to create a new QueryJob each time. Does this mean that the return value of `result` might not be the same instance?

This could be considered a breaking change if `result` no longer modifies the job in-place.

tests/system/test_query_retry.py
jimfulton: Fixed, here and in unit test. Sorry.

google/cloud/bigquery/client.py
jimfulton: What type checker?

google/cloud/bigquery/client.py
jimfulton: `result()` creates a new job, if necessary, and then replaces the contents of this job with the new one.  So the instance never changes, but it's insides do.

google/cloud/bigquery/client.py
tswast: `pytype` https://github.com/googleapis/python-bigquery/blob/7016f69b6064be101a359bc093ea74fc2a305ac7/noxfile.py#L103-L113

jimfulton: Ah. No pytypes doesn't care.

I added a `_` to that attribute to make it clear that it's an implementation detail.

google/cloud/bigquery/job/query.py
```diff
@@ -1300,12 +1300,47 @@ def result(
...
+                if first:
+                    first = False
+                else:
+                    # Note that we won't get here if retry_do_query is
+                    # None, because we won't use a retry.
+
+                    # The orinal job is failed. Create a new one.
+                    job = retry_do_query()
+
+                    # If it's already failed, we might as well stop:
```
tswast, line 1319: We don't want to stop after a single retry, right? I'd expect there to be a retry object that inspects the exception to see if it's able to reissue the query.

```suggestion
                    # If it's already failed, we might as well let the query retry handle the failure:
```

google/cloud/bigquery/job/query.py
```diff
@@ -1300,12 +1300,47 @@ def result(
                 If the job did not complete in the given timeout.
         """
         try:
-            super(QueryJob, self).result(retry=retry, timeout=timeout)
+            retry_do_query = getattr(self, "_retry_do_query", None)
+            first = True
```
tswast, line 1304: `first` might be unnecessary if we use a `retry_query` object here that catches the exceptions for failed jobs.

https://github.com/googleapis/python-api-core/blob/d566f2893eb708c109ebc3b4fa16f57aa4c65f26/google/api_core/retry.py#L257

Does `super(QueryJob, self).result(retry=retry, timeout=timeout)` raise if the job is already failed from the initial POST request (`_begin`)? I recall we have some logic not to reload the job state / call getQueryResults if the job state is already DONE. -- If not, maybe we should do that, since we document it as such and it'd be weird to be inconsistent with how we raise depending on when the job failed.

tests/system/test_query_retry.py
tswast: Needs to be the Google header in this repo. https://opensource.google/docs/releasing/preparing/#Apache-header

tests/system/test_query_retry.py
```diff
@@ -0,0 +1,65 @@
...
+    job = bigquery_client.query(f"select count(*) from {table_name}")
+    job_id = job.job_id
+
+    # We can already know that the job failed, but we're not supposed
+    # to find out until we call result, which is where retry happend
+    assert job.done()
+    assert job.exception() is not None
+
+    @thread
+    def create_table():
```
tswast, line 52: Neat! Maybe add a comment explaining that the sleep is so that the first query fails.

tests/system/test_query_retry.py
```diff
@@ -0,0 +1,65 @@
...
+        time.sleep(1)
+        with contextlib.closing(google.cloud.bigquery.Client()) as client:
+            client.query(f"create table {table_name} (id int64)")
+
+    retry_policy = Retry(predicate=if_exception_type(exceptions.NotFound))
+    [[count]] = list(job.result(retry=retry_policy))
```
tswast, line 58: Oh, is the `retry` used to determine if we should re-issue a query? I think we should have a separate object for that, since retrying individual API requests can still be useful in cases where we don't want to retry the whole query if its failed.

tests/unit/test_query_retry.py
```diff
@@ -0,0 +1,153 @@
+# Copyright (c) 2021 The PyBigQuery Authors
```
tswast, line 1: Google in this repo: https://opensource.google/docs/releasing/preparing/#Apache-header

tswast (changes requested):

tests/system/test_query_retry.py
jimfulton: OK, I changed it to use `DEFAULT_RETRY` for the 2 sub-requests, however...

There's a problem here.  I don't think users understand retry.  They expect it to deal with transient semantic errors, like NotFound and rate limiting.  They don't get that network errors and application errors are handled separately.  They probably aren't aware of what network errors they should worry about.  It took me a while to grok the difference between errors raised by `api_request` and other errors that cause a job to be in a failed state.  I'm not 100% sure I understand the difference now. 0.5 ;)

IMO, it would **not** help to let them supply two different retries.

Some problems with this PR as it stands ATM:

1. The meaning of the `retry` argument to `result()` depends on whether a job id was supplied to `query`.  This is bad. :)
2. Users probably expect to specify application-level retry when calling `query`, not when calling `result`.

A fundamental problem, IMO, is that there are two kinds of retry and the distinction between them is subtle.

Maybe, a breaking change we could make, since we're making breaking changes :), would be to:

1. Define and use a new retry for network retry. The user doesn't influence this.
2. Use the `retry` argument to `query` and `result` to specify how failed jobs should be retried.
   - Default them to a new sensible default, perhaps limited to rate-limit errors.
   - The retry passed to query is only used as the default retry for `result`.
3. If a user supplies a `job id` to `query`, then error if they pass a `retry` argument to `query` or `result`, to the effect that their job isn't retryable, because the supplied a custom job id. 

I suspect we should brainstorm this a bit.

tests/unit/test_query_retry.py
jimfulton: sorry, fixed

tests/system/test_query_retry.py
tswast: I'd hate for us to retry the query if there was just a transient API error, so I do think we need separate logic.

> A fundamental problem, IMO, is that there are two kinds of retry and the distinction between them is subtle.

I agree. I'm hoping the with good defaults, most users won't have to worry about it. But I also don't think it's something we should hide from them. It's not intuitive and requires some detailed knowledge about how the API works to get right, but I think that's true for custom retry logic in general.

tswast: > The retry passed to query is only used as the default retry for result.

This could work. Especially if we use the `jobs.query` API. https://github.com/googleapis/python-bigquery/issues/589 This is all way too complicated. 😭

tests/system/test_query_retry.py
jimfulton: done

tests/system/test_query_retry.py
jimfulton: > I'd hate for us to retry the query if there was just a transient API error, so I do think we need separate logic.

I'm not suggesting we don't retry API/network errors. I was suggesting we don't give users control over them.

However, in the interest of backward compatibility, I propose:
- Document the existing `retry` argument as API retry. IN the documentation, I'd say something to the effect of "this isn't the retry you're looking for." 
- Add a new `job_retry` argument to `query` and `result` to retry failed jobs. (The `job_retry` passed to `query` is simply the default for `result`.)
- Error if `job_retry` is provided for non-retryable jobs, due to user-suplied job id.

tswast: I like this proposal.

jimfulton: Done, see https://github.com/googleapis/python-bigquery/pull/837/commits/4c6ef5bd48b51c4b676114a46d687158b30fb8c3.

google/cloud/bigquery/client.py
```diff
@@ -3163,6 +3163,7 @@ def query(
         project: str = None,
         retry: retries.Retry = DEFAULT_RETRY,
         timeout: float = None,
+        job_retry: retries.Retry = None,
```
tswast, line 3166: Do we want `None` to be a way to explicitly disable retries? It could be confusing if someone doesn't want jobs to retry for some reason.

google/cloud/bigquery/retry.py
```diff
@@ -56,3 +56,15 @@ def _should_retry(exc):
 on ``DEFAULT_RETRY``. For example, to change the deadline to 30 seconds,
 pass ``retry=bigquery.DEFAULT_RETRY.with_deadline(30)``.
 """
+
+
+@retry.Retry
```
tswast, line 61: Default deadline is only 2 minutes. I expect we'll want more than that. Especially since `jobs.insert` has a 4 minute timeout on the backend.

google/cloud/bigquery/client.py
jimfulton: OK, I'll make None work to disable.

google/cloud/bigquery/retry.py
jimfulton: Do you want to use 4 minutes? 5?

tswast: Let's do 10 minutes since that was the recommendation the backend folks gave me. Should give us time for at least 2 tries retry, possibly 3.

jimfulton: Done

google/cloud/bigquery/client.py
jimfulton: Done

google/cloud/bigquery/retry.py
```diff
@@ -56,3 +58,19 @@ def _should_retry(exc):
...
 pass ``retry=bigquery.DEFAULT_RETRY.with_deadline(30)``.
 """
+
+
+def _job_should_retry(exc):
+    if not hasattr(exc, "errors") or len(exc.errors) == 0:
+        return False
+
+    reason = exc.errors[0]["reason"]
+    return reason == "rateLimitExceeded"
```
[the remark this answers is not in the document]
tswast: Looking at the internal source code: `backendError` is the "reason" corresponding to this job failure. I think we can add that to the default job retry reasons.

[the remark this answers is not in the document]
jimfulton: > this job failure

Do you mean `InternalServerError`?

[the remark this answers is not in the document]
tswast: I think we can disregard the exception type and just look at the `reason` field like you're currently doing.

I mean the kind of job failure that customer encountered in #707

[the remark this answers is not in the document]
jimfulton: done

tswast (approved):
Wonderful!

PR #840, opened 2021-07-29
feat: add support for INTERVAL data type to `list_rows`

tswast:
Implementation of most other features for `INTERVAL` is blocked by backend changes.

Towards googleapis/google-cloud-python#15740 🦕


tswast:
Some of these commits are from https://github.com/googleapis/python-bigquery/pull/829

google/cloud/bigquery/_helpers.py
```diff
@@ -114,6 +126,38 @@ def _int_from_json(value, field):
...
+def _interval_from_json(
+    value: Optional[str], field
+) -> Optional[relativedelta.relativedelta]:
+    """Coerce 'value' to an interval, if set or not nullable."""
+    if not _not_null(value, field):
+        return
+    if value is None:
+        raise TypeError(f"got {value} for REQUIRED field: {repr(field)}")
+
+    parsed = _INTERVAL_PATTERN.match(value)
```
jimfulton, line 138: Check for parsed being `None`.  Something like "Invalid format for interval data, ..." is better that `None has no attribute group`.

google/cloud/bigquery/_helpers.py
```diff
@@ -114,6 +126,38 @@ def _int_from_json(value, field):
...
+    calendar_sign = -1 if parsed.group("calendar_sign") == "-" else 1
+    years = calendar_sign * int(parsed.group("years"))
+    months = calendar_sign * int(parsed.group("months"))
+    days = int(parsed.group("days"))
+    time_sign = -1 if parsed.group("time_sign") == "-" else 1
+    hours = time_sign * int(parsed.group("hours"))
+    minutes = time_sign * int(parsed.group("minutes"))
+    seconds = time_sign * int(parsed.group("seconds"))
+    fraction = parsed.group("fraction")
+    microseconds = time_sign * int(fraction.ljust(6, "0")) if fraction else 0
```
jimfulton, line 148: I can't tell from https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type if there are limits on any of these fields. :shrug:

google/cloud/bigquery/enums.py
```diff
@@ -254,28 +254,29 @@ class SqlTypeNames(str, enum.Enum):
...
     GEOGRAPHY = ScalarQueryParameterType("GEOGRAPHY")
-    TIMESTAMP = ScalarQueryParameterType("TIMESTAMP")
-    DATE = ScalarQueryParameterType("DATE")
+    INT64 = ScalarQueryParameterType("INT64")
+    INTEGER = ScalarQueryParameterType("INT64")
+    NUMERIC = ScalarQueryParameterType("NUMERIC")
+    STRING = ScalarQueryParameterType("STRING")
     TIME = ScalarQueryParameterType("TIME")
-    DATETIME = ScalarQueryParameterType("DATETIME")
+    TIMESTAMP = ScalarQueryParameterType("TIMESTAMP")
```
jimfulton, line 279: Did you intend to include INTERVAL?

google/cloud/bigquery/query.py
```diff
@@ -329,18 +333,41 @@ class ScalarQueryParameter(_AbstractQueryParameter):
...
-            'FLOAT64', 'NUMERIC', 'BIGNUMERIC', 'BOOL', 'TIMESTAMP', 'DATETIME', or
-            'DATE'.
+        type_ (Union[str, google.cloud.bigquery.query.ScalarQueryParameterType]):
+            Name of parameter type. See
+            :class:`google.cloud.bigquery.enums.SqlTypeNames` and
+            :class:`google.cloud.bigquery.enums.SqlParameterScalarTypes` for
+            supported types.
 
-        value (Union[str, int, float, decimal.Decimal, bool, datetime.datetime, datetime.date]):
+        value (Union[ \
```
jimfulton, line 342: Why the backslashes?

google/cloud/bigquery/query.py
```diff
@@ -329,18 +333,41 @@ class ScalarQueryParameter(_AbstractQueryParameter):
...
+                decimal.Decimal,
+                bool,
+                datetime.datetime,
+                datetime.date,
+            ]
+        ],
+    ):
         self.name = name
-        self.type_ = type_
+        if isinstance(type_, ScalarQueryParameterType):
```
jimfulton, line 367: +1 for accepting `ScalarQueryParameterType`

google/cloud/bigquery/query.py
plamut: IIRC, Sphinx (or something else in the docs build) emits a warning about the badly formatted docstring, thus the param type need to be in a single line.

However, by introducing native type hints, we can start getting rid of duplicating the same info in the docstrings.

google/cloud/bigquery/enums.py
tswast: I did but then reverted it because the query parameter support is incomplete. I could revert the alphabetization, but I figure we should do that at some point.

google/cloud/bigquery/query.py
tswast: > However, by introducing native type hints, we can start getting rid of duplicating the same info in the docstrings.

True. I can see how it renders in the docs without these.

google/cloud/bigquery/query.py
tswast: Thanks. I think I'll split this out into its own PR.

tswast: Mailed https://github.com/googleapis/python-bigquery/pull/850 with just this change since it's unrelated to INTERVAL (which actually doesn't support query parameters yet)

google/cloud/bigquery/_helpers.py
tswast: I agree. The docs don't really explain much about how this data format actually works. From what is documented and trial and error, I think these are the limits:

Min

* years: -10000
* months: -11 (I think?)
* days: -3660000
* hours: -87840000
* minutes: -59 (I think?)
* seconds: -59 (I think?)
* microseconds: -999999 (I think?)

Max

* years: 10000
* months: 11 (I think?)
* days: 3660000
* hours: 87840000
* minutes: 59 (I think?)
* seconds: 59 (I think?)
* microseconds: 999999 (I think?)

I don't think we'll need any client-side validation for this, but it does remind me that we should have some unit tests that exercise these limits.

tswast:
I think this is ready, but I'll keep the Do Not Merge label, as I think https://github.com/googleapis/python-bigquery/pull/829 should go in first.

tswast:
Also waiting for https://github.com/googleapis/python-bigquery/pull/829, as I pulled some of the test refactoring changes into that PR.

docs/conf.py
```diff
@@ -360,6 +360,7 @@
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "python": ("https://python.readthedocs.org/en/latest/", None),
+    "dateutil": ("https://dateutil.readthedocs.io/en/latest/", None),
```
tswast, line 363: TODO(tswast): Need to add this to `intersphinx_dependencies` in owlbot config.

tswast: Done: 68035bae2df27283af084eccd6dcc8e6ed7dc36e

tswast:
Ready for review.

google/cloud/bigquery/_helpers.py
```diff
@@ -45,6 +46,15 @@
     re.VERBOSE,
 )
 
+# BigQuery sends INTERVAL data in "canonical format"
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type
+_INTERVAL_PATTERN = re.compile(
+    r"(?P<calendar_sign>-?)(?P<years>[0-9]+)-(?P<months>[0-9]+) "
+    r"(?P<days>-?[0-9]+) "
+    r"(?P<time_sign>-?)(?P<hours>[0-9]+):(?P<minutes>[0-9]+):"
+    r"(?P<seconds>[0-9]+)\.?(?P<fraction>[0-9]+)?$"
```
plamut, line 55: Is a lone dot without fractional digits for microseconds possible? In other words, do we accept `H:M:S.` and interpret the missing digits as zero? (which is what Python does, `123.` is acceptable form of `123.0`).

If not, it might make more sense to treat the `.F` part as an atomic optional part, because in the current from, the regex hints that the time part can also be `H:M:SF`, but of course the seconds part will eat the F part, too, since we don't limit the number of matching digits.

Do we know yet what the limits on the number if digits are? I assume the backend does not allow multiple representations of the same interval? (e.g. 125 seconds cannot be expressed as `0:0:125`, but only as `0:2:5`)
(although `relativedelta` can handle that just fine)

google/cloud/bigquery/_helpers.py
```diff
@@ -180,6 +190,41 @@ def _int_from_json(value, field):
         return int(value)
 
 
+def _interval_from_json(
+    value: Optional[str], field
+) -> Optional[relativedelta.relativedelta]:
+    """Coerce 'value' to an interval, if set or not nullable."""
+    if not _not_null(value, field):
+        return
```
plamut, line 198: (nit) Maybe return `None` explicitly? I almost expected the linter to complain, but it appears it didn't.

owlbot.py
```diff
@@ -98,8 +98,9 @@
     microgenerator=True,
     split_system_tests=True,
     intersphinx_dependencies={
-        "pandas": "http://pandas.pydata.org/pandas-docs/dev",
+        "dateutil": "https://dateutil.readthedocs.io/en/latest/",
         "geopandas": "https://geopandas.org/",
+        "pandas": "http://pandas.pydata.org/pandas-docs/dev",
```
plamut, line 103: (nit, to avoid an unnecessary redirect when visiting the link)
```suggestion
        "pandas": "https://pandas.pydata.org/pandas-docs/dev",
```

setup.py
```diff
@@ -42,6 +42,7 @@
     "google-resumable-media >= 0.6.0, < 3.0dev",
     "packaging >= 14.3",
     "protobuf >= 3.12.0",
+    "python-dateutil >= 2.7.0, <3.0dev",
```
plamut, line 45: Let's maybe use [2.7.2](https://github.com/dateutil/dateutil/releases/tag/2.7.2) as a minimum version, which has been released 14 days after `2.7.0` and includes a fix for the setup script in non-unicode environments? Or the current most recent 2.7.x release with even more fixes and `tzdata` updates?

(didn't test if any of those affect this code, though, just went through the release notes)

tests/unit/helpers/test_from_json.py
```diff
@@ -0,0 +1,134 @@
...
+        ("0-0 0 -0:0:59", relativedelta(seconds=-59)),
+        ("0-0 0 -0:0:1", relativedelta(seconds=-1)),
+        ("0-0 0 0:0:1", relativedelta(seconds=1)),
+        ("0-0 0 0:0:59", relativedelta(seconds=59)),
+        # SELECT (INTERVAL -1 SECOND) / 1000000
+        ("0-0 0 -0:0:0.000001", relativedelta(microseconds=-1)),
+        ("0-0 0 -0:0:59.999999", relativedelta(seconds=-59, microseconds=-999999)),
+        ("0-0 0 -0:0:59.999", relativedelta(seconds=-59, microseconds=-999000)),
+        ("0-0 0 0:0:59.999", relativedelta(seconds=59, microseconds=999000)),
+        ("0-0 0 0:0:59.999999", relativedelta(seconds=59, microseconds=999999)),
```
plamut, line 86: If we want to support the optional microsecond digits, we should also add a test case for that, e.g. `"0-0 0 0:0:59."`

tests/unit/helpers/test_from_json.py
```diff
@@ -0,0 +1,134 @@
...
+                minutes=59,
+                seconds=59,
+                microseconds=999999,
+            ),
+        ),
+        # Test with fraction that is not microseconds.
+        ("0-0 0 0:0:0.1", relativedelta(microseconds=100000)),
+        ("0-0 0 0:0:0.12", relativedelta(microseconds=120000)),
+        ("0-0 0 0:0:0.123", relativedelta(microseconds=123000)),
+        ("0-0 0 0:0:0.1234", relativedelta(microseconds=123400)),
```
plamut, line 129: Note - If we want to support the optional microsecond digits (see my earlier comment at the regex), we should also add a test case for that, e.g. `"0-0 0 0:0:42."`

plamut (commented):
My main question is about how strict we want our parser, similarly to what Jim already brought up. The rest are a few random nits.

google/cloud/bigquery/_helpers.py
jimfulton: Actually, that regex has a bigger problem, `fraction` can be present without the dot -- well not really, but it kind of looks like that at first glance and makes me think too hard :).

I think 

`r"(?P<seconds>[0-9]+)(?P<fraction>[.][0-9]*)?`

would be better.

Or even:

`r"(?P<seconds>[0-9]+(?:[.][0-9]*)?)`

then something like (ignoring the time sign):

```
seconds, microseconds = divmod(float(seconds), 1)
micoseconds = int(micoseconds*1000000)
```

which seems simpler to me than the current logic.

plamut: > Actually, that regex has a bigger problem, `fraction` can be present without the dot -- well not really, but it kind of looks like that at first glance and makes me think too hard :)

Indeed, that was the main point, perhaps I  expressed it in a too convoluted way, sorry. :)

setup.py
tswast: Sounds good. I could see those issues biting us.

google/cloud/bigquery/_helpers.py
```diff
@@ -45,6 +46,15 @@
     re.VERBOSE,
 )
 
+# BigQuery sends INTERVAL data in "canonical format"
+# https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type
+_INTERVAL_PATTERN = re.compile(
+    r"(?P<calendar_sign>-?)(?P<years>[0-9]+)-(?P<months>[0-9]+) "
+    r"(?P<days>-?[0-9]+) "
```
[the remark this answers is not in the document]
tswast: > Edit: I see below in _interval_from_json(), the signs for years/months and time affect multiple parts, while days is a single number and can as such be converted to int directly.

Correct!

google/cloud/bigquery/_helpers.py
tswast: `divmod` looks elegant, but I found a rounding problem when I attempted to use this implementation. :-( https://github.com/googleapis/python-db-dtypes-pandas/issues/18

tswast: >  I assume the backend does not allow multiple representations of the same interval?

When sending data, it seemed to allow it. But when getting data from the API, it always normalized it.

plamut (commented):
Went over this again, it looks fine (just one nit).

Good that we caught the issue in the `divmod`-based implementation, keeping the regex is the right way to go.

plamut (approved):

PR #849, opened 2021-08-03
feat: add support for transaction statistics

### LINKED ISSUE #844: bigquery: support transactioninfo in query statistics
This FR tracks support for the transactionInfo in query statistics, which enables users to correlated jobs to a specific transaction.  Details about this change are shared out of band with library maintainers.

plamut:
Fixes #844.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -93,6 +93,8 @@ class TransactionInfo(typing.NamedTuple):
     """[Alpha] Information of a multi-statement transaction.
 
     https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#TransactionInfo
+
+    .. versionadded:: 2.24.0
     """
 
     transaction_id: str
@@ -354,6 +356,8 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
     @property
     def transaction_info(self) -> Optional[TransactionInfo]:
         """Information of the multi-statement transaction if this job is part of one.
+
+        .. versionadded:: 2.24.0
         """
         info = self._properties.get("statistics", {}).get("transactionInfo")
         if info is None:
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -19,6 +19,7 @@ import copy
 import http
 import threading
 import typing
+from typing import Dict, Optional
 
 from google.api_core import exceptions
 import google.api_core.future.polling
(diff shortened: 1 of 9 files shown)


google/cloud/bigquery/job/query.py
```diff
@@ -1116,6 +1130,16 @@ def dml_stats(self) -> Optional[DmlStats]:
...
 
+    @property
+    def transaction_info(self) -> Optional[TransactionInfo]:
+        """Information of the multi-statement transaction if this job is part of one.
+        """
+        info = self._properties.get("statistics", {}).get("transactionInfo")
+        if info is None:
+            return None
+        else:
+            return TransactionInfo.from_api_repr(info)
```
plamut, line 1141: Should this be part of the base job class?

At the time of writing only Query jobs can initiate transactions, but the stats are returned in the top level stats dict, i.e. not under the "query" sub-dict.

plamut: Answer: Wherever the reservations stats live.

tseaver (approved):

tseaver (approved):

google/cloud/bigquery/job/base.py
```diff
@@ -88,6 +89,22 @@ def _error_result_to_exception(error_result):
 )
 
 
+class TransactionInfo(typing.NamedTuple):
```
tswast, line 92: Does `NamedTuple` make sense here? I suspect we don't want `TransactionInfo` to be a sequence / iterable in our public interface.

Perhaps a data class would be more appropriate. We might have to use the backport for Python 3.6 support. https://pypi.org/project/dataclasses/

tswast: Or we can just use our usual pattern of a private `_properties` and public properties that pull out the relevant fields.

tswast (commented):
Let's make sure we add this class to the docs: https://github.com/googleapis/python-bigquery/blob/master/docs/reference.rst

tswast (changes requested):
Requesting changes, since I just did comment in the last review

google/cloud/bigquery/job/base.py
plamut: I think a (named) tuple very much makes sense here, yes, since it's a read-only property set on the server. A (frozen) dataclass could do the trick as well, but then it's not that much different from a namedtuple. Ditto for a traditional class with `_properties`, same thing but with more code.

Do you still believe we should change the implementation here?

plamut: As discussed offline - we'll go with the namedtuple.

tswast (approved):

PR #850, opened 2021-08-03
feat: support `ScalarQueryParameterType` for `type_` argument in `ScalarQueryParameter` constructor

tswast:
Follow-up to https://github.com/googleapis/python-bigquery/pull/840/files#r679880582

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -110,6 +110,7 @@ language = None
 # directories to ignore when looking for source files.
 exclude_patterns = [
     "_build",
+    "**/.nox/**/*",
     "samples/AUTHORING_GUIDE.md",
     "samples/CONTRIBUTING.md",
     "samples/snippets/README.rst",
--- a/google/cloud/bigquery/enums.py
+++ b/google/cloud/bigquery/enums.py
@@ -191,7 +191,6 @@ _SQL_SCALAR_TYPES = frozenset(
         "DATE",
         "TIME",
         "DATETIME",
-        "INTERVAL",
         "GEOGRAPHY",
         "NUMERIC",
         "BIGNUMERIC",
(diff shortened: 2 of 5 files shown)


google/cloud/bigquery/query.py
```diff
@@ -325,35 +332,46 @@ class ScalarQueryParameter(_AbstractQueryParameter):
     """Named / positional query parameters for scalar values.
 
     Args:
-        name (Optional[str]):
```
tswast, line 328: With the type annotations in the actual method signature, these are unnecessary.

Screen Shot 2021-08-03 at 3 46 15 PM

docs/conf.py
```diff
@@ -110,6 +110,7 @@
 # directories to ignore when looking for source files.
 exclude_patterns = [
     "_build",
+    "**/.nox/**/*",
```
plamut, line 113: Yeah!

google/cloud/bigquery/enums.py
```diff
@@ -259,23 +259,23 @@ class SqlTypeNames(str, enum.Enum):
...
     GEOGRAPHY = ScalarQueryParameterType("GEOGRAPHY")
-    TIMESTAMP = ScalarQueryParameterType("TIMESTAMP")
-    DATE = ScalarQueryParameterType("DATE")
+    INT64 = ScalarQueryParameterType("INT64")
+    INTEGER = ScalarQueryParameterType("INT64")
+    NUMERIC = ScalarQueryParameterType("NUMERIC")
+    STRING = ScalarQueryParameterType("STRING")
     TIME = ScalarQueryParameterType("TIME")
-    DATETIME = ScalarQueryParameterType("DATETIME")
+    TIMESTAMP = ScalarQueryParameterType("TIMESTAMP")
```
plamut, line 278: Isn't INTERVAL too a scalar? Or why don't we include it here?

google/cloud/bigquery/query.py
```diff
@@ -16,14 +16,21 @@
...
 
 from google.cloud.bigquery.table import _parse_schema_resource
 from google.cloud.bigquery._helpers import _rows_from_json
 from google.cloud.bigquery._helpers import _QUERY_PARAMS_FROM_JSON
 from google.cloud.bigquery._helpers import _SCALAR_VALUE_TO_JSON_PARAM
 
 
+_SCALAR_VALUE_TYPES = Optional[
+    Union[str, int, float, decimal.Decimal, bool, datetime.datetime, datetime.date]
+]
```
plamut, line 31: (nit) Shouldn't a singular form be used, since it's a definition of a single type? (that just happens to be an union of multiple types)

plamut (commented):
Looks good, with two remarks that might be (not) relevant.

google/cloud/bigquery/enums.py
tswast: It can't actually at the moment. When I tried in a system test, I got a backend error. The team is aware and fixing it in internal issue 195050789

plamut:
Nevermind, there's only a single change since the last time. Looks good.

plamut (approved):

PR #855, opened 2021-08-06
chore!: remove google.cloud.bigquery_v2 code

### LINKED ISSUE #814: [Breaking Change] Remove `google.cloud.bigquery_v2` directory and related code generation
The `google.cloud.bigquery_v2` modules are frequently out-of-date. This primarily affect the usability of BigQuery ML.

Since BigQuery is a REST API, protobuf changes must be synced manually in internal code repo before they can be published to https://github.com/googleapis/googleapis

In the years since we first introduced `google.cloud.bigquery_v2`, we've seen that BigQuery ML adds model stats and model types much more frequently than the protobuf changes are actually published.

We've worked around this by:

* I and @shollyman (mostly @shollyman lately) republish the protos when a customer has a problem, such as in https://github.com/googleapis/python-bigquery/issues/293
* Avoiding the worst exceptions when an enum or property isn't present, such as in https://github.com/googleapis/python-bigquery/issues/334

**Solution**

* [x] Remove `google.cloud.bigquery_v2` modules
* [x] Update owlbot config to stop generating "client"
* [x] In the BigQuery ML classes, whereever we were returning protobuf object, return the JSON-parsed API response, instead.
  * [x] Complex types become dictionaries.
  * [x] Enums become strings.
* [x] In the routines logic, create manual wrappers and enums where needed.

Supersedes https://github.com/googleapis/python-bigquery/issues/319

plamut:
Closes #814.

The first preview. Still need to address Model types, disable code generation, BigQuery ML classes...

**PR checklist:**


plamut:
This is now ready for review.

It's probably easier to review commit-by-commit, as I made the effort to keep commits mostly self-contained and independent from each other.

kurtisvg (approved):
LGTM for `/samples`.

tswast (changes requested):
Looking good! I think we can simplify this just a tad further.

google/cloud/bigquery/types/__init__.py
```diff
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-# Copyright 2020 Google LLC
+# Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
plamut: Strictly speaking, no. I preserved the structure to avoid unnecessary additional friction when migrating user code from 2x -> 3x. However, it eventually turned out that all that's left in here are the standard SQL types, meaning that the `types` module could indeed be renamed to `standard_sql` - if we want to do it, now is the chance.

User imports will still break, but there's not much difference between renaming `bigquery_v2` to `bigquery` in the import line or creating a slightly different import alias:
```patch
- from google.cloud.bigquery_v2 import types
+ from google.cloud.bigquery import standard_sql as types
```
As an alternative, users can also opt for the plain old find/replace, which requires more or less the same amount of work in both cases.

I think I'll just hoist the `standard_sql` module directly under `bigquery` and that should be it.

google/cloud/bigquery/types/standard_sql.py
```diff
@@ -0,0 +1,224 @@
...
+                    {
+                        name="y",
+                        type={type_kind="ARRAY", array_element_type="DATE"}
+                    }
+                ]
+            }
+        }
+    """
+
+    class TypeKind(str, enum.Enum):
```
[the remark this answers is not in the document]
plamut: We could use plain strings, but they are susceptible to typos, thus I'd rather use `TypeKind.INT64` instead of `"INT64"` in the code, not to mention cases where a (non-existing) `"INTEGER"` string literal is used (actually happened to me a few times).

Presumably the set of available SQL types does not really change often?

owlbot.py
```diff
@@ -24,48 +24,6 @@
 default_version = "v2"
 
 for library in s.get_staging_dirs(default_version):
```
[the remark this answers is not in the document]
plamut: `.kokoro/*.sh` and `CONTRIBUTING.rst`?

As for code files, I also can't think of any right now, except for samples noxfiles if that counts...

google/cloud/bigquery/model.py
```diff
@@ -272,140 +243,120 @@ def encryption_configuration(self):
...
-            this._proto = json_format.ParseDict(
-                resource, types.Model()._pb, ignore_unknown_fields=True
-            )
-        except json_format.ParseError:
-            resource["modelType"] = "MODEL_TYPE_UNSPECIFIED"
-            this._proto = json_format.ParseDict(
-                resource, types.Model()._pb, ignore_unknown_fields=True
-            )
+        # NOTE: No check for invalid model types, unlike before when a protobuf model
+        # did the enum validations - but we don't have that generated enum anymore.
```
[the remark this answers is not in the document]
plamut: Removing then. :)

samples/create_routine.py
```diff
@@ -33,8 +32,8 @@ def create_routine(routine_id):
         arguments=[
             bigquery.RoutineArgument(
                 name="x",
-                data_type=bigquery_v2.types.StandardSqlDataType(
-                    type_kind=bigquery_v2.types.StandardSqlDataType.TypeKind.INT64
+                data_type=bigquery.types.StandardSqlDataType(
```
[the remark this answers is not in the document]
plamut: After moving them to `bigquery/standard_sql.py`, I'll also import them in `bigquery/__initi__.py`. :+1:

google/cloud/bigquery/types/standard_sql.py
```diff
@@ -0,0 +1,224 @@
...
+        def _generate_next_value_(name, start, count, last_values):
+            return name
+
+        TYPE_KIND_UNSPECIFIED = enum.auto()
```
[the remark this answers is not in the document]
plamut: They are string-like values? At least when comparing them to other strings.

Or are you suggesting ditching the enum altogether?

google/cloud/bigquery/types/standard_sql.py
```diff
@@ -78,10 +83,61 @@ def __post_init__(self):
...
+                struct_type = self.struct_type.to_api_repr()
+            result["structType"] = struct_type
+
+        return result
+
+    @classmethod
+    def from_api_repr(cls, resource: Dict[str, Any]):
+        """Construct an SQL data type instance given its API representation."""
+        type_kind = resource.get("typeKind")
+        if type_kind not in cls.TypeKind.__members__:
```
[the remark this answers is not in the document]
plamut: What if this method is used by someone manually constructing a resource dict and converting it to one of the standard SQL types? Probably not a typical use case, though.

I suppose we cannot prevent all such errors anyway, and if a user ends up with invalid SQL type instances and sends that to the backend, it should be up to them to understand the server's error message and fix the code accordingly?

Also, if we decide to ditch the `TypeKind` enum altogether, we won't have a choice anyway.

google/cloud/bigquery/types/standard_sql.py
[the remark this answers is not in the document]
plamut: Another thought - if we decide to keep the `TypeKind` enum, I would unnest it. Being defined under the `StandardSqlDataType`, it makes the code that uses it unnecessarily verbose.

plamut:
The snippets check failure seems to be caused by service account keys expiring, we need to use a different service account for CI... I will bring the changes from `main` to the `v3` branch, should work.

owlbot.py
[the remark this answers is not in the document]
tswast: > .kokoro/*.sh and CONTRIBUTING.rst?

Those are added by `py_library`, not the `get_staging_dirs` loop.

https://github.com/googleapis/python-bigquery/blob/afdc1edf36914a512ef5c738efcb1e11055d903f/owlbot.py#L95

tswast:
@googlebot I consent.

google/cloud/bigquery/model.py
```diff
@@ -66,140 +62,122 @@ def __init__(self, model_ref):
...
-        """Union[datetime.datetime, None]: Datetime at which the model was
-        created (:data:`None` until set from the server).
+    def created(self) -> Optional[datetime.datetime]:
+        """Datetime at which the model was created (:data:`None` until set from the server).
 
         Read-only.
         """
-        value = self._proto.creation_time
+        value = self._properties.get("creationTime")
         if value is not None and value != 0:
```
tswast: I don't think we need to check for `value != 0` anymore. I believe we added this a side-effect of protocol buffers v3 behavior with optional fields -- they'd always return the default value.

google/cloud/bigquery/model.py
```diff
@@ -272,140 +243,117 @@ def encryption_configuration(self):
...
-        # Keep a reference to the resource as a workaround to find unknown
-        # field values.
+
+        resource = copy.deepcopy(resource)
         this._properties = resource
 
         # Convert from millis-from-epoch to timestamp well-known type.
         # TODO: Remove this hack once CL 238585470 hits prod.
-        resource = copy.deepcopy(resource)
         for training_run in resource.get("trainingRuns", ()):
```
tswast: Regardless of whether the original change to well-known types is in production (it should be as that change was submitted in Mar 15, 2019), I think we should remove this. The goal is not to transform the data for fields such as `trainingRuns` and just give the developer exactly what the API returns.

google/cloud/bigquery/standard_sql.py
```diff
@@ -0,0 +1,207 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import Any, Dict, Iterable, Optional
+
+from google.cloud.bigquery.enums import StandardSqlTypeNames
+
+
+@dataclass
```
tswast, line 21: After seeing how these classes are used elsewhere, I think the StandardSql classes are worth going with our `_properties` pattern.

There's a particular feature that can support, which is that folks can modify the `_properties` dictionary directly when they are testing out a feature in private preview before we add support for it in the client libraries.

tswast (changes requested):

samples/tests/conftest.py
```diff
@@ -125,8 +124,8 @@ def routine_id(client, dataset_id):
     routine.arguments = [
         bigquery.RoutineArgument(
             name="x",
-            data_type=bigquery_v2.types.StandardSqlDataType(
-                type_kind=bigquery_v2.types.StandardSqlDataType.TypeKind.INT64
+            data_type=bigquery.StandardSqlDataType(
```
tswast, line 127: Let's make sure we mention these changes to Standard SQL types and enums in the migration guide. https://github.com/googleapis/python-bigquery/issues/960

plamut: Definitely

google/cloud/bigquery/enums.py
```diff
@@ -256,28 +221,6 @@ class SqlTypeNames(str, enum.Enum):
     DATETIME = "DATETIME"
 
 
-class SqlParameterScalarTypes:
```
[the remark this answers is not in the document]
plamut: Indeed, it was causing too many problems, failing tests, etc., and moving it out of the enums was the most efficient solution. As a rule of thumb, since enums are (in)directly imported almost everywhere, enums themselves should not import much.

google/cloud/bigquery/standard_sql.py
plamut: Even if not the purest solution, _practicality beats purity_ - if users put `_properties` to good use (even at their own risk), then let's preserve that. I'll refactor the standard SQL classes implementation.

google/cloud/bigquery/schema.py
```diff
@@ -308,9 +308,14 @@ def to_standard_sql(self) -> standard_sql.StandardSqlField:
...
             if array_element_type == StandardSqlTypeNames.STRUCT:  # noqa: E721
-                sql_type.array_element_type.struct_type = standard_sql.StandardSqlStructType(
+                new_struct_type = standard_sql.StandardSqlStructType(
                     fields=(field.to_standard_sql() for field in self.fields)
                 )
+                arr_element_typ = sql_type.array_element_type
+                arr_element_typ.struct_type = new_struct_type
+                sql_type.array_element_type = arr_element_typ
+                # TODO: should work like this, gotcha!!!
+                # sql_type.array_element_type.struct_type = new_struct_type
```
plamut, line 318: This probably needs to be fixed to avoid unpleasant surprises.

Changing an attribute of a subtype is not reflected on the parent containing that subtype, because _retrieving_ a subtype returns a _new_ subtype instance (constructed from parent's `_properties`). Parent types should therefore reference subtype instances directly and only (de)serialize them to/from API representation when explicitly requested (in `to_api_repr()` and `from_api_repr()`).

Also, the coverage check could also fail, a lot of corner cases. :)

google/cloud/bigquery/standard_sql.py
```diff
@@ -43,52 +42,137 @@ class StandardSqlDataType:
...
+                raise ValueError(
+                    "Cannot change {StandardSqlTypeNames.ARRAY} type_kind, first set "
+                    "array_element_type attribute to None."
+                )
+
+            if current == StandardSqlTypeNames.STRUCT and self.struct_type.fields:
+                raise ValueError(
+                    "Cannot change {StandardSqlTypeNames.STRUCT} type_kind, first set "
+                    "struct_type attribute to None."
+                )
```
plamut, line 108: @tswast There's some logic to assure the internal consistency of the `StandardSqlDataType` class, i.e. making sure that `type_kind`, `array_element_type`, and `struct_type` attributes have compatible values.

If you think we should ditch that entirely and treat SQL type classes as merely thin wrappers around the JSON dicts exchanged with the backend, please let me know.

google/cloud/bigquery/schema.py
tswast: Definitely worth fixing. I wonder if we avoid copies in `to_api_repr()` and `from_api_repr()` if this would work as expected?

google/cloud/bigquery/standard_sql.py
tswast: I'm in favor of removing client-side validations. There's too much of a risk that we get out-of-sync with the server, otherwise.

google/cloud/bigquery/schema.py
plamut: The problem is that getters construct new objects based on what they find in `self._properties`, and once that is returned to the user, changes to those subtype instances are not reflected on the original `self._properties` dict (the parent type will not see the changes).

We could directly assigning a subdict from `_properties` as `_properties` of the child instance, but that would be fragile, e.g. multiple instances of a subtype would share the same `_properties` dict (yikes!).

Or did I misunderstand the idea?

It's probably the most straightforward if we keep the state in SQL type objects and only convert to/from `_properties` dicts when needed. Users that want to directly manipulate the `_properties` dict could still do so, they would just need to re-construct the instance:
```py
foo_type._properties["bar"] = "new_bar"
foo_type = foo_type.from_api_repr(foo_type._properties)
```
Would that be acceptable?

After all that's messing with internals, and a user consciously doing that should know that the changes are not automatically reflected on the manipulated instance.

tswast: > We could directly assigning a subdict from _properties as _properties of the child instance, but that would be fragile, e.g. multiple instances of a subtype would share the same _properties dict (yikes!).

You mean that

```python
instance1 = sql_type.array_element_type
instance2 = sql_type.array_element_type
```

would share a `_properties` instance? I'm not seeing a problem with that. In fact, that seems desirable, as it's not super intuitive that `array_element_type` would construct a new instance anyway, which is what we're trying to fix, right?

google/cloud/bigquery/standard_sql.py
plamut: OK, I'll remove the validation checks and early warnings completely.

google/cloud/bigquery/schema.py
plamut: That's what I meant with that, yes. Any change to either of them will be reflected on the other, as well as on the parent instance that shares a subset of its `_properties` dict with them.

Although `None` values could still cause problems, as they can be currently _removed_ from `_properties`, I need to see how that behaves. TBH, I'm not entirely comfortable with such state sharing, but I'll have another look tomorrow and see what the tests will say.

tswast: Sounds good.

I still think you're getting spooked by the fact that it's a `@property`.

If this were just a normal instance attribute, state sharing would be expected.

```
class MyClass:
    def __init__(self):
        self.my_state = {}

parent = MyClass()
instance1 = parent.my_state
instance2 = parent.my_state
```

I would expect instance1 and instance2 to point at the same object.

plamut: > I would expect instance1 and instance2 to point at the same object.

Me too, but I would not expect them to point to two different instances that point to the same underlying `_properties` dict. :)

It feels more natural to think in terms of objects and references to sub-objects, while the dict we currently store as `_properties` is essentially just a (de)serialization format for exchanging data with the outside world (the backend API in this case).

I can give the shared `_properties` a shot and see how it goes with cases when we set an attribute to `None`, which can actually remove a sub-dict from `_properties` as opposed to manipulating the former's contents.

plamut: Good news, it turned out that this was not an issue after all. If a sub-property on the parent type is removed from the latter's`_properties`, it just means that any sub-instances that were created from that sub-property simply get "disconnected" from the parent, which is expected behavior.

plamut:
@tswast Pushed the implementation changes early for preview, but there are now only a few coverage misses left, thus this is close to done.

docs/conf.py
```diff
@@ -364,8 +363,6 @@
     "google-auth": ("https://googleapis.dev/python/google-auth/latest/", None),
     "google.api_core": ("https://googleapis.dev/python/google-api-core/latest/", None,),
     "grpc": ("https://grpc.github.io/grpc/python/", None),
-    "proto-plus": ("https://proto-plus-python.readthedocs.io/en/latest/", None),
```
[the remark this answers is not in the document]
tswast: Edit: Nevermind. I see you edited owlbot. We are probably going to want to revert this once we add the proto classes back (just static copies, no longer generated) for migration, though.

setup.py
```diff
@@ -29,20 +29,19 @@
 # 'Development Status :: 5 - Production/Stable'
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
+    "dataclasses >= 0.8, < 1.0; python_version < '3.7'",  # Until we drop Python 3.6 support.
```
tswast, line 32: Do we still need `dataclasses`? I don't see it used anywehre.

tswast (commented):
LGTM once coverage is there, with some small nits re: setup.py and owlbot.

setup.py
plamut: Nope, you're right. After refactoring the models again, `dataclasses` dependency is not needed anymore.

google/cloud/bigquery/model.py
```diff
@@ -66,140 +61,122 @@ def __init__(self, model_ref):
...
         """
-        return self._proto.model_type
+        return self._properties.get("modelType", "MODEL_TYPE_UNSPECIFIED")
 
     @property
-    def training_runs(self):
-        """Sequence[google.cloud.bigquery_v2.types.Model.TrainingRun]: Information
-        for all training runs in increasing order of start time.
+    def training_runs(self) -> Sequence[Dict[str, Any]]:
+        """Information for all training runs in increasing order of start time.
```
tswast, line 150: Let's make sure we link to the REST API docs.

```suggestion
        """Information for all training runs in increasing order of start time.

        Dictionaries are in REST API format. See:
        https://cloud.google.com/bigquery/docs/reference/rest/v2/models#trainingrun
```

tswast (approved):

tswast:
Thanks!

PR #865, opened 2021-08-09
fix: make unicode characters working well in load_table_from_json

### LINKED ISSUE #864: some emoji will become wrong character stored in bigquery while using load_table_from_json

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-bigquery/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS type and version: macOS Big Sur
  - Python version: `python --version` 3.8.8
  - pip version: `pip --version`: 21.0.1
  - `google-cloud-bigquery` version: `pip show google-cloud-bigquery`: 2.6.1

#### Steps to reproduce

1. insert any emoji which is over `0xFFFF` into a `dict`, e.g. 🏦 (`\U0001F3E6`)
2. use `client.load_table_from_json(rows_to_insert...` to upload
3. see that emoji icon become �� in BigQuery web console

#### Code example

```python
    bank = "\U0001F3E6" 
    rows_to_insert = [{"emoji": bank}]
    client.load_table_from_json(rows_to_insert, table_id, job_config=job_config)
```

#### Stack trace
```
# example
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

grimmerk:

Fixes #864 🦕

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -7777,7 +7777,6 @@ class TestClientUpload(object):
 
     def test_load_table_from_json_unicode_emoji_data_case(self):
         from google.cloud.bigquery.client import _DEFAULT_NUM_RETRIES
-        from google.cloud.bigquery import job
 
         client = self._make_client()
 
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -7775,6 +7775,43 @@ class TestClientUpload(object):
         err_msg = str(exc.value)
         assert "Expected an instance of LoadJobConfig" in err_msg
 
+    def test_load_table_from_json_unicode_emoji_data_case(self):
+        from google.cloud.bigquery.client import _DEFAULT_NUM_RETRIES
+        from google.cloud.bigquery import job
+
+        client = self._make_client()
+
+        json_row = {"emoji": "\U0001F3E6"}
+        json_rows = [
+            json_row
+        ]
+
+        load_patch = mock.patch(
+            "google.cloud.bigquery.client.Client.load_table_from_file", autospec=True
+        )
+
+        with load_patch as load_table_from_file:
+            client.load_table_from_json(json_rows, self.TABLE_REF)
+
+        load_table_from_file.assert_called_once_with(
+            client,
+            mock.ANY,
+            self.TABLE_REF,
+            size=mock.ANY,
+            num_retries=_DEFAULT_NUM_RETRIES,
+            job_id=mock.ANY,
+            job_id_prefix=None,
+            location=client.location,
+            project=client.project,
+            job_config=mock.ANY,
+            timeout=None,
+        )
+
+        sent_data_file = load_table_from_file.mock_calls[0][1][1]
+
+        # make sure json_row's unicode characters are only encoded one time
+        assert sent_data_file.getvalue() == json.dumps(json_row, ensure_ascii=False).encode()
+
     # Low-level tests
 
     @classmethod
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -7782,9 +7782,7 @@ class TestClientUpload(object):
 
         emoji = "\U0001F3E6"
         json_row = {"emoji": emoji}
-        json_rows = [
-            json_row
-        ]
+        json_rows = [json_row]
 
         load_patch = mock.patch(
             "google.cloud.bigquery.client.Client.load_table_from_file", autospec=True
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -7781,7 +7781,8 @@ class TestClientUpload(object):
 
         client = self._make_client()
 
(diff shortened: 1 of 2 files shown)


tseaver (changes requested):
@grimmer0125 Thanks very much for the patch!  Can you please add a corresponding unit test to `tests/unit/test_client.py`, exercising the bug?  You can likely just copy the `test_load_table_from_json_basic_use` test to use they example data you have in #864, and change the `assert_called_once_with` assertion to ensure that the emoji data is encoded properly.

jimfulton (dismissed):
Thanks!

Can you add a test?

I think a unit test that verifies that the file passed to `load_table_from_file` (mocked) is correct would be pretty easy to write (if you're familiar with unit testing).

grimmer0125:
@tseaver, @jimfulton

Thank you for the reviews. I have pushed the added unit test. To avoid comparing io.BytesIO object issue, I made the assertion of emoji data after using `assert_called_once_with`.

tests/unit/test_client.py
```diff
@@ -7775,6 +7775,43 @@ def test_load_table_from_json_w_invalid_job_config(self):
         err_msg = str(exc.value)
         assert "Expected an instance of LoadJobConfig" in err_msg
 
+    def test_load_table_from_json_unicode_emoji_data_case(self):
+        from google.cloud.bigquery.client import _DEFAULT_NUM_RETRIES
+        from google.cloud.bigquery import job
+
+        client = self._make_client()
+
+        json_row = {"emoji": "\U0001F3E6"}
```
tseaver, line 7784: Can you make the emoji a constant, so that we can use it directly in an assertion below?  E.g.:
```suggestion
        emoji = "\U0001F3E6"
        json_row = {"emoji": emoji}
```

tests/unit/test_client.py
```diff
@@ -7775,6 +7775,43 @@ def test_load_table_from_json_w_invalid_job_config(self):
...
+            location=client.location,
+            project=client.project,
+            job_config=mock.ANY,
+            timeout=None,
+        )
+
+        sent_data_file = load_table_from_file.mock_calls[0][1][1]
+
+        # make sure json_row's unicode characters are only encoded one time
+        assert sent_data_file.getvalue() == json.dumps(json_row, ensure_ascii=False).encode()
```
tseaver, line 7813: It would be cleaner to assert that the emoji is present, rather than using the same logic as the test.  E.g.:
```suggestion
        sent_data_file.seek(0)  # rewind
        sent_row = json.load(sent_data_file)
        assert sent_row["emoji"] == emoji
```

tseaver (changes requested):

tests/unit/test_client.py
tseaver: Hmmm, my proposed change here is defective, because `json.load()` returns the correct value whether or not the `sent_data_file` was generated with `ensure_ascii=False`.

tseaver: So, for a better choice, let's format the bytes we actually expect to see on the wire.
```python
        expected_bytes = b'{"emoji": "' + emoji.encode("utf8") + b'"}'
        assert sent_data_file.getvalue() == expected_bytes
```

grimmer0125: Thank you. It makes sense. Originally I thought of it this way but somehow I did not. I have pushed a new commit for this.

tswast (approved):
Thank you for the contribution!

tests/unit/test_client.py
```diff
@@ -7775,6 +7775,45 @@ def test_load_table_from_json_w_invalid_job_config(self):
         err_msg = str(exc.value)
         assert "Expected an instance of LoadJobConfig" in err_msg
 
+    def test_load_table_from_json_unicode_emoji_data_case(self):
+        from google.cloud.bigquery.client import _DEFAULT_NUM_RETRIES
+        from google.cloud.bigquery import job
```
tseaver, line 7780: The linter is complaining about this unused import:
```suggestion
```

tseaver (approved):

tests/unit/test_client.py
```diff
@@ -7775,6 +7775,44 @@ def test_load_table_from_json_w_invalid_job_config(self):
...
+    def test_load_table_from_json_unicode_emoji_data_case(self):
+        from google.cloud.bigquery.client import _DEFAULT_NUM_RETRIES
+
+        client = self._make_client()
+
+        emoji = "\U0001F3E6"
+        json_row = {"emoji": emoji}
+        json_rows = [
+            json_row
+        ]
```
tseaver, line 7787:
```suggestion
        json_rows = [json_row]
```

tseaver (approved):

PR #867, opened 2021-08-10
feat: make the same `Table*` instances equal to each other

### LINKED ISSUE #854: Table instances created with the same arguments do NOT evaluate as equals
Two table objects with the same arguments do not evaluate as equal. This makes unit testing and comparing calls to `Client.insert_rows` more complicated than necessary.

#### Environment details

  - OS type and version: Fedora 32
  - Python version: `3.9.5`
  - pip version: `21.1.3`
  - `google-cloud-bigquery` version: `2.23.2`

#### Steps to reproduce

  1. Create table with set of args
  2. Create separate instance of table with same set of args
  3. Test equality `t1 == t2`
  4. returns false

#### Code example

```python
from google.cloud.bigquery import Table
t1 = Table('project.dataset.table')
t2 = Table('project.dataset.table')
t1 == t2
# returns: False
```

plamut:
Closes #854.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1011,6 +1011,13 @@ class Table(object):
         """Generate a resource for ``update``."""
         return _helpers._build_resource_from_properties(self, filter_fields)
 
+    def __eq__(self, other):
+        if not isinstance(other, Table):
+            return NotImplemented
+        return self._properties == other._properties
+
+    __hash__ = None
+
     def __repr__(self):
         return "Table({})".format(repr(self.reference))
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1014,7 +1014,7 @@ class Table(object):
     def __eq__(self, other):
         if not isinstance(other, Table):
             return NotImplemented
-        return self.path == other.path
+        return self._properties["tableReference"] == other._properties["tableReference"]
 
     __hash__ = None
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1016,7 +1016,8 @@ class Table(object):
             return NotImplemented
         return self._properties["tableReference"] == other._properties["tableReference"]
 
-    __hash__ = None
+    def __hash__(self):
+        return hash((self.project, self.dataset_id, self.table_id))
 
     def __repr__(self):
         return "Table({})".format(repr(self.reference))
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1014,7 +1014,7 @@ class Table(object):
     def __eq__(self, other):
         if not isinstance(other, Table):
             return NotImplemented
-        return self._properties == other._properties
+        return self.path == other.path
 
     __hash__ = None
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -255,9 +255,16 @@ class TableReference(object):
         return (self._project, self._dataset_id, self._table_id)
 
     def __eq__(self, other):
-        if not isinstance(other, TableReference):
+        if isinstance(other, (Table, TableListItem)):
+            return (
+                self.project == other.project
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/table.py
```diff
@@ -1011,6 +1011,13 @@ def _build_resource(self, filter_fields):
         """Generate a resource for ``update``."""
         return _helpers._build_resource_from_properties(self, filter_fields)
 
+    def __eq__(self, other):
+        if not isinstance(other, Table):
+            return NotImplemented
+        return self._properties == other._properties
```
jimfulton, line 1017: I think this is overly stringent.

```
>>> from google.cloud.bigquery import table
>>> t1 = table.Table("precise-truck-742.riversnake.lake")
>>> t2 = client.get_table("precise-truck-742.riversnake.lake")
>>> t1.__class__, t2.__class__
(<class 'google.cloud.bigquery.table.Table'>, <class 'google.cloud.bigquery.table.Table'>)
>>> t1._properties == t2._properties
False
```

`t1` and `t2` refer to the same table, and should be equal IMO.  They only differ by meta data that is dynamic. `t2` has extra data that is effectively cached.

I think you should just compare the table reference.

jimfulton (changes requested):

google/cloud/bigquery/table.py
plamut: I was also considering that option, yes. And now you even made a convincing case with `t2` fetched by the client, I also think it should be equal to `t1`.

google/cloud/bigquery/table.py
```diff
@@ -1011,6 +1011,13 @@ def _build_resource(self, filter_fields):
         """Generate a resource for ``update``."""
         return _helpers._build_resource_from_properties(self, filter_fields)
 
+    def __eq__(self, other):
+        if not isinstance(other, Table):
+            return NotImplemented
+        return self.path == other.path
```
jimfulton, line 1017: This is correct, but involves a lot of computation.

Is there a reason you can't just compare the `tableReference` property?

plamut: I'm not too concerned about that overhead, and the path seemed just the right property to compare, short and clean. But I'm fine with either way, I'll change.

jimfulton:
I wonder if we should implement `__hash__` while we're at this.  That would allow us to have sets of tables and even use tables as `dict` keys.

plamut:
> I wonder if we should implement `__hash__` while we're at this. That would allow us to have sets of tables and even use tables as `dict` keys.

Hm... now that comparison key has changed and is immutable, `Table` instances, despite being mutable, can indeed be hashable. I'll give it a thought.

google/cloud/bigquery/table.py
```diff
@@ -1011,6 +1011,13 @@ def _build_resource(self, filter_fields):
         """Generate a resource for ``update``."""
         return _helpers._build_resource_from_properties(self, filter_fields)
 
+    def __eq__(self, other):
+        if not isinstance(other, Table):
```
tswast, line 1015: Should we allow TableReference and TableListItem as well?

plamut: Hm... since we only care about table reference and ignore other metadata properties, and because `Table` and `TableReference` can already be used interchangeably as an argument to many methods, this could actually make sense.

jimfulton (approved):

google/cloud/bigquery/table.py
plamut: Done.

However, I now see quite a lot of repetition - all these `Table*` classes have the same `__hash__()` definition, the equality comparison is very similar, and they all have the same core properties - `project`, `dataset_id`, `table_id`.

This seems like a good refactoring opportunity - I would extract this into a common base class. What do you think?

tswast: Yes, a common base class would make sense. I think there are a few other methods / properties that are duplicated besides `__hash__`. Let's do that in another PR just for the refactoring.

tswast (approved):

PR #872, opened 2021-08-11
chore: merge changes from master

tswast:

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -255,9 +255,16 @@ class TableReference(object):
         return (self._project, self._dataset_id, self._table_id)
 
     def __eq__(self, other):
-        if not isinstance(other, TableReference):
+        if isinstance(other, (Table, TableListItem)):
+            return (
+                self.project == other.project
+                and self.dataset_id == other.dataset_id
+                and self.table_id == other.table_id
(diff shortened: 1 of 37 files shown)


tseaver (approved):

google/cloud/bigquery/table.py
```diff
@@ -1690,6 +1728,8 @@ def to_arrow(
         """
         self._maybe_warn_max_results(bqstorage_client)
 
+        self._maybe_warn_max_results(bqstorage_client)
+
```
plamut, line 1732: IIRC I fixed that recently, but it might have been in the `v3` branch. Anyhow, the duplicate call should be removed.
```suggestion
```

Hm... but I can't seem to suggest a completely empty replacement, thus we'll end up with two blank lines and the linter will likely complain... manual change on the PR branch might be needed.

plamut (changes requested):
Found a duplicate warning call, the rest seems good. As mentioned in the comment, it might not be possible to fix it through a GitHub suggestion.

google/cloud/bigquery/table.py
tswast: Good eye! Just pushed a fix.

plamut (approved):

PR #875, opened 2021-08-12
fix: remove pytz dependency and require pyarrow>=3.0.0

busunkim96:
For https://github.com/googleapis/python-api-core/issues/250

One of the samples uses `pytz` for timezones. I was not sure how datetimes work with pandas so I left it alone: https://github.com/googleapis/python-bigquery/blob/master/samples/load_table_dataframe.py

Since pandas requires `pytz` it should continue to be runnable. https://github.com/pandas-dev/pandas/blob/d037ff6a4757bf8af2ca2431ba7d4b22b1959075/setup.cfg#L32-L35


tswast (approved):
Excellent. Thanks!

plamut:
Ugh, the test failure seems legitimate?

```
>   ???
E   pyarrow.lib.ArrowInvalid: Casting from timestamp[us, tz=UTC] to timestamp[ns] would result in out of bounds timestamp: 81953424000000000

pyarrow/error.pxi:84: ArrowInvalid
```

tseaver (approved):

tseaver:
I don't really understand the `unit-3.6` failure here:  running with `--pdb` indicates that `timestamp_as_object` is False, which likely means that the cast of the "out-of-bounds" value did *not* raise an exception:

```python
(Pdb) pp list(record_batch)
[<pyarrow.lib.ChunkedArray object at 0x7f58e35bca98>
[
  [
    4567-01-01 00:00:00.000000,
    9999-12-31 00:00:00.000000
  ]
]]
(Pdb) pp record_batch[0][0].cast("timestamp[ns]")
<pyarrow.TimestampScalar: Timestamp('2228-10-14 01:41:45.161793536')>
(Pdb) pp record_batch[0][1].cast("timestamp[ns]")
<pyarrow.TimestampScalar: Timestamp('1816-03-29 05:56:08.066277376')>
```

tswast:
Interesting finding. Perhaps this was a bug in `pyarrow` if `pytz` is not installed?

What this test is trying to do is get some timestamps that are supported in BigQuery but outside of the valid range in pandas. In this case we need to fallback to using `datetime.datetime` objects. This is what the `timestamp_as_object` parameter is intended to do. We have some logic here to determine if we should be asking for `timestamp_as_object`:

https://github.com/googleapis/python-bigquery/blob/cd21df1716e9ab163c779a716d94a850a6b2d253/google/cloud/bigquery/table.py#L1975-L1983

tswast:
There were some fixes to pyarrow Timestamps in 2.0 and 3.0. Since we plan to require pyarrow 3.0.0+ soon (in v3 branch), let's bump our minimum version to 3.0.0 while it's still in extras in this v2 branch.

busunkim96:
Thanks @tswast! pyarrow==3.0.0 seems to fix the unit test on a local run.

tswast:
And cover failed... I forgot we had some version check for pyarrow < 3.0: https://github.com/googleapis/python-bigquery/blob/cd21df1716e9ab163c779a716d94a850a6b2d253/google/cloud/bigquery/_pandas_helpers.py#L153-L160

tswast:
@googlebot I consent

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -146,23 +145,15 @@ def pyarrow_timestamp():
...
-        BQ_TO_ARROW_SCALARS["BIGNUMERIC"] = pyarrow_bignumeric
         # The exact decimal's scale and precision are not important, as only
         # the type ID matters, and it's the same for all decimal256 instances.
-        ARROW_SCALAR_IDS_TO_BQ[pyarrow.decimal256(76, scale=38).id] = "BIGNUMERIC"
-        _BIGNUMERIC_SUPPORT = True
-    else:
-        _BIGNUMERIC_SUPPORT = False
+        pyarrow.decimal128(38, scale=9).id: "NUMERIC",
+        pyarrow.decimal256(76, scale=38).id: "BIGNUMERIC",
+    }
```
tseaver, line 152: W00t!

PR #891, opened 2021-08-20
feat: Destination tables are no-longer removed by create_job

### LINKED ISSUE #483: Query jobs destinationTable attribute is explicitly removed

jimfulton:

Fixes #483 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -1937,7 +1937,6 @@ class Client(ClientWithProject):
             )
         elif "query" in job_config:
             copy_config = copy.deepcopy(job_config)
-            _del_sub_prop(copy_config, ["query", "destinationTable"])
             query_job_config = google.cloud.bigquery.job.QueryJobConfig.from_api_repr(
                 copy_config
             )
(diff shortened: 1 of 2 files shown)


plamut (approved):
Trivial lint error aside (unused import), this looks good.

google/cloud/bigquery/client.py
```diff
@@ -1937,7 +1936,6 @@ def create_job(
             )
         elif "query" in job_config:
             copy_config = copy.deepcopy(job_config)
-            _del_sub_prop(copy_config, ["query", "destinationTable"])
```
tseaver, line 1940: Given that we are no longer removing the property, we could just collapse this whole clause into:

```python
        elif "query" in job_config:
            query_job_config = google.cloud.bigquery.job.QueryJobConfig.from_api_repr(
                job_config
            )
```

jimfulton: Good idea. https://github.com/googleapis/python-bigquery/pull/893

PR #892, opened 2021-08-23
fix: use REST API in cell magic when requested

### LINKED ISSUE #876: --use_rest_api still seems to use storage api
#### Environment details
Ubuntu 18.04.5 LTS

Python 3.7.10

pip 21.1.3

google-cloud-bigquery              2.20.0
google-cloud-bigquery-datatransfer 3.2.0
google-cloud-bigquery-storage      2.4.0

#### Steps to reproduce

  1. Use %%bigquery magic with the --use_rest_api flag on a project where I don't bigquery.readsession permissions. 
  2. Observe that the query fails due to not having readsession permissions. 

#### Code example

```python
%%bigquery my_test --use_rest_api
SELECT
*
FROM `aaa-willyn-test.test_dataset.my_test`
```

#### Stack trace
```
PermissionDenied                          Traceback (most recent call last)
<ipython-input-9-f594b2ed9ec4> in <module>
----> 1 get_ipython().run_cell_magic('bigquery', 'my_test --use_rest_api', 'SELECT\n*\nFROM `aaa-willyn-test.test_dataset.abi_copy`\n')

/opt/conda/lib/python3.7/site-packages/IPython/core/interactiveshell.py in run_cell_magic(self, magic_name, line, cell)
   2401             with self.builtin_trap:
   2402                 args = (magic_arg_s, cell)
-> 2403                 result = fn(*args, **kwargs)
   2404             return result
   2405 

...skipping several frames here...

/opt/conda/lib/python3.7/site-packages/google/api_core/grpc_helpers.py in error_remapped_callable(*args, **kwargs)
     67             return callable_(*args, **kwargs)
     68         except grpc.RpcError as exc:
---> 69             six.raise_from(exceptions.from_grpc_error(exc), exc)
     70 
     71     return error_remapped_callable

/opt/conda/lib/python3.7/site-packages/six.py in raise_from(value, from_value)

PermissionDenied: 403 request failed: the user does not have 'bigquery.readsessions.create' permission for 'projects/fc-product-demo'
```

plamut:
Fixes #876.

The `--use_rest_api` option did not work as expected and this commit fixes it.

**PR checklist:**


google/cloud/bigquery/magics/magics.py
```diff
@@ -671,7 +671,10 @@ def _cell_magic(line, query):
                 _handle_error(ex, args.destination_var)
                 return
 
-            result = rows.to_dataframe(bqstorage_client=bqstorage_client)
+            result = rows.to_dataframe(
+                bqstorage_client=bqstorage_client,
+                create_bqstorage_client=use_bqstorage_api,
```
tswast, line 676: Since we already construct a `bqstorage_client` (I believe so we can use a customer user-agent), can't we just always set `create_bqstorage_client` to `False`?

plamut: Hm... if we don't want to use the BQ Storage API, we set it to `False`, and if we do, we pass in a BQ Storage client instance ourselves, meaning that the value is ignored anyway.

Yeah, always passing `False` is more simple.

plamut: FWIW, I left the `mock.ANY` assertion in some of the tests, since that's still technically correct (the value is ignored if a BQ Storage client instance is explicitly provided).

tswast (approved):

PR #893, opened 2021-08-23
chore: Simplify create_job slightly

jimfulton:

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -1935,9 +1935,8 @@ class Client(ClientWithProject):
                 source_type=source_type,
             )
         elif "query" in job_config:
-            copy_config = copy.deepcopy(job_config)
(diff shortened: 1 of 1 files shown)


google/cloud/bigquery/client.py
```diff
@@ -1935,9 +1935,8 @@ def create_job(
                 source_type=source_type,
             )
         elif "query" in job_config:
-            copy_config = copy.deepcopy(job_config)
             query_job_config = google.cloud.bigquery.job.QueryJobConfig.from_api_repr(
-                copy_config
+                copy.deepcopy(job_config)
```
tseaver: Are we making a copy because we expect `from_api_repr` to mutate the dict?

jimfulton: No, I made a copy because I didn't pay close enough attention to your suggestion. :)

Fixed.

tseaver (approved):

PR #910, opened 2021-08-25
chore: migrate default branch to main

### LINKED ISSUE #909: Migrate from master to main branch
As part of the umbrella issue googleapis/google-cloud-python#10579, we need to switch the default branch from `master` to `main`. Also, all occurrences of `master` should be renamed to `main` (except in cases where URLs could be broken, because the migration has not happened yet).

plamut:
Fixes #909.

Probably needs a few extra replacement rules in `owlbot.py`.


jimfulton (commented):
I don't see changes in any .kokoro files or in CONTRIBUTING.rst.

I still see `master` references there.

Is the idea that owlbot will fix those later?

parthea (approved):

parthea (approved):

PR #921, opened 2021-08-27
feat: set the X-Server-Timeout header when timeout is set

### LINKED ISSUE #919: populate X-Server-Timeout when `timeout` is set in `jobs.insert` requests
https://cloud.google.com/apis/docs/system-parameters#definitions

While investigating internal issue 195337762, it was determined that some internal servers can get stuck waiting for longer than is necessary. It is believed that if `X-Server-Timeout` is set, then those internal servers can stop processing more quickly.

Whenever `timeout` is set in the JSONConnection call API method, also populate the `X-Server-Timeout` header.

It's possible the local client-side timeout may be sufficient.

 **Additional context**

It's still not 100% clear to me that this header will help. A Googler on the internal issue 195337762 promises to share some docs on how this header bubbles through the system.

jimfulton:

Fixes #919 🦕

### DIFF
--- a/tests/unit/helpers.py
+++ b/tests/unit/helpers.py
@@ -18,6 +18,38 @@ import mock
 import pytest
 
 
+def add_header_assertion_to_kwargs(kwargs):
+    timeout = kwargs.get("timeout")
+    if timeout is not None:
+        headers = kwargs.setdefault("headers", {})
+        if headers is None:
+            kwargs["headers"] = headers = {}
+        headers[google.cloud.bigquery.client.TIMEOUT_HEADER] = str(kwargs["timeout"])
+
+    return kwargs
+
+
+def add_header_assertion(mock, name):
+    """
+    Modify assert_called_with-ish assertions to add timeout headers
+
+    if there's a timeout
+    """
+    orig = getattr(mock, name)
+
+    def repl(*args, **kw):
+        return orig(*args, **add_header_assertion_to_kwargs(kw))
+
+    setattr(mock, name, repl)
+
+
+def api_call(*args, **kw):
+    """
+    Replacement for mock.call that adds a timeout header, if necessary
+    """
+    return mock.call(*args, **add_header_assertion_to_kwargs(kw))
+
+
 def make_connection(*responses):
     import google.cloud.bigquery._http
     import mock
@@ -26,6 +58,9 @@ def make_connection(*responses):
     mock_conn = mock.create_autospec(google.cloud.bigquery._http.Connection)
     mock_conn.user_agent = "testing 1.2.3"
     mock_conn.api_request.side_effect = list(responses) + [NotFound("miss")]
+    for name in "assert_called_with", "assert_called_once_with":
+        add_header_assertion(mock_conn.api_request, name)
+
     mock_conn.API_BASE_URL = "https://bigquery.googleapis.com"
     mock_conn.get_api_base_url_for_mtls = mock.Mock(return_value=mock_conn.API_BASE_URL)
     return mock_conn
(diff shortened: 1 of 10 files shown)


jimfulton:
For reviewers:

The change here is very simple: 
https://github.com/googleapis/python-bigquery/pull/921/files#diff-fc3199dcfb64ea8758d31d37f30dc25c4885b1c21b03812fbc0f1914f48763f3R752-R762

If timeout is set, also set the `X-Server-Timeout` header.

**What's not simple is the impact on testing.**  

Many tests make assertions about HTTP requests and this change invalidated most of those assertions.  I would argue that the assertions were too broad, exposing them to cross-cutting concerns, like this one.  I would also argue that the assertions were pragmatic. :)

To mitigate the impact on existing test assertions, I did 2 things:

- Add a new `api_call` API to replace `mock.call`.  This takes expected calls and adds `X-Server-Timeout` headers to the expected calls if the expected calls include timeouts.
- All but one test use `make_connection` (including job tests that has their own version and didn't need to).  `make_connection` sets up the mock used to make assertions about HTTP calls.  I modified `assert_called_with` and `assert_called_once_with` on this mock to normalize expectations in the same way as `api_call` above.

tests/unit/helpers.py
```diff
@@ -18,6 +18,38 @@
 import pytest
 
 
+def add_header_assertion_to_kwargs(kwargs):
```
tswast, line 21: My initial reaction to this is duplicating too much logic from the actual implementation and encouraging more "change detector tests". The need for it probably indicates we've been too strict on our header assertions to begin with.

Note to self: any additional thoughts after seeing the rest of the test updates?

tswast: This is a clever solution. I think my initial reaction is still correct, but it's not worth it to rip out our (probably too low level) `api_request` call assertions for something less specific.

If we were writing these tests for the first time, I think separate assertions for specific headers would be desired.

tswast (changes requested):
Could you add a few unit tests specifically for `_call_api` so we have tests documenting the header behavior without relying on your `api_call` and `add_header_assertion_to_kwargs` helpers?

tests/unit/helpers.py
tswast: Not only specific header assertions, but a separate assertion for path, a separate assertion for body, etc.

jimfulton:
> Could you add a few unit tests specifically for `_call_api` so we have tests documenting the header behavior without relying on your `api_call` and `add_header_assertion_to_kwargs` helpers?

Sure.

tests/unit/helpers.py
jimfulton: Here's another idea, perhaps benefiting from a weekend's distance. :)

Abstract the addition of the `X-Server-Timeout` header into a function.  Add an autouse test fixture that replaces that function with a noop except on the tests that test addition of the header, by adding a special marker to those tests and and checking for the marker in the autouse fixture.

(https://stackoverflow.com/questions/38748257/disable-autouse-fixtures-on-specific-pytest-marks)

jimfulton: @tswast shall I? :)

tswast: That does sound preferable if it means we can remove the logic that checks for timeout in our tests.

tswast:
@jimfulton Did you mean to close this PR?

jimfulton:
> @jimfulton Did you mean to close this PR?

Yes. I'm going to make a new one.  We can reopen if we change our minds. :)

jimfulton:
See #927

PR #927, opened 2021-08-31
feat: set the X-Server-Timeout header when timeout is set

### LINKED ISSUE #919: populate X-Server-Timeout when `timeout` is set in `jobs.insert` requests

jimfulton:

Fixes #919 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -131,6 +131,8 @@ _MIN_GET_QUERY_RESULTS_TIMEOUT = 120
 # https://github.com/googleapis/python-bigquery/issues/781#issuecomment-883497414
 _PYARROW_BAD_VERSIONS = frozenset([packaging.version.Version("2.0.0")])
 
+TIMEOUT_HEADER = "X-Server-Timeout"
+
 
 class Project(object):
     """Wrapper for resource describing a BigQuery project.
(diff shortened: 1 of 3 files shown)


jimfulton:
Reviewers note:

This PR adds the `X-Server-Timeout` header when a non-`None` `timeout` parameter is passed to `_call_api`.

To avoid breaking many test assertions, we use an auto-use text fixture to disable this behavior.

For tests that test this behavior, we use a mark that disables the behavior disabling in the auto-use fixture. :)

Another small change in this PR is that we don't pass falsey headers to the underlying transport.  This affected a few test assertions, which have been updated.

tswast:
Re: test failure: https://source.cloud.google.com/results/invocations/473c748f-bcb0-467f-8d54-96cfc13eb268/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fsystem-3.8/log

I don't think this change should have affected `QueryJob.transaction_info`. Unfortunately I don't have a Job ID, so I don't think it'll be possible to investigate further.

tswast (approved):

PR #932, opened 2021-09-01
test: fix routine DDL sample test exits too early

### LINKED ISSUE #931: A test for create routine DDL sample makes (almost) no assertions
[This test](https://github.com/googleapis/python-bigquery/blob/fcb0bc68c972c2c98bb8542f54e9228308177ecb/samples/tests/test_routine_samples.py#L27-L35) returns too early and thus only one of the assertions is actually executed.

plamut:
Fixes #931.

**PR checklist:**


parthea (approved):

tseaver (approved):

tseaver:
@plamut I'm puzzled why coverage didn't catch this, since the lines after that return statment are unreachable.

tswast (approved):
Thanks!

PR #946, opened 2021-09-06
fix: Arrow extension-type metadata was not set when calling the REST API or when there are no rows

### LINKED ISSUE #894: feat: ensure Arrow type metadata is consistent between REST and BQ Storage APIs
The BQ Storage API annotates certain types with additional information so that they can be disambiguated. For example, to tell GEOGRAPHY from STRING columns.

We don't currently annotate the types we create when data is initially downloaded from the REST API.

If we make these consistent, the user could register Arrow extension types, which can hook into the default pyarrow -> pandas conversions. See discussion on: https://github.com/googleapis/python-bigquery/pull/848#issuecomment-904025863

jimfulton:

Fixes #894  🦕

### DIFF
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -173,6 +173,13 @@ if pyarrow:
         pyarrow.decimal128(38, scale=9).id: "NUMERIC",
         pyarrow.decimal256(76, scale=38).id: "BIGNUMERIC",
     }
+    BQ_FIELD_TYPE_TO_ARROW_FIELD_METADATA = {
+        "GEOGRAPHY": {
+            b"ARROW:extension:name": b"google:sqlType:geography",
+            b"ARROW:extension:metadata": b'{"encoding": "WKT"}',
+        },
+        "DATETIME": {b"ARROW:extension:name": b"google:sqlType:datetime"},
+    }
 
 else:  # pragma: NO COVER
     BQ_TO_ARROW_SCALARS = {}  # pragma: NO COVER
@@ -227,7 +234,12 @@ def bq_to_arrow_field(bq_field, array_type=None):
         if array_type is not None:
             arrow_type = array_type  # For GEOGRAPHY, at least initially
         is_nullable = bq_field.mode.upper() == "NULLABLE"
-        return pyarrow.field(bq_field.name, arrow_type, nullable=is_nullable)
+        metadata = BQ_FIELD_TYPE_TO_ARROW_FIELD_METADATA.get(
+            bq_field.field_type.upper() if bq_field.field_type else ""
+        )
+        return pyarrow.field(
+            bq_field.name, arrow_type, nullable=is_nullable, metadata=metadata
+        )
 
     warnings.warn("Unable to determine type for field '{}'.".format(bq_field.name))
     return None
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1810,7 +1810,7 @@ class RowIterator(HTTPIterator):
             if owns_bqstorage_client:
                 bqstorage_client._transport.grpc_channel.close()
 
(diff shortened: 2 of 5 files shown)


google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -227,7 +234,12 @@ def bq_to_arrow_field(bq_field, array_type=None):
         if array_type is not None:
             arrow_type = array_type  # For GEOGRAPHY, at least initially
         is_nullable = bq_field.mode.upper() == "NULLABLE"
-        return pyarrow.field(bq_field.name, arrow_type, nullable=is_nullable)
+        metadata = BQ_FIELD_TYPE_TO_ARROW_FIELD_METADATA.get(
+            bq_field.field_type.upper() if bq_field.field_type else ""
```
plamut, line 238: Looking at [pyarrow.field](https://arrow.apache.org/docs/python/generated/pyarrow.field.html#pyarrow.field) docs, `metadata` should be `dict` or `None`, but we sometimes use an empty string. Is there a reason for that?

I suppose it still works, but if nothing else a type checker might start complaining at some point when it realizes that.

google/cloud/bigquery/table.py
```diff
@@ -1810,7 +1810,7 @@ def to_arrow(
             if owns_bqstorage_client:
                 bqstorage_client._transport.grpc_channel.close()
 
-        if record_batches:
+        if record_batches and bqstorage_client is not None:
```
plamut, line 1813: Curious what bug did this extra check for the BQ Storage client fix?

tests/system/test_arrow.py
```diff
@@ -110,3 +110,56 @@ def test_list_rows_nullable_scalars_dtypes(
...
+    dataset_client.query(f"create table {test_table_name} ({columns})").result()
+    if do_insert:
+        names = list(types)
+        values = ", ".join(types[name][1] for name in names)
+        names = ", ".join(names)
+        dataset_client.query(
+            f"insert into {test_table_name} ({names}) values ({values})"
+        ).result()
+    at = dataset_client.query(f"select * from {test_table_name}").result().to_arrow()
+    smd = {at.field(i).name: at.field(i).metadata for i in range(at.num_columns)}
```
plamut, line 152: (nit)
What does `smd` stand for, Stored metadata? And `rmd` is retrieved metadata? Or is it Storage- and REST- metadata, respectively?

Had some trouble deciphering that, maybe find a better name? :)

plamut (commented):
A few nits/questions, but generally looks good to me.

google/cloud/bigquery/_pandas_helpers.py
jimfulton: The empty string is just the argument to `get`. It's not metadata.  We only pass a `dict` or `None`.

google/cloud/bigquery/table.py
jimfulton: When `bqstorage_client` is `None`, we're using the REST API, which doesn't add arrow extension metadata.

I improved the comment.

tests/system/test_arrow.py
jimfulton: > (nit)
> What does `smd` stand for, Stored metadata? And `rmd` is retrieved metadata? Or is it Storage- and REST- metadata, respectively?

The later.

> Had some trouble deciphering that, maybe find a better name? :)

Good idea. Done.

google/cloud/bigquery/_pandas_helpers.py
plamut: Meh, of course, and an empty string won't mach any keys, resulting in `None`. And I though I saw a comma somewhere making `""` a second argument. Nevermind. :)

plamut (approved):
LGTM!

tswast (commented):
Thanks

PR #947, opened 2021-09-06
chore: sync v3 branch with main

plamut:
Bringing the `v3` branch back to sync with `main`. There were quite a few merge conflicts, thus proceed carefully. :lady_beetle: 

Should also fix the snippets test failure in #855.


tseaver:
@plamut Is the target branch for this PR supposed to be `v3`?

plamut:
@tseaver Yes, and thanks for noticing this bummer.

noxfile.py
```diff
@@ -94,9 +94,16 @@ def unit(session):
...
-@nox.session(python=UNIT_TEST_PYTHON_VERSIONS[-1])
+@nox.session(python=[UNIT_TEST_PYTHON_VERSIONS[0], UNIT_TEST_PYTHON_VERSIONS[-1]])
 def unit_noextras(session):
     """Run the unit test suite."""
+
+    # Install optional dependencies that are out-of-date.
+    # https://github.com/googleapis/python-bigquery/issues/933
+    # There is no pyarrow 1.0.0 package for Python 3.9.
+    if session.python == UNIT_TEST_PYTHON_VERSIONS[0]:
+        session.install("pyarrow==1.0.0")
```
tswast, line 105: Seems like you can manually revert the changes to `unit_noextras` too since pyarrow is a required dependency now.

tswast (approved):
That's quite the diff! Other than a suggestion for `noxfile.py`, this looks good to me.

tswast:
Adding `Do Not Merge` just so you see the `noxfile.py` suggestion. Once you've addressed that, this looks good to merge.

google/cloud/bigquery/_helpers.py
```diff
@@ -74,7 +74,31 @@ def is_read_session_optional(self) -> bool:
         return self.installed_version >= _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION
 
 
+class PyarrowVersions:
```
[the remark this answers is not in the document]
plamut: Yes, I manually deleted that while fixing the merge conflicts, as we indeed don't need it anymore (`pyarrow` becoming required in `v3`).

noxfile.py
plamut: Thanks for spotting that. Yes, it's indeed not needed.

plamut:
@tswast Thanks for going through this.

The diff was large, indeed, thus we should probably do these syncs more frequently (I'm thinking at least once a week). Even just fixing the merge conflicts took quite some time, not to mention that those are prone to accidentally introducing bugs.

tswast:
I triggered a rebuild of the `Kokoro prerelease-deps` job and filed https://github.com/googleapis/python-bigquery/issues/958 to allow us to have enough info to investigate these flakes further.

PR #956, opened 2021-09-09
refactor: extract common logic to TableBase class

### LINKED ISSUE #871: Extract common features of `Table*` classes to a new base class
In #854 `Table*` instances were made equal to each other if they referenced the same table.

During the implementation it turned out that `Table`, `TableReference`, and `TableListItem` have several nearly identical parts that could be placed in a new base class. For example:
 - The `project`, `table_id`, and `dataset_id` properties.
 - The `__hash__()` and `__eq__()` special methods.

Extract these and any other features into a new common base class.

plamut:
Closes #871.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -69,6 +69,7 @@ if typing.TYPE_CHECKING:  # pragma: NO COVER
     import geopandas
     import pyarrow
     from google.cloud import bigquery_storage
+    from google.cloud.bigquery.dataset import DatasetReference
 
 
 _NO_PANDAS_ERROR = (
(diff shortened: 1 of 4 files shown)


google/cloud/bigquery/table.py
```diff
@@ -234,9 +283,9 @@ def to_api_repr(self) -> dict:
             Dict[str, object]: Table reference represented as an API resource
         """
         return {
-            "projectId": self._project,
-            "datasetId": self._dataset_id,
-            "tableId": self._table_id,
+            "projectId": self.project,
+            "datasetId": self.dataset_id,
+            "tableId": self.table_id,
```
tseaver, line 288: Shouldn't this be a `copy.deepcopy` of `self._properties`?

tests/unit/test_table.py
```diff
@@ -102,6 +102,21 @@ def test_ctor_with_key(self):
...
+        return TableBase
+
+    def _make_one(self, *args, **kw):
+        return self._get_target_class()(*args, **kw)
+
+    def test_ctor_defaults(self):
+        instance = self._make_one()
+        assert instance._properties == {}
+
+
```
tseaver, line 119: ISTM we need direct tests of the shared properties (`project`, `dataset_id`, `table_id`, `path`) and dunder-methods (`__eq__`, `__hash__`).

plamut: These are already tested through the individual Table* classes' tests, the common logic has just been hoisted into a base class. Or do you feel there's still value in additional tests specifically targeting `TableBase`?

tseaver: I'd rather have explicit tests tied to the implementatoin class.  If the derived classes no longer have the methods, then they shouldn't need to test them.

tswast: I'm ambivalent about this. In some ways it's nice just to test the "public" interface (subclasses). That way if we refactor again, code coverage can let us know when there's dead code in the base class.

That said, maybe we want `TableBase` to be public to simplify some of our `Union` types in the client?

google/cloud/bigquery/table.py
```diff
@@ -126,45 +127,93 @@ def _view_use_legacy_sql_getter(table):
         return True
 
 
-class TableReference(object):
-    """TableReferences are pointers to tables.
+class TableBase:
```
tswast, line 130: Let's either make this private (`_TableBase`) or document it + update relevant types in the `client` module.

tswast (changes requested):

google/cloud/bigquery/table.py
plamut: I don't see the end users using `TableBase` directly, it just serves as a central place where we keep the common functionality. An implementation detail, in other words.

Let's mark it as private.

tests/unit/test_table.py
plamut: I'd say the dilemma stems from the fact that the base class was created after the now-subclasses, resulting in subclasses having dedicated tests, and no explicit base class tests.

If it was the other way around and the base class would be added at the same time as the other `Table*` classes, I'm pretty sure we would not be writing several copies of tests (one per subclass) for the common properties.

Taking the same point to the extreme, all classes inherit from `object`, but we don't actually test things such as their default `__str_()` or `__hash__()` behavior, unless our class explicitly overrides the default implementation.

I thus lean more towards refactoring the tests as well.

google/cloud/bigquery/table.py
tseaver: The contrarian point is that parameter types might use a public `TableBase` to signal the minimal contract that their arguments require, which is more focused than `Union[Table, TableRef, ...]`.  I don't see any case where `TableBase` as a return type make sense.

google/cloud/bigquery/table.py
plamut: Fun fact - switching to the deepcopy implementation revealed a bug in the constructor. A string that was passed to the `_set_sub_prop()` helper resulted in a dysfunctional `_properties`  dict: `{"k": {"e": {"y": "value"}}}`, as that helper expects a sequence.

google/cloud/bigquery/table.py
plamut: After some thought I'm leaning towards keeping the base class private if the main reason for making it public is cleaner type annotations. I view type annotations more as an optional helper, and not as an end goal that would require adjusting the code just for the annotations' sake. As a bonus we still have an option to define our own `TableType` alias if we don't like `Union[Table, ...]`.

Or does anybody have a strong contrary opinion here?

tswast (approved):

PR #959, opened 2021-09-09
feat: add pandas date and time arrays to hold BigQuery DATE and TIME data

### LINKED ISSUE #862: convert time columns to dbtime by default in `to_dataframe`
Currently TIME columns are just exposed as string objects. This would be a better experience and align with better with the expectations for working with timeseries in pandas https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html

Presumably one could combine a date column with a time column to create a datetime by adding them.

jimfulton:
- [] Ensure the tests and linter pass
- [] Code coverage does not decrease (if any source code was changed)
- [] Appropriate docs were updated (if necessary)

Fixes #862 🦕


google/cloud/bigquery/dtypes.py
```diff
@@ -0,0 +1,247 @@
...
+    _epoch = datetime.datetime(1970, 1, 1)
+
+    @classmethod
+    def _datetime(cls, scalar):
+        if isinstance(scalar, datetime.time):
+            return datetime.datetime.combine(cls._epoch, scalar)
+        elif isinstance(scalar, str):
+            # iso string
+            h, m, s = map(float, scalar.split(":"))
+            s, us = divmod(s, 1)
```
tswast, line 199: Mod 1? Isn't that just "s"?

I notice in INTERVAL that BigQuery doesn't always return the full 6 digits.

https://github.com/googleapis/python-bigquery/pull/840/files#diff-c8c849e0dbfe70fa3d5124f55d379f53fe4caff056addea0b981f679857bcfd5R82-R86

In fact, sometimes it doesn't have any partial seconds at all.

Maybe we should put some of the time parsing in a shared `_helpers` module to be used by both INTERVAL and TIME? https://github.com/googleapis/python-bigquery/pull/840/files#diff-f06352fa086344932b4922958dc31556b5cb0be634f640e2e009063955d1ca95R215

jimfulton: If `s` is a float, `divmod(s, 1)` gives the whole and fractional parts.  We then multiply the fractional part by 1,000,000 to get the number of microseconds.  It doesn't matter how many digits are after the dot.

jimfulton: > Maybe we should put some of the time parsing in a shared `_helpers` module to be used by both INTERVAL and TIME?

Sure.

jimfulton:
Gaaa, I should have based this on the V3 branch.

Cllosing to rebase.

google/cloud/bigquery/dtypes.py
jimfulton: > > Maybe we should put some of the time parsing in a shared `_helpers` module to be used by both INTERVAL and TIME?
> 
> Sure.

Well, :)

- We're talking about 2 lines of code.
- Given that we might want to split out the dtype implementations, I think it's best to keep the dtype code decoupled.

PR #961, opened 2021-09-10
test: ensure prerelease versions of pandas and arrow are tested nightly

### LINKED ISSUE #882: Add "continuous" job for prerelease deps tests
Follow-up to https://github.com/googleapis/python-bigquery/issues/95

We should have a continuous (nightly) job that tests with presubmits. We usually have PRs open, but it'd be good to have timely notifications from flakey bot if something happens when we aren't making a PR.

Also, we should double-check that as we add pyarrow and such to required dependencies that we're still able to test with pre-release versions. It'd be good to ignore our pins so that we know when a dependency makes a breaking change.

tswast:

Fixes #882 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -250,15 +250,23 @@ def prerelease_deps(session):
         "pytest-cov",
     )
 
+    # Because we test minimum dependency versions on the minimum Python
+    # version, the first version we test with in the unit tests sessions has a
+    # constraints file containing all dependencies and extras.
     with open(
-        CURRENT_DIRECTORY / "testing" / "constraints-3.6.txt", encoding="utf-8"
+        CURRENT_DIRECTORY
+        / "testing"
+        / f"constraints-{UNIT_TEST_PYTHON_VERSIONS[0]}.txt",
+        encoding="utf-8",
     ) as constraints_file:
         constraints_text = constraints_file.read()
-    constraints_text = re.sub(r"==\S+", "\n", constraints_text)
+
+    # Ignore leading whitespace and comment lines.
     deps = [
-        line.strip()
-        for line in constraints_text.split("\n")
-        if line.strip() and line[0] != "#"
+        match.group(1)
+        for match in re.finditer(
+            r"^\s*(\S+)(?===\S+)", constraints_text, flags=re.MULTILINE
+        )
     ]
 
     # We use --no-deps to ensure that pre-release versions aren't overwritten
--- a/noxfile.py
+++ b/noxfile.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import
 
 import pathlib
 import os
+import re
 import shutil
 
 import nox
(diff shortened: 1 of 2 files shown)


tswast:
Local output, verifying expected package versions are actually used now:

```
nox > python -m pip install --no-deps -e .[all]
nox > python -c import grpc; print(grpc.__version__)
1.40.0
nox > python -c import pandas; print(pandas.__version__)
1.4.0.dev0+639.g00e10a5c48
nox > python -c import pyarrow; print(pyarrow.__version__)
6.0.0.dev226
```

tswast:
Internal change 396006532 pending review to actually enable this continuous job.

noxfile.py
```diff
@@ -234,7 +249,22 @@ def prerelease_deps(session):
...
+
+    with open(
+        CURRENT_DIRECTORY / "testing" / "constraints-3.6.txt", encoding="utf-8"
+    ) as constraints_file:
+        constraints_text = constraints_file.read()
+    constraints_text = re.sub(r"==\S+", "\n", constraints_text)
+    deps = [
+        line.strip()
+        for line in constraints_text.split("\n")
+        if line.strip() and line[0] != "#"
```
plamut, line 261: What if we somehow end up with a comment line that starts with whitespace? Unlikely, indeed, but it would feel clumsy if we accidentally break the pipeline with redundant whitespace in a text file. :laughing: 

BTW, should we instead match the entire lines containing version pins and directly extract the dependecy name with grouping in the regex pattern itself? Feels more straightforward, to me at lest.

plamut (commented):
Looks good, just had a comment about possibly improving/simplifying pattern matching.

noxfile.py
```diff
@@ -234,7 +249,22 @@ def prerelease_deps(session):
         "pytest",
         "pytest-cov",
     )
-    session.install("-e", ".[all]")
+
+    with open(
+        CURRENT_DIRECTORY / "testing" / "constraints-3.6.txt", encoding="utf-8"
```
plamut, line 254: Oh, BTW, we might not want to hardcode the `3.6` version, but instead use `UNIT_TEST_PYTHON_VERSIONS[0]`, i.e. the oldest one supported?

tswast: It's a bit odd to use `UNIT_TEST_PYTHON_VERSIONS` in something that's a system test, but I agree it's worth having. I added a comment explaining why it's that version.

noxfile.py
tswast: I like it! I made one modification, which was to add `^` to only match from the start of a line. Otherwise, it picked up "foo" from the comments.

tswast:
Expected pre-release dependencies are being used:

```
nox > python -c import grpc; print(grpc.__version__)
1.40.0
nox > python -c import pandas; print(pandas.__version__)
1.4.0.dev0+678.ga54aa399dd
nox > python -c import pyarrow; print(pyarrow.__version__)
6.0.0.dev233
```

https://source.cloud.google.com/results/invocations/c54bbd58-2026-421d-a174-153f13f8f86b/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fprerelease-deps-3.8/log

plamut (approved):

plamut:
@tswast Do we first need to wait for the internal change to be applied, or can we already go ahead and merge this?

PR #964, opened 2021-09-13
feat: include key metadata in Job representation

### LINKED ISSUE #958: feat: repr of QueryJob (and other jobs) should include job reference info (project ID, location, job ID)
Without a job ID, there's not much I can do to follow-up with an investigation into test flakes. Project ID, location, and job ID are the first things backend engineers request if there is a problem with a query job. We should ensure that our exceptions include this information (I believe we do much of the time) and just as importantly the `repr` of the job, so that assertion errors like this include the right information.

**Context**

https://github.com/googleapis/python-bigquery/pull/947 had a system test flake.

```
        # Insert a few rows and check the stats.
        sql = f"""
            BEGIN TRANSACTION;
              INSERT INTO `{table_id}`
              VALUES ("one", 1), ("two", 2), ("three", 3), ("four", 4);

              UPDATE `{table_id}`
              SET bar = bar + 1
              WHERE bar > 2;
            COMMIT TRANSACTION;
        """
        query_job = Config.CLIENT.query(sql)
        query_job.result()

        # Transaction ID set by the server should be accessible
>       assert query_job.transaction_info is not None
E       AssertionError: assert None is not None
E        +  where None = <google.cloud.bigquery.job.query.QueryJob object at 0x7fa1b2732610>.transaction_info

tests/system/test_client.py:1583: AssertionError
```

https://source.cloud.google.com/results/invocations/10d5f32f-86da-4278-85d6-25799e2f8155/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fprerelease-deps-3.8/log

Without a job ID, there's little I can do to investigate further.

plamut:
Closes #958.

FWIW, I actually hit the very same test flakiness today, thus adding more info to Job representation will indeed be helpful for investigation.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -722,6 +722,14 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
             and self.error_result.get("reason") == _STOPPED_REASON
         )
 
+    def __repr__(self):
+        result = (
+            f"{self.__class__.__name__}<"
+            f"project={self.project}, location={self.location}, id={self.job_id}"
+            ">"
+        )
+        return result
+
 
 class _JobConfig(object):
     """Abstract base class for job configuration objects.
--- a/tests/unit/job/test_base.py
+++ b/tests/unit/job/test_base.py
@@ -1043,6 +1043,12 @@ class Test_AsyncJob(unittest.TestCase):
 
         self.assertTrue(job.cancelled())
 
+    def test_repr(self):
+        client = _make_client(project="project-foo")
+        job = self._make_one("job-99", client)
+        job._properties.setdefault("jobReference", {})["location"] = "ABC"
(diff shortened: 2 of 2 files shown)


google/cloud/bigquery/job/base.py
```diff
@@ -722,6 +722,14 @@ def cancelled(self):
             and self.error_result.get("reason") == _STOPPED_REASON
         )
 
+    def __repr__(self):
+        result = (
+            f"{self.__class__.__name__}<"
+            f"project={self.project}, location={self.location}, id={self.job_id}"
+            ">"
```
plamut, line 729: Note - I chose chevrons over parentheses so that the repr string does not appear as something `eval()`-able, as it's often the case with `repr()` output.

tswast: Makes sense, thanks.

tswast (approved):

PR #973, opened 2021-09-16
cleanup: get rid of unit test warnings caused by our code

### LINKED ISSUE #957: Unit tests emit quite a few warnings
It appears that recent additions resulted in several warning messages when running unit tests:
```
PytestUnknownMarkWarning: Unknown pytest.mark.enable_add_server_timeout_header - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.enable_add_server_timeout_header
...
FutureWarning: Assigning CRS to a GeoDataFrame without a geometry column is now deprecated and will not be supported in the future.
    return geopandas.GeoDataFrame(crs=_COORDINATE_REFERENCE_SYSTEM)
...
UserWarning: Geometry is in a geographic CRS. Results from 'area' are likely incorrect. Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.
```
At least those that we have control over (e.g. user warnings) should be addressed to reduce the noise.

plamut:
Fixes #957.

This PR should get rid of almost all warnings when running unit tests, except from the one that seems to come from `geopandas` internals.

I initially read a few things on the geographic coordinate systems to understand the problem, and then tried to convert the resulting geodataframes to a different CRS. But that turned out to be a bad idea, because we actually want to actual dataframes in `EPSG:4326`, thus I reverted to just ignoring the user warnings in the test code.

**PR checklist:**

### DIFF
--- a/tests/unit/test_table.py
+++ b/tests/unit/test_table.py
@@ -1866,8 +1866,7 @@ class Test_EmptyRowIterator(unittest.TestCase):
         df = row_iterator.to_geodataframe(create_bqstorage_client=False)
         self.assertIsInstance(df, geopandas.GeoDataFrame)
         self.assertEqual(len(df), 0)  # verify the number of rows
-        self.assertEqual(df.crs.srs, "EPSG:4326")
-        self.assertEqual(df.crs.name, "WGS 84")
+        self.assertIsNone(df.crs)
 
 
 class TestRowIterator(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


tests/unit/test_table.py
```diff
@@ -4027,8 +4026,14 @@ def test_to_geodataframe(self):
...
         self.assertIsInstance(df.geog, geopandas.GeoSeries)
-        self.assertEqual(list(map(str, df.area)), ["0.0", "nan", "0.5"])
-        self.assertEqual(list(map(str, df.geog.area)), ["0.0", "nan", "0.5"])
+
+        with warnings.catch_warnings():
+            # Computing the area on a GeoDataFrame that uses a geographic Coordinate
+            # Reference System (CRS) produces a warning that we are not interested in.
+            warnings.filterwarnings("ignore", category=UserWarning)
+            self.assertEqual(list(map(str, df.area)), ["0.0", "nan", "0.5"])
+            self.assertEqual(list(map(str, df.geog.area)), ["0.0", "nan", "0.5"])
```
tseaver, line 4035: `warnings.filterwarnings` is a global change:  I believe what you want is to trap the warnings for this test case, i.e.:
```suggestion
            with warnings.catch_warnings():
                warnings.filterwarnings("ignore", category=UserWarning)
                self.assertEqual(list(map(str, df.area)), ["0.0", "nan", "0.5"])
                self.assertEqual(list(map(str, df.geog.area)), ["0.0", "nan", "0.5"])
```

tseaver (changes requested):

tests/unit/test_table.py
plamut: That's exactly the code in the test? Do you see it differently perhaps?

tseaver: I saw that you used `warnings.catch_warnings()` in the other locations, but somehow I couldn't see it here.

tseaver (approved):

tswast (approved):

Issue #975, opened 2021-09-18
Test TestBigQuery.test_transaction_info is flaky.


jimfulton:
https://source.cloud.google.com/results/invocations/2ca6ab31-ba0f-4a6b-ab53-55943a5bac42/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-bigquery%2Fpresubmit%2Fsystem-3.8/log

This test fails pretty regularly for me.

tswast:
I checked the job state for a recent failure:

```json
{
  "kind": "bigquery#job",
  "etag": "NSFsBEFD1+My1joEvfuzGQ==",
  "id": "precise-truck-742:US.a9a56a2e-836e-4679-add5-941543bfbf8f",
  "selfLink": "https://content-bigquery.googleapis.com/bigquery/v2/projects/precise-truck-742/jobs/a9a56a2e-836e-4679-add5-941543bfbf8f?location=US",
  "user_email": "REDACTED",
  "configuration": {
    "query": {
      "query": "\n            BEGIN TRANSACTION;\n              INSERT INTO `precise-truck-742.python_bigquery_tests_system_bq_system_test_1632082968448.test_dml_statistics`\n              VALUES (\"one\", 1), (\"two\", 2), (\"three\", 3), (\"four\", 4);\n\n              UPDATE `precise-truck-742.python_bigquery_tests_system_bq_system_test_1632082968448.test_dml_statistics`\n              SET bar = bar + 1\n              WHERE bar \u003e 2;\n            COMMIT TRANSACTION;\n        ",
      "priority": "INTERACTIVE",
      "useLegacySql": false
    },
    "jobType": "QUERY"
  },
  "jobReference": {
    "projectId": "precise-truck-742",
    "jobId": "a9a56a2e-836e-4679-add5-941543bfbf8f",
    "location": "US"
  },
  "statistics": {
    "creationTime": "1632082969228",
    "startTime": "1632082969269",
    "endTime": "1632082974219",
    "totalBytesProcessed": "55",
    "query": {
      "totalBytesProcessed": "55",
      "totalBytesBilled": "10485760",
      "totalSlotMs": "19050",
      "statementType": "SCRIPT"
    },
    "totalSlotMs": "19050",
    "numChildJobs": "4"
  },
  "status": {
    "state": "DONE"
  }
}
```

Looks like no `transactionInfo` key was ever added. I've filed internal issue 200542305 to have the backend team investigate further.

[...]

PR #976, opened 2021-09-19
feat: Add py.typed for PEP 561 compliance

jdufresne:
Type annotations were added in commit
f8d4aaa335a0eef915e73596fc9b43b11d11be9f. For these annotations to be
useful by library users, the package should install a py.typed file.
This tells mypy and other tools to consume and use these types.

For more details, see:
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

### DIFF
--- a/google/cloud/bigquery/py.typed
+++ b/google/cloud/bigquery/py.typed
@@ -1,0 +1,2 @@
+# Marker file for PEP 561.
+# The google-cloud-bigquery package uses inline types.


jdufresne:
Is the CLA required for something so trivial?

plamut:
@jdufresne Thank you for your contribution!

CLA is required, yes, regardless of the contribution size, otherwise the bot will not allow to merge a PR.

jdufresne:
@googlebot I signed it!

plamut:
FWIW, that failed system test is a [known flakiness](https://github.com/googleapis/python-bigquery/issues/975).

@tswast We now actually have the job ID in the logs to investigate, I have shared it with you privately.

plamut (commented):
I would just suggest putting the following content in the `py.typed` file for consistency, similarly to [bigquery_v2/py.typed](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery_v2/py.typed) and `py.typed` files in other Python client libraries:
```
# Marker file for PEP 561.
# The google-cloud-bigquery package uses inline types.
```

jdufresne:
@plamut Thanks for the review. I've applied the suggestion to the latest revision.

plamut (approved):
Looks good as far as I'm concerned. And thanks again!

tswast (approved):

PR #977, opened 2021-09-20
chore: relocate owl bot post processor

SurferJeffAtGoogle:
chore: relocate owl bot post processor


plamut (commented):
The Kokoro check fails with status CANCELLED, but I don't have insight into what the reason could be. Seems like the same error as in the Python Pub/Sub repo.

PR #895, opened 2021-08-23
test: Add test of datetime and time pandas load

jimfulton:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #56  🦕

PR #980, opened 2021-09-21
fix: support ARRAY data type when loading from DataFrame with Parquet

### LINKED ISSUE #19: BigQuery: Upload pandas DataFrame containing arrays

judahrand:

Fixes #19 🦕


judahrand:
I've made this default to `True` because it is technically the correct thing to do from a BigQuery perspective. BQ expects Parquet so we should be giving it compliant Parquet. However, the argument could be made that it should default to `False` for backwards compatibility?

judahrand:
I suppose another question is whether or not we should set `parquet_options.enable_list_inference = True` by default in this method?

judahrand:
These suggestion might be acceptable if released in `3.0.0`?

plamut:
@judahrand First of all, thanks for the research and the fix!

> These suggestions might be acceptable if released in `3.0.0`?

BigQuery version 3.0 is definitely an opportunity to make such breaking changes. But if it turns out that the change fixes the issue without negative consequences, then it can probably be accepted even in one of the upcoming `2.x` releases. Since no tests broke, this is already a good sign.

I'm just not sure about bumping `pyarrow` to  `>= 4.0.0` - we generally try to keep the version pins wide to reduce the chance of conflicting requirements with other cloud libraries or 3rd party dependencies. @tswast do you know if bumping the minimum `pyarrow` version would affect some users?

As for the fix itself, can you also add a corresponding test which proves that the issue has indeed been fixed?

Since it has to do with parquet decoding on the server side, a system test would be needed, but that requires access to an active Google Cloud Project. If you do not have time/resources to write such test, that's OK, we can add one. It's probably just the matter of adjusting one of the code snippets posted in #19.

tswast:
> @tswast do you know if bumping the minimum pyarrow version would affect some users?

I would prefer not to bump pyarrow minimum to 4.0, especially since it was only released this year. I learned recently that there are some Apache Beam / Dataflow systems that are stuck on pyarrow 2.0.

judahrand:
> > @tswast do you know if bumping the minimum pyarrow version would affect some users?
> 
> I would prefer not to bump pyarrow minimum to 4.0, especially since it was only released this year. I learned recently that there are some Apache Beam / Dataflow systems that are stuck on pyarrow 2.0.

I agree but I can't really see another way to address this (and it is quite annoying for some use cases). For example (Feast)[https://github.com/feast-dev/feast] is currently having to implement an ugly hack around this. It's a shame that PyArrow just didn't write proper Parquet for so long! 🤣 

Plus, tbh we're already on `3.0.0` which was Jan 2021. `4.0.0` is April 2021. So only a few months difference.

plamut:
We can conditionally apply the fix using a feature flag whose value depends on the detected dependency version (`pyarrow` in our case). We've done it several times in the past, and it proved successful - it improved user experience while at the same time *not* spoiled it for those who were stuck on older versions (for any reason).

We can take the best from both worlds. :slightly_smiling_face:

judahrand:
> We can conditionally apply the fix using a feature flag whose value depends on the detected dependency version (`pyarrow` in our case). We've done it several times in the past, and it proved successful - it improved user experience while at the same time _not_ spoiled it for those who were stuck on older versions (for any reason).
> 
> We can take the best from both worlds. 🙂
> 
> P.S.: And for the record, we've also done a few innocent-looking version bumps only to later realize that the bump caused problems at other places, i.e. outside the Python BigQuery client - hence all the caution.

Can you point me to an example of where this has been done before so I can replicate the pattern here?

tswast:
> Can you point me to an example of where this has been done before so I can replicate the pattern here?

I recommend adding a property like `use_compliant_nested_type` to to `PyarrowVersions`:

https://github.com/googleapis/python-bigquery/blob/ee1e25c68fff8f35b595a832a582bdd7dd8dc73b/google/cloud/bigquery/_helpers.py#L107

See a similar feature flag for the BQ Storage client:

https://github.com/googleapis/python-bigquery/blob/ee1e25c68fff8f35b595a832a582bdd7dd8dc73b/google/cloud/bigquery/_helpers.py#L76

google/cloud/bigquery/client.py
```diff
@@ -2456,6 +2457,7 @@ def load_table_from_dataframe(
         project: str = None,
         job_config: LoadJobConfig = None,
         parquet_compression: str = "snappy",
+        parquet_use_compliant_nested_type: bool = True,
```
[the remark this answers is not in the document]
judahrand: I'm more than happy to not expose this!! I agree that it's just plain wrong.

I was just thinking about anyone who needs a work around for a table they've already created with a 'wrong' format.

judahrand:
> > Can you point me to an example of where this has been done before so I can replicate the pattern here?
> 
> I recommend adding a property like `use_compliant_nested_type` to to `PyarrowVersions`:
> 
> https://github.com/googleapis/python-bigquery/blob/ee1e25c68fff8f35b595a832a582bdd7dd8dc73b/google/cloud/bigquery/_helpers.py#L107
> 
> See a similar feature flag for the BQ Storage client:
> 
> https://github.com/googleapis/python-bigquery/blob/ee1e25c68fff8f35b595a832a582bdd7dd8dc73b/google/cloud/bigquery/_helpers.py#L76

Cool 😁 Implemented this plus a bit of tidy up.

google/cloud/bigquery/client.py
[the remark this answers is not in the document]
judahrand: I think I should clarity this. The reason it might break old users tables is that the format that `pyarrow` uses for nested lists by default is `list<item: list<item: value>>`. By default (without `use_compliant_nested_type`) it turns this into `list<list<item: value>>` when writing to Parquet. 

When `enable_list_inference` is off in BigQuery this ends up as a `ARRAY<STRUCT> in BigQuery where the `STRUCT` looks like:
```
{
    "list: {
        "item: value
    }
}
```

When `enable_list_inference` is on in BigQuery this ends up as a `ARRAY<STRUCT> in BigQuery where the `STRUCT` looks like:
```
{
    "item": value
}
```

When PyArrow is told to `use_compliant_nested_type` it outputs `list<list<element: value>>`.

When `enable_list_inference` is off in BigQuery this ends up as a `ARRAY<STRUCT> in BigQuery where the `STRUCT` looks like:
```
{
    "list: {
        "element: value
    }
}
```

When `enable_list_inference` is on in BigQuery this ends up as a `ARRAY<value> in BigQuery. Which is what it should be!!!!

My point being that if we switch on `use_compliant_nested_type` without giving the option to turn it off the `item` becomes `element` always. This will be guaranteed to be incompatible with schemas created with the legacy version.

Does that make sense?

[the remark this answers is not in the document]
judahrand: If you're happy to have this breaking change but 'officially' support list types for the pandas connector - then I'm happy to not expose it! And will make the change?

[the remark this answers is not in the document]
tswast: Yes, this change seems to be what's needed to officially support list types. 😁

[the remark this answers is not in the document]
tswast: Makes sense. I hadn't really expected anyone to actually use that rather strange schema...

Since https://github.com/googleapis/python-bigquery/issues/19 is still open, I don't consider lists supported as of today, so I'm okay making this breaking change.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -444,7 +444,14 @@ def augment_schema(dataframe, current_bq_schema):
             continue
 
         arrow_table = pyarrow.array(dataframe[field.name])
-        detected_type = ARROW_SCALAR_IDS_TO_BQ.get(arrow_table.type.id)
+
+        if arrow_table.type.id == 25:
```
tswast, line 448: Is there a constant we can use to derive this number? I'm a bit wary of having the magic number 25 in tests.

tswast: Or better yet, can we use [pyarrow.types.is_list(arrow_table.type)](https://arrow.apache.org/docs/python/generated/pyarrow.types.is_list.html#pyarrow.types.is_list)?

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -551,14 +564,29 @@ def dataframe_to_parquet(dataframe, bq_schema, filepath, parquet_compression="SN
...
+
+            This argument is ignored for ``pyarrow`` versions earlier than ``4.0.0``.
     """
     pyarrow = _helpers.PYARROW_VERSIONS.try_import(raise_if_error=True)
 
     import pyarrow.parquet
 
+    kwargs = (
+        {"use_compliant_nested_type": parquet_use_compliant_nested_type}
+        if _helpers.PYARROW_VERSIONS.use_compliant_nested_type
```
tswast, line 581: 👍

google/cloud/bigquery/client.py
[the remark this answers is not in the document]
tswast: That said, if you think it would be disruptive we can consider adding a field now, but then we immediately remove it in `google-cloud-bigquery` 3.0 🤔

google/cloud/bigquery/client.py
```diff
@@ -2647,9 +2668,19 @@ def load_table_from_dataframe(
...
+                        parquet_use_compliant_nested_type=parquet_use_compliant_nested_type,
                     )
                 else:
-                    dataframe.to_parquet(tmppath, compression=parquet_compression)
+                    dataframe.to_parquet(
+                        tmppath,
+                        engine="pyarrow",
+                        compression=parquet_compression,
+                        **{
+                            "use_compliant_nested_type": parquet_use_compliant_nested_type
```
tswast, line 2679: Do we need to avoid passing this in if the pandas version doesn't support it? Or does pandas just pass through to pyarrow in older versions of pandas?

google/cloud/bigquery/client.py
```diff
@@ -2562,6 +2576,13 @@ def load_table_from_dataframe(
         if job_config.source_format is None:
             # default value
             job_config.source_format = job.SourceFormat.PARQUET
+
+        if job_config.parquet_options is None:
```
tswast, line 2580: Since CSV is also supported, should we check `source_format` here?

google/cloud/bigquery/client.py
```diff
@@ -2562,6 +2576,13 @@ def load_table_from_dataframe(
...
+            parquet_options = ParquetOptions()
+            # default value
+            parquet_options.enable_list_inference = True
```
tswast, line 2583: Does this change the behavior with non-compliant parquet files? Should we add a `pyarrow` version check here too?

That could be nice in that folks can drop to pyarrow 3.0.0 if they really want the old behavior.

tests/system/test_pandas.py
```diff
@@ -112,6 +188,21 @@ def test_load_table_from_dataframe_w_automatic_schema(bigquery_client, dataset_i
         bigquery.SchemaField("uint8_col", "INTEGER"),
         bigquery.SchemaField("uint16_col", "INTEGER"),
         bigquery.SchemaField("uint32_col", "INTEGER"),
+        bigquery.SchemaField("array_bool_col", "BOOLEAN", mode="REPEATED"),
+        bigquery.SchemaField("array_ts_col", "TIMESTAMP", mode="REPEATED"),
+        # BigQuery does not support uploading DATETIME values from
+        # Parquet files. See:
+        # https://github.com/googleapis/google-cloud-python/issues/9996
```
tswast, line 195: FYI: This actually has been fixed in the backend, you just have to explicitly ask for DATETIME in the schema when loading the dataframe to avoid breaking changes. https://github.com/googleapis/python-bigquery/pull/895

I've filed https://github.com/googleapis/python-bigquery/issues/985 to address this in google-cloud-bigquery v3.

Let's add a TODO here pointing at https://github.com/googleapis/python-bigquery/issues/985 instead.

tswast (commented):
Awesome work! A few comments, but looking good, especially with those system tests.

google/cloud/bigquery/client.py
judahrand: I've opened a PR with them about this same thing but yes, currently it goes straight through to `pyarrow` so need to catch it.

google/cloud/bigquery/client.py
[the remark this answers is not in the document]
judahrand: I think it's reasonable to make the breaking change. If someone cares enough to maintain the old behaviour they can write the Dataframe using whatever options they want manually and then use whatever JobConfig they like with `load_table_from_file`. Doesn't seem worth it to add the option and immediately remove it. Especially, as as you say... the old schema is.... difficult to use.

google/cloud/bigquery/_pandas_helpers.py
judahrand: Will implement this tomorrow.

google/cloud/bigquery/client.py
judahrand: Yes it does. Goes from `[{list: {item: value}}]` to `[{item: value}]` I think.

judahrand: They could always drop to pyarrow 3.0.0 and set `enable_list_inference` to `False` 😛

google/cloud/bigquery/client.py
judahrand: Yup, you're right. Probably would be fine. But also best not to be going around editing the JobConfig if not necessary.

judahrand:
Do you think that we can strip this bit out of the docs now too? 
https://github.com/googleapis/python-bigquery/blob/5f915cd39a42051b3041cda70b363cffbb604414/google/cloud/bigquery/client.py#L2464

judahrand:
Do you think we can also strip the bit about Parquet being weird out of the docs? Or worth leaving in for people with old pyarrow?

https://github.com/googleapis/python-bigquery/blob/5f915cd39a42051b3041cda70b363cffbb604414/google/cloud/bigquery/client.py#L2468

tswast:
> Or worth leaving in for people with old pyarrow?

Let's just say REPEATED/ARRAY types are only supported when using PARQUET format with pyarrow 4.0+ installed.

google/cloud/bigquery/client.py
tswast: Based on the coverage failure, I think we need a unit test where `parquet_options` is manually set. Perhaps we can have a test where `parquet_options.enable_list_inference = False` and we verify that we don't override that?

tests/system/test_pandas.py
judahrand: Done.

google/cloud/bigquery/client.py
judahrand: Done.

google/cloud/bigquery/_pandas_helpers.py
judahrand: Done.

judahrand:
> > Or worth leaving in for people with old pyarrow?
> 
> Let's just say REPEATED/ARRAY types are only supported when using PARQUET format with pyarrow 4.0+ installed.

😁 Think I've taken care of all your reviews. Thanks for all your help!

google/cloud/bigquery/client.py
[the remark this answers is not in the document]
tswast: Cool. Let's remove it, then. Especially since folks have a workaround if they really need the unsupported schema.

judahrand:
@tswast Are we all good here? 😄

google/cloud/bigquery/client.py
```diff
@@ -2647,9 +2648,17 @@ def load_table_from_dataframe(
...
                     )
                 else:
-                    dataframe.to_parquet(tmppath, compression=parquet_compression)
+                    dataframe.to_parquet(
+                        tmppath,
+                        engine="pyarrow",
+                        compression=parquet_compression,
+                        **{"use_compliant_nested_type": True}
+                        if _helpers.PYARROW_VERSIONS.use_compliant_nested_type
+                        else {},
```
tswast, line 2660: Let's wrap this to be clear that we are doing the `**` on the `else` branch, too.

```suggestion
                        **(
                            {"use_compliant_nested_type": True}
                            if _helpers.PYARROW_VERSIONS.use_compliant_nested_type
                            else {}
                        ),
```

google/cloud/bigquery/client.py
```diff
@@ -2562,6 +2556,13 @@ def load_table_from_dataframe(
         if job_config.source_format is None:
             # default value
             job_config.source_format = job.SourceFormat.PARQUET
+
+        if job_config.parquet_options is None:
```
tswast, line 2560:
```suggestion
        if (
            job_config.source_format == job.SourceFormat.PARQUET
            and job_config.parquet_options is None
        ):
```

tswast (approved):

PR #983, opened 2021-09-22
fix: disambiguate missing policy tags from explicitly unset policy tags

### LINKED ISSUE #981: disambiguate missing policy tags from explicitly unset policy tags
As seen in internal bug 182204971, https://github.com/googleapis/python-bigquery/pull/557, if the `policyTags` key is included in a load job, the user can get ` 403 POST ... User does not have permission to get taxonomy ...` errors when uploading data, even if they have write access to the table.

Based on the behavior seen in https://github.com/googleapis/python-bigquery-pandas/issues/387, I believe this fix has accidentally been reverted in https://github.com/googleapis/python-bigquery/pull/703#issuecomment-925107546.

TODO:

- [x] Manually test to see if the error from 182204971 is reproduced
- [x] ~Add a system test to avoid future regressions~
- [x] Update `SchemaField` to allow disambiguating unset  vs explicitly set to empty list / none.

### LINKED ISSUE #982: missing docs for `google.cloud.bigquery.schema.PolicyTagList`
Need link here: https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.schema.SchemaField.html#google.cloud.bigquery.schema.SchemaField.policy_tags

Also, need link here:

https://googleapis.dev/python/bigquery/latest/reference.html#schema

tswast:

Fixes #981 
Fixes #982
Towards https://github.com/googleapis/python-bigquery-pandas/issues/387
🦕

### DIFF
--- a/google/cloud/bigquery/schema.py
+++ b/google/cloud/bigquery/schema.py
@@ -15,6 +15,7 @@
 """Schemas for BigQuery tables / queries."""
 
 import collections
+import enum
 from typing import Iterable, Union
 
 from google.cloud.bigquery_v2 import types
@@ -48,11 +49,20 @@ LEGACY_TO_STANDARD_TYPES = {
 """String names of the legacy SQL types to integer codes of Standard SQL types."""
 
 
-class _DefaultSentinel:
-    """Object used as 'sentinel' indicating default value should be used."""
+class _DefaultSentinel(enum.Enum):
+    """Object used as 'sentinel' indicating default value should be used.
+
+    Uses enum so that pytype/mypy knows that this is the only possible value.
+    https://stackoverflow.com/a/60605919/101923
+
+    Literal[_DEFAULT_VALUE] is an alternative, but only added in Python 3.8.
+    https://docs.python.org/3/library/typing.html#typing.Literal
+    """
+
+    DEFAULT_VALUE = object()
 
 
-_DEFAULT_VALUE = _DefaultSentinel()
+_DEFAULT_VALUE = _DefaultSentinel.DEFAULT_VALUE
 
 
 class SchemaField(object):
--- a/google/cloud/bigquery/schema.py
+++ b/google/cloud/bigquery/schema.py
@@ -15,12 +15,12 @@
 """Schemas for BigQuery tables / queries."""
 
 import collections
-from typing import Optional, Tuple
+from collections.abc import Iterable
+from typing import Union
 
 from google.cloud.bigquery_v2 import types
 
 
-_DEFAULT_VALUE = object()
 _STRUCT_TYPES = ("RECORD", "STRUCT")
 
 # SQL types reference:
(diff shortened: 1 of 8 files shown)


google/cloud/bigquery/schema.py
```diff
@@ -152,10 +136,10 @@ def from_api_repr(cls, api_repr: dict) -> "SchemaField":
         mode = api_repr.get("mode", "NULLABLE")
         description = api_repr.get("description", _DEFAULT_VALUE)
         fields = api_repr.get("fields", ())
+        policy_tags = api_repr.get("policyTags", _DEFAULT_VALUE)
 
-        policy_tags = cls._determine_policy_tags(
-            field_type, PolicyTagList.from_api_repr(api_repr.get("policyTags"))
-        )
+        if policy_tags is not None and policy_tags is not _DEFAULT_VALUE:
+            policy_tags = PolicyTagList.from_api_repr(api_repr.get("policyTags"))
```
plamut, line 142: It seems we can just pass `policy_tags` as an argument directly, because we already checked that `"policyTags"`  key exists in `api_repr` (it's not `_DEFAULT_VALUE`) and is not `None`.

plamut (changes requested):
Looks good (with one minor nit), seems like we just need to adjust the type annotation and it should be good.

google/cloud/bigquery/schema.py
tswast: Good call, done.

Also, I confused myself as to why I had `_DEFAULT_VALUE` in the first place, so I added some assertions in tests/unit/test_schema.py to make sure we aren't including the `policyTags` key when it's not present in the original API representation.

google/cloud/bigquery/schema.py
```diff
@@ -81,15 +81,15 @@ class SchemaField(object):
...
-        mode="NULLABLE",
-        description=_DEFAULT_VALUE,
-        fields=(),
-        policy_tags=None,
-        precision=_DEFAULT_VALUE,
-        scale=_DEFAULT_VALUE,
-        max_length=_DEFAULT_VALUE,
+        name: str,
+        field_type: str,
+        mode: Optional[str] = "NULLABLE",
```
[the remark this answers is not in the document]
tswast: That seems to have helped, but it looks like I might have to switch up a bit how we define the sentinel value to make the type checker happy. https://stackoverflow.com/a/60605919/101923

google/cloud/bigquery/schema.py
```diff
@@ -49,47 +49,62 @@
...
             Precison (number of digits) of fields with NUMERIC or BIGNUMERIC type.
 
-        scale (Optional[int]):
+        scale:
             Scale (digits after decimal) of fields with NUMERIC or BIGNUMERIC type.
 
-        max_length (Optional[int]):
-            Maximim length of fields with STRING or BYTES type.
-
+        max_length: Maximim length of fields with STRING or BYTES type.
```
plamut, line 94: Typo:
```suggestion
        max_length: Maximum length of fields with STRING or BYTES type.
```

plamut (commented):
Looks good, except for a typo.

Edit: Just fixed it myself.

plamut (approved):

google/cloud/bigquery/schema.py
tswast: Thanks!

PR #988, opened 2021-09-27
fix: remove py.typed since package fails mypy check

tswast:
Reverts googleapis/python-bigquery#976

Context: https://issues.apache.org/jira/browse/BEAM-12975

### DIFF
--- a/google/cloud/bigquery/py.typed
+++ b/google/cloud/bigquery/py.typed
@@ -1,2 +1,0 @@
-# Marker file for PEP 561.
-# The google-cloud-bigquery package uses inline types.


tswast:
Will check the library with `mypy` before reverting.

dandhlee (approved):

plamut:
Do we know what exactly caused that error? Is it an issue with some of our annotations, or does `mypy` behave differently from `pytype`?

It would be good to fix this while still retaining `py.typed`, although removing it as a short term fix to unblock maintainers of other is fine.

PR #994, opened 2021-09-29
feat: add `AvroOptions` to configure AVRO external data

tswast:
Also:

* Unify `ExternalConfig` class to use `_properties` for everything. This does
  result in more code, but it should make maintenance easier as it aligns with
  our other mutable resource classes.
* Adds `bigtable_options`, `csv_options`, and `google_sheets_options`
  properties. This aligns with `parquet_options`.

In response to internal request. 🦕

### DIFF
--- a/google/cloud/bigquery/external_config.py
+++ b/google/cloud/bigquery/external_config.py
@@ -952,10 +952,6 @@ class ExternalConfig(object):
                 A dictionary in the format used by the BigQuery API.
         """
         config = copy.deepcopy(self._properties)
-        if self.options is not None:
-            r = self.options.to_api_repr()
-            if r != {}:
-                config[self.options._RESOURCE_NAME] = r
         return config
 
     @classmethod
--- a/google/cloud/bigquery/external_config.py
+++ b/google/cloud/bigquery/external_config.py
@@ -22,13 +22,13 @@ from __future__ import absolute_import
 
 import base64
 import copy
-from typing import FrozenSet, Iterable, Optional
+from typing import FrozenSet, Iterable, Optional, Union
 
 from google.cloud.bigquery._helpers import _to_bytes
 from google.cloud.bigquery._helpers import _bytes_to_json
 from google.cloud.bigquery._helpers import _int_or_none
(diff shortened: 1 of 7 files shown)


plamut (commented):
Nothing much to add, except for the type hint duplication nit.

A few lines are missing from the coverage, please just address that.

google/cloud/bigquery/external_config.py
```diff
@@ -815,23 +824,125 @@ def schema(self, value):
         self._properties["schema"] = prop
 
     @property
-    def parquet_options(self):
-        """Optional[google.cloud.bigquery.format_options.ParquetOptions]: Additional
-        properties to set if ``sourceFormat`` is set to PARQUET.
+    def avro_options(self) -> Optional[AvroOptions]:
+        """Optional[google.cloud.bigquery.format_options.AvroOptions]:
```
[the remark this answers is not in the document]
tswast: True. I was hoping that I could get Sphinx to link these up since it didn't with just the type annotation. Didn't work though, so I can remove.

google/cloud/bigquery/external_config.py
```diff
@@ -897,8 +894,7 @@ def csv_options(self, value):
 
     @property
     def google_sheets_options(self) -> Optional[GoogleSheetsOptions]:
-        """Optional[google.cloud.bigquery.external_config.GoogleSheetsOptions]:
-        Additional properties to set if ``sourceFormat`` is set to
+        """Additional properties to set if ``sourceFormat`` is set to
         GOOGLE_SHEETS.
```
plamut: (nit) Does this fit into a single line?

plamut (approved):
LGTM. You can address a nit if you want.

google/cloud/bigquery/external_config.py
tswast: Not within 79 characters, which is what I've got VS Code configured to wrap at.

Looks like Sphinx did the right thing. Probably because there's no blank line.

Screen Shot 2021-09-30 at 11 52 47 AM

plamut: Now that surprised me, but even better then!

PR #996, opened 2021-09-30
chore: sync v3 branch with main

plamut:
A yet another episode of the well-know series. :)


google/cloud/bigquery/table.py
```diff
@@ -117,46 +118,136 @@ def _view_use_legacy_sql_getter(table):
...
-    def project(self):
-        """str: Project bound to the table"""
-        return self._project
+    def __init__(
+        self,
+        dataset_ref: "DatasetReference",
+        table_id: str,
+        project_id_alternative: Sequence[str] = None,
+        dataset_id_alternative: Sequence[str] = None,
+        table_id_alternative: Sequence[str] = None,
```
plamut, line 213: These three "alternative" attributes are new, observed them in the generated code. Had to port them manually to the v3 refactored models.

tswast: Weird. I'm investigating what their purpose are. From what I can tell they were added for weird API-translation layer reasons. Doubtful that they should ever have been exposed to the public.

tswast: Found this comment "[When a BigQuery client] passes an array for some fields when it should be passing a single value, this currently works due to a bug".

Let's remove these fields. They are unnecessary and confusing.

plamut: Done https://github.com/googleapis/python-bigquery/pull/996/commits/fc6f7fcf9873311c921ded4edda68d2c3d08aaed

google/cloud/bigquery/client.py
```diff
@@ -116,6 +116,8 @@
 
 TIMEOUT_HEADER = "X-Server-Timeout"
 
+TIMEOUT_HEADER = "X-Server-Timeout"
+
```
tswast, line 120: Looks like a merge conflict doubled up this constant.

```suggestion
```

google/cloud/bigquery/table.py
```diff
@@ -1749,6 +1690,8 @@ def to_arrow(
         """
         self._maybe_warn_max_results(bqstorage_client)
 
+        self._maybe_warn_max_results(bqstorage_client)
+
```
tswast, line 1694: Another merge problem.

```suggestion
```

google/cloud/bigquery/client.py
plamut: What's scary is that it even wasn't a conflict, some of the code just silently creeps in. The most notable example are the `TableReference` methods that have been removed in  the `main` branch (moved to `_TableBase`), but the merged file contained _both_ without any conflicts.

Thanks for catching duplicate declarations! The caught a few that I missed, but apparently even that is no 100% proof, as all CI checks passed.

tswast (approved):

PR #1000, opened 2021-10-01
feat: add a static copy of legacy proto-based types

### LINKED ISSUE #999: Keep legacy generated SQL types available for import
As a follow-up to #814 and as discussed with @tswast on the chat, some users, that otherwise could upgrade to a `3.x` version, would still need the legacy generated SQL model types that the `3.x` version will remove.

We should add these models back to the codebase, but with the following restrictions:
- These legacy files would be copy-pasted to their old place, but will not be generated anymore.
- The legacy models would not be exposed in the library's top-level namespace, nor would they be imported or used anywhere else in the code. Users will have to import them using the full module path, and at their own risk.
- These types will not be maintained in any way, and will eventually get out of sync with what the code generator would produce. This is intentional.
- When importing anything from that legacy sub-package, a warning should be issued.

@tswast Feel free to update/correct the requirements above in case I misunderstood/misremembered them.

plamut:
Closes #999.

Adding unmaintained legacy types definitions.

### DIFF
--- a/google/cloud/bigquery_v2/__init__.py
+++ b/google/cloud/bigquery_v2/__init__.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+from .types.encryption_config import EncryptionConfiguration
+from .types.model import DeleteModelRequest
+from .types.model import GetModelRequest
+from .types.model import ListModelsRequest
+from .types.model import ListModelsResponse
+from .types.model import Model
+from .types.model import PatchModelRequest
+from .types.model_reference import ModelReference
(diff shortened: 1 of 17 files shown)


google/cloud/bigquery_v2/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+from .types.standard_sql import StandardSqlTableType
+from .types.table_reference import TableReference
+
+
+_LEGACY_MSG = (
+    "Legacy proto-based types from bigquery_v2 are not maintained anymore, "
+    "use types defined in google.cloud.bigquery instead."
+)
+
+warnings.warn(_LEGACY_MSG, category=DeprecationWarning)
```
tseaver, line 39: Nice -- I had been thinking to add a request for import-time warnings.

tseaver (approved):

google/cloud/bigquery_v2/__init__.py
plamut: Thanks!

plamut:
Oh, wait... didn't we say we want to add back the proto* dependencies, so that those imports will not result in an error? Maybe as a "legacy" extra?

And what about the intersphinx links? Should we perhaps expose these models in the docs with a big fat "OBSOLETE!" warning at the top?

@tswast

tswast:
> Oh, wait... didn't we say we want to add back the proto* dependencies, so that those imports will not result in an error? Maybe as a "legacy_types" extra?

Yes, please. I think we've added the BQ Storage API as a required dependency now, right? That should mean we pull those in transitively anyway, but it wouldn't hurt to be explicit.

> And what about the intersphinx links? Should we perhaps expose these models in the docs with a big fat "OBSOLETE!" warning at the top?

Yeah, let's make sure we generate docs for it, but with a warning & link to our migration guide.

tswast (approved):

PR #1004, opened 2021-10-06
chore: sync v3 with main

tswast:
Thought: maybe if we merge this as a merge commit, we can avoid some conflicts the next time we sync?


google/cloud/bigquery/table.py
```diff
@@ -1690,6 +1688,8 @@ def to_arrow(
         """
         self._maybe_warn_max_results(bqstorage_client)
 
+        self._maybe_warn_max_results(bqstorage_client)
+
```
tswast, line 1692: Duplicate

```suggestion
```

tseaver:
@tswast I think the change to the samples image is [forthcoming soon](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/6813).

owlbot.py
```diff
@@ -32,8 +32,6 @@
     intersphinx_dependencies={
         "pandas": "http://pandas.pydata.org/pandas-docs/stable/",
         "geopandas": "https://geopandas.org/",
-        "proto-plus": ("https://proto-plus-python.readthedocs.io/en/latest/", None),
-        "protobuf": ("https://googleapis.dev/python/protobuf/latest/", None),
```
plamut, line 36: IIRC these were added these ack when we re-added legacy proto-based types to `v3`. We should thus retain them.

owlbot.py
```diff
@@ -55,10 +54,6 @@
     ],
 )
 
-# Remove unneeded intersphinx links, the library does not use any proto-generated code.
-s.replace("docs/conf.py", r'\s+"(proto-plus|protobuf)":.*$', "")
```
plamut, line 59: Related to the comment above, we actually want to include proto* intersphinx links.

plamut (changes requested):
It seems fine, apart from the intersphinx links that should be retained.

The unit tests still produce no warnings, and the `skipif` decorators in tests do not check for dependencies that have been made required in `v3` (BQ storage, pyarrow...) - these are the two main things that can sneak in without a merge conflict, as far as I can remember.

plamut:
Added the `do not merge` label as an explicit reminder to use a _merge commit_ when merging this (do NOT squash). Hopefully the person who eventually merges this will notice this comment when seeking the reason why the label was added in the first place. :upside_down_face:

plamut (approved):

PR #1018, opened 2021-10-12
chore: disable required linear history on v3 branch

tswast:
This should prevent conflicts when we sync v3 with main.

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -4,7 +4,7 @@ branchProtectionRules:
 # Identifies the protection rule pattern. Name of the branch to be protected.
 # Defaults to `main`
 - pattern: main
-  mergeCommitAllowed: false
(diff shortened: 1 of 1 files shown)


tseaver (approved):

tswast:
Hmmm... Looks like sync-repo-settings config doesn't allow a per-branch setting for `mergeCommitAllowed`, but maybe I do still need to set it.

https://github.com/googleapis/repo-automation-bots/blob/9ca793a0420286600bb075e23430d594131ba34b/packages/sync-repo-settings/src/sync-repo-settings.ts#L277

`requiresLinearHistory` seems to be the right name. Not sure why it didn't match the spelling of the GitHub API.

https://github.com/googleapis/repo-automation-bots/blob/9ca793a0420286600bb075e23430d594131ba34b/packages/sync-repo-settings/src/sync-repo-settings.ts#L183

I think I need both.

PR #1024, opened 2021-10-15
feat: add session and connection properties to QueryJobConfig

### LINKED ISSUE #1026: Add session support
This feature request is to add support for sessions to BigQuery clients. The details have been shared internally with the library maintainers.

See also:
- Reference [implementation](https://github.com/googleapis/java-bigquery/pull/1652/files) for the Java client.
- Reference [implementation](https://github.com/googleapis/google-cloud-go/pull/4754/files) for the Go client.

tswast:

Fixes internal issue 202792922 🦕
Closes https://github.com/googleapis/python-bigquery/issues/1026
Closes https://github.com/googleapis/python-bigquery/pull/1029

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -204,7 +204,10 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
 
     @property
     def session_info(self) -> Optional["SessionInfo"]:
-        """[Preview] Information of the session if this job is part of one."""
+        """[Preview] Information of the session if this job is part of one.
+
+        .. versionadded:: 2.29.0
+        """
         resource = _helpers._get_sub_prop(
             self._properties, ["statistics", "sessionInfo"]
         )
@@ -1002,6 +1005,8 @@ class ScriptStatistics(object):
 
 class SessionInfo(object):
     """[Preview] Information of the session if this job is part of one.
+
+    .. versionadded:: 2.29.0
 
     Args:
         resource (Map[str, Any]): JSON representation of object.
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -203,6 +203,16 @@ class _AsyncJob(google.api_core.future.polling.PollingFuture):
         return ScriptStatistics(resource)
 
     @property
+    def session_info(self) -> Optional["SessionInfo"]:
+        """[Preview] Information of the session if this job is part of one."""
+        resource = _helpers._get_sub_prop(
+            self._properties, ["statistics", "sessionInfo"]
+        )
+        if resource is None:
+            return None
+        return SessionInfo(resource)
+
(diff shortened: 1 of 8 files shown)


google/cloud/bigquery/job/query.py
```diff
@@ -31,11 +31,14 @@
...
-from google.cloud.bigquery.query import StructQueryParameter
-from google.cloud.bigquery.query import UDFResource
+from google.cloud.bigquery.query import (
+    _query_param_from_api_repr,
+    ArrayQueryParameter,
+    ConnectionProperty,
+    ScalarQueryParameter,
+    StructQueryParameter,
+    UDFResource,
+)
```
plamut, line 41: Curious - is this the new style guide for import from the same module/package? I thought we should have a separate `from foo import ...` line for each import?

google/cloud/bigquery/job/query.py
```diff
@@ -269,6 +272,18 @@ def allow_large_results(self):
     def allow_large_results(self, value):
         self._set_sub_prop("allowLargeResults", value)
 
+    @property
+    def connection_properties(self) -> Sequence[ConnectionProperty]:
```
plamut, line 276: (nit) We can be more specific about the return type, it's always `List[...]` (which of course is a specific type of Sequence)

If changing this, the annotation of the other `connection_properties` property should be adjusted as well.

google/cloud/bigquery/job/base.py
```diff
@@ -202,6 +202,16 @@ def script_statistics(self) -> Optional["ScriptStatistics"]:
             return None
         return ScriptStatistics(resource)
 
+    @property
+    def session_info(self) -> Optional["SessionInfo"]:
+        """[Preview] Information of the session if this job is part of one."""
```
plamut, line 207: Do we add the versionadded notice to the docstring for preview properties?
It would probably also be nice to include a URL to the product reference like we already do in several other classes.

(applies to the other new properties as well)

google/cloud/bigquery/job/query.py
```diff
@@ -269,6 +272,18 @@ def allow_large_results(self):
...
+        """Connection properties."""
+        resource = self._get_sub_prop("connectionProperties", [])
+        return [ConnectionProperty.from_api_repr(prop) for prop in resource]
+
+    @connection_properties.setter
+    def connection_properties(self, value: Sequence[ConnectionProperty]):
```
plamut, line 282: `value` can actually be any iterable here, not necessarily a Sequence.

plamut (commented):
Don't know the exact acceptance criteria for this addition, but at a face value it more or less looks good to me. Had just a few nit comments about type annotations and docstrings.

plamut (commented):
Looks good, just a question about exposing `SessionInfo` in the generated docs.

plamut (approved):

PR #1043, opened 2021-11-03
feat: support Python 3.10

### LINKED ISSUE #1006: Python 3.10 support
Hello!

It would be great to get python 3.10 supported by this excellent package.

Thanks,
Rustam

tswast:

Fixes #1006 🦕
Towards https://github.com/googleapis/python-bigquery-datatransfer/issues/244

### DIFF
--- a/samples/geography/requirements.txt
+++ b/samples/geography/requirements.txt
@@ -24,14 +24,12 @@ importlib-metadata==4.8.1
 libcst==0.3.21
 munch==2.5.0
 mypy-extensions==0.4.3
-numpy==1.19.5; python_version < "3.7"
-numpy==1.21.2; python_version > "3.6"
 packaging==21.0
 pandas==1.1.5; python_version < '3.7'
-pandas==1.3.2; python_version >= '3.7'
+pandas==1.3.4; python_version >= '3.7'
 proto-plus==1.19.2
 protobuf==3.18.0
-pyarrow==5.0.0
+pyarrow==6.0.0
 pyasn1==0.4.8
 pyasn1-modules==0.2.8
 pycparser==2.20
@@ -43,7 +41,7 @@ pytz==2021.1
 PyYAML==5.4.1
 requests==2.26.0
 rsa==4.7.2
-Shapely==1.7.1
+Shapely==1.8.0
 six==1.16.0
 typing-extensions==3.10.0.2
 typing-inspect==0.7.1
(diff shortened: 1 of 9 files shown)


samples/geography/requirements.txt
```diff
@@ -24,14 +24,12 @@ importlib-metadata==4.8.1
 libcst==0.3.21
 munch==2.5.0
 mypy-extensions==0.4.3
-numpy==1.19.5; python_version < "3.7"
-numpy==1.21.2; python_version > "3.6"
```
plamut, line 28: Was `numpy` a redundant dependency?

samples/snippets/requirements.txt
```diff
@@ -3,10 +3,10 @@ google-cloud-bigquery-storage==2.9.0
...
 grpcio==1.41.0
 ipython==7.16.1; python_version < '3.7'
-ipython==7.17.0; python_version >= '3.7'
+ipython==7.29.0; python_version >= '3.7'
 matplotlib==3.3.4; python_version < '3.7'
 matplotlib==3.4.1; python_version >= '3.7'
 pandas==1.1.5; python_version < '3.7'
-pandas==1.3.2; python_version >= '3.7'
-pyarrow==5.0.0
+pandas==1.3.4; python_version >= '3.7'
```
plamut, line 10: Hm... I noticed that in `constraints-3.6` we pin `pandas` to `0.24.2`, and `setup.py` pins it to as low as `pandas>=0.23.0`.

I forgot if we intentionally only test samples with up-to-date dependencies?

plamut (commented):
Where's the Samples check with Python 3.10? And shouldn't we also run the system tests with the latest supported Python version?

Currently we only run them with Python 3.8.

samples/snippets/requirements.txt
tswast: For samples we try to use the latest working version of everything, but that's a good point re: setup.py. We likely forgot to update that when updating the constraints to fix some test.

samples/geography/requirements.txt
tswast: Yes. We don't use it directly, it comes in as a dependency of pandas. Rather than try and find the right combo of versions to support Python 3.10, I just dropped it.

tseaver (approved):

tswast:
Need to skip installing geopandas on 3.10

tswast:
I didn't see an open issue for geopandas Python 3.10 support, so I filed https://github.com/geopandas/geopandas/issues/2212

tswast:
Mailed internal change 407388245 to enable system tests on Python 3.10

PR #1046, opened 2021-11-04
fix: error if eval()-ing repr(SchemaField)

### LINKED ISSUE #1041: SchemaField cannot initialize from __repr__
#### Environment details

  - OS type and version: macOS Big Sur 11.6.1
  - Python version: `3.9.7`
  - pip version: `21.3.1`
  - `google-cloud-bigquery` version: `2.29.0`

#### Description

Prior versions of `google-cloud-bigquery` (`2.3.1`) allowed `SchemaField` instantiation based on an existing `SchemaField` `__repr__`.

#### Use Case

1. System generates list of `SchemaField` and passes it through a Jinja template as part of building hundreds of Airflow DAGs.
2. Jinja prints `__repr__` from the `SchemaField` into the templated `.py`. 
3. Airflow picks up the `.py` files and executes them. 
4. Execution results in the below error.

#### Code example

In version `2.3.1`

```python
from google.cloud.bigquery import SchemaField
bigquery.__version__
'2.3.1'

# This works
field = SchemaField('name', 'STRING', 'REQUIRED', 'Name') 
field.__repr__()
"SchemaField('name', 'STRING', 'REQUIRED', 'Name', (), ())"
# Jinja prints ^^ to a .py file
# When the file is executed the SchemaField is created without issue
```

Version `2.29.0`:

```python
from google.cloud.bigquery import SchemaField
bigquery.__version__
'2.29.0'

# This raises an error for the tuple in the last field
field = SchemaField('name', 'STRING', 'REQUIRED', 'Name') 
field.__repr__()
"SchemaField('name', 'STRING', 'REQUIRED', 'Name', (), ())"
# Jinja prints ^^ to a .py file
# When the file is executed, an error is raised:

  File "/Users/wes.roach/.pyenv/versions/upgr-dbt/lib/python3.9/site-packages/google/cloud/bigquery/schema.py", line 125, in __init__
    policy_tags.to_api_repr() if policy_tags is not None else None
AttributeError: 'tuple' object has no attribute 'to_api_repr'
```

Setting the `policy_tags` to `None` still results in `()`.

```python
field = SchemaField('name', 'STRING', 'REQUIRED', 'Name', (), None)
field.__repr__()
"SchemaField('name', 'STRING', 'REQUIRED', 'Name', (), ())"
```

plamut:
Fixes #1041.

If policy tags are not set, the corresponding key should be `None` to denote a missing value. An empty tuple should only be used when the schema has explicitly empty `PolicyTagList` assigned.

**PR checklist:**

### DIFF
--- a/tests/unit/test_schema.py
+++ b/tests/unit/test_schema.py
@@ -513,8 +513,20 @@ class TestSchemaField(unittest.TestCase):
         expected = "SchemaField('field1', 'STRING', 'NULLABLE', None, (), None)"
         self.assertEqual(repr(field1), expected)
 
-    def test___repr__evaluable(self):
+    def test___repr__evaluable_no_policy_tags(self):
         field = self._make_one("field1", "STRING", "REQUIRED", "Description")
+        field_repr = repr(field)
+        SchemaField = self._get_target_class()  # needed for eval  # noqa
+
+        evaled_field = eval(field_repr)
+
+        assert field == evaled_field
+
+    def test___repr__evaluable_with_policy_tags(self):
+        policy_tags = PolicyTagList(names=["foo", "bar"])
+        field = self._make_one(
+            "field1", "STRING", "REQUIRED", "Description", policy_tags=policy_tags,
+        )
         field_repr = repr(field)
         SchemaField = self._get_target_class()  # needed for eval  # noqa
 
--- a/tests/unit/test_schema.py
+++ b/tests/unit/test_schema.py
@@ -510,8 +510,17 @@ class TestSchemaField(unittest.TestCase):
 
     def test___repr__(self):
         field1 = self._make_one("field1", "STRING")
-        expected = "SchemaField('field1', 'STRING', 'NULLABLE', None, (), ())"
+        expected = "SchemaField('field1', 'STRING', 'NULLABLE', None, (), None)"
         self.assertEqual(repr(field1), expected)
+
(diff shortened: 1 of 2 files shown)


tests/unit/test_schema.py
```diff
@@ -510,9 +510,18 @@ def test___hash__not_equals(self):
...
 
+    def test___repr__evaluable(self):
+        field = self._make_one("field1", "STRING", "REQUIRED", "Description")
+        field_repr = repr(field)
+        SchemaField = self._get_target_class()  # needed for eval  # noqa
+
+        evaled_field = eval(field_repr)
+
+        assert field == evaled_field
+
```
tseaver, line 524: For completeness, do we need a round-trip test for the case where there is a policy tuple?

plamut: Would not hurt, I'll add it.

tseaver (approved):

tests/unit/test_schema.py
plamut: Guess what, that extra test actually revealed deeper issues in the `repr()` of `SchemaField` and `PolicyTagList`.

tswast (approved):
Thanks!

tests/unit/test_schema.py
tseaver: W00t!

PR #1050, opened 2021-11-05
deps: support OpenTelemetry >= 1.1.0

### LINKED ISSUE #1038: Bump minimum OpenTelemetry version to support type checks
See the following comment for details: https://github.com/googleapis/python-bigquery/pull/1036#discussion_r739787329

If confirmed, we should bump to at least `opentelemetry-*==1.1.0`, and adjust our OpenTelemetry logic to the changes in the library API.

plamut:
Closes #1038.

This PR makes sure that the tracing logic is compatible with `OpenTelemetry >= 1.1.0` (the first version that supports type checks), and that tests actually fail if a future `python-opentelemetry` version contains backward-incompatible changes in its API (previously the tests would just be skipped).

**PR checklist:**

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -36,16 +36,24 @@ try:
     import pandas
 except (ImportError, AttributeError):  # pragma: NO COVER
     pandas = None
+
 try:
     import opentelemetry
-    from opentelemetry import trace
-    from opentelemetry.sdk.trace import TracerProvider
-    from opentelemetry.sdk.trace.export import SimpleExportSpanProcessor
-    from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
-        InMemorySpanExporter,
-    )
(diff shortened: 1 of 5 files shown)


tests/unit/test_client.py
```diff
@@ -36,16 +36,24 @@
...
+    try:
+        from opentelemetry import trace
+        from opentelemetry.sdk.trace import TracerProvider
+        from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+        from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
+            InMemorySpanExporter,
+        )
+    except (ImportError, AttributeError) as exc:  # pragma: NO COVER
+        msg = "Error importing from opentelemetry, is the installed version compatible?"
+        raise ImportError(msg) from exc
```
tswast, line 55: Good call on raising this here. I'd rather we catch incompatibilities early.

tswast (approved):

PR #1053, opened 2021-11-08
feat: allow cell magic body to be a $variable

### LINKED ISSUE #1040: allow the body of `%%bigquery` cell magic to be a `$variable`

When using BigQuery notebooks. There are queries that would be nice to be parametrized, but the parameters are in places where BigQuery parameterized queries don't work. For example, you can't parametrize the table ID.

When a `$variable` is the only thing in the query body, replace it with the value from the variable in Python. This is similar to how the first `%%bigquery` line works.

Cell 1

```
yourprojectid = "swast-scratch"
query = "SELECT * FROM something"
```

Cell 2

```
%%bigquery --project $yourprojectid
$query
```

One could try and find any `$variable` in the query body, but this would be a breaking change.

 **Additional context**

Currently the `%%bigquery` has some logic to see if the body is a table ID or a full query. We should retain this logic, and it should happen after the replacement of the variable. Detecting if the body is only a `$variable` should be similar to the detection of if the body is only a table ID.

Feature request is from an email thread "python magics - fully parameterized query" with @minhaz and @shollyman

plamut:
Closes #1040.

Thought - we probably want to follow up with a sample demonstrating this new use case.

Also, the length of the `_cell_magic()` function should be ignored for now, it's been waiting to be split into smaller units since forever. :slightly_smiling_face: 

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/magics/magics.py
+++ b/google/cloud/bigquery/magics/magics.py
@@ -600,6 +600,10 @@ def _cell_magic(line, query):
         if query.startswith("$"):
             query_var_name = query[1:]
 
+            if not query_var_name:
+                missing_msg = 'Missing query variable name, empty "$" is not allowed.'
+                raise NameError(missing_msg)
+
             if query_var_name.isidentifier():
                 ip = IPython.get_ipython()
                 query = ip.user_ns.get(query_var_name, ip)  # ip serves as a sentinel
--- a/google/cloud/bigquery/magics/magics.py
+++ b/google/cloud/bigquery/magics/magics.py
@@ -611,8 +611,8 @@ def _cell_magic(line, query):
                 else:
                     if not isinstance(query, (str, bytes)):
                         raise TypeError(
-                            f"Query variable {query_var_name} must be string "
-                            "or bytes-like value."
+                            f"Query variable {query_var_name} must be a string "
+                            "or a bytes-like value."
                         )
 
         # Any query that does not contain whitespace (aside from leading and trailing whitespace)
--- a/google/cloud/bigquery/magics/magics.py
+++ b/google/cloud/bigquery/magics/magics.py
@@ -602,12 +602,12 @@ def _cell_magic(line, query):
 
             if query_var_name.isidentifier():
                 ip = IPython.get_ipython()
-                try:
-                    query = ip.user_ns[query_var_name]
-                except KeyError:
+                query = ip.user_ns.get(query_var_name, ip)  # ip serves as a sentinel
+
+                if query is ip:
                     raise NameError(
                         f"Unknown query, variable {query_var_name} does not exist."
-                    ) from None
+                    )
                 else:
                     if not isinstance(query, (str, bytes)):
                         raise TypeError(
--- a/google/cloud/bigquery/magics/magics.py
+++ b/google/cloud/bigquery/magics/magics.py
@@ -596,6 +596,25 @@ def _cell_magic(line, query):
             _handle_error(error, args.destination_var)
             return
 
+        # Check if query is given as a reference to a variable.
+        if query.startswith("$"):
+            query_var_name = query[1:]
+
+            if query_var_name.isidentifier():
+                ip = IPython.get_ipython()
+                try:
+                    query = ip.user_ns[query_var_name]
+                except KeyError:
+                    raise NameError(
+                        f"Unknown query, variable {query_var_name} does not exist."
+                    ) from None
+                else:
+                    if not isinstance(query, (str, bytes)):
+                        raise TypeError(
+                            f"Query variable {query_var_name} must be string "
+                            "or bytes-like value."
+                        )
+
         # Any query that does not contain whitespace (aside from leading and trailing whitespace)
         # is assumed to be a table id
         if not re.search(r"\s", query):
--- a/tests/unit/test_magics.py
+++ b/tests/unit/test_magics.py
@@ -584,7 +584,7 @@ def test_bigquery_magic_does_not_clear_display_in_verbose_mode():
 
 
 @pytest.mark.usefixtures("ipython_interactive")
-def test_bigquery_magic_clears_display_in_verbose_mode():
+def test_bigquery_magic_clears_display_in_non_verbose_mode():
     ip = IPython.get_ipython()
     ip.extension_manager.load_extension("google.cloud.bigquery")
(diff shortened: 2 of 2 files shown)


google/cloud/bigquery/magics/magics.py
```diff
@@ -596,6 +596,25 @@ def _cell_magic(line, query):
...
+            query_var_name = query[1:]
+
+            if query_var_name.isidentifier():
+                ip = IPython.get_ipython()
+                try:
+                    query = ip.user_ns[query_var_name]
+                except KeyError:
+                    raise NameError(
+                        f"Unknown query, variable {query_var_name} does not exist."
+                    ) from None
```
tseaver, line 610: Should we chain this error, or is that likely to be more confusing than helpful for `IPython` users?

google/cloud/bigquery/magics/magics.py
```diff
@@ -596,6 +596,25 @@ def _cell_magic(line, query):
...
+                else:
+                    if not isinstance(query, (str, bytes)):
+                        raise TypeError(
+                            f"Query variable {query_var_name} must be string "
+                            "or bytes-like value."
```
tseaver, line 615:
```suggestion
                            f"Query variable {query_var_name} must be a string "
                            "or a bytes-like value."
```

google/cloud/bigquery/magics/magics.py
plamut: My thinking was that `KeyError` is just an implementation detail here, as it could easily be written in a LBYL style, too (although that would require one extra dict lookup).

google/cloud/bigquery/magics/magics.py
plamut: Ah, and I already almost feel sorry for committing a suggestion, please appease the CLA bot. :)

google/cloud/bigquery/magics/magics.py
tseaver: You could avoid the extra lookup via:

```python
                query = ip.user_ns.get(query_var_name)
                if query is None:
                    raise NameError(
                        f"Unknown query, variable {query_var_name} does not exist."
                    )
```

tseaver:
@googlebot I consent.

tseaver:
@plamut Sorry, I tried. :-<

plamut:
OK, I'll edit the commit message then to silence the bot.

google/cloud/bigquery/magics/magics.py
plamut: Right... should probably not work such late hours. :upside_down_face: 

(side note - we would still need to distinct between the case when the variable is missing and when the user explicitly set it to `None`, but a sentinel default value would solve that easily)

Anyhow, let me know if there's a strong preference for LBYL over EAFP here and I'll refactor if needed.

tseaver (approved):

tests/unit/test_magics.py
```diff
@@ -1710,6 +1710,117 @@ def test_bigquery_magic_with_improperly_formatted_params():
...
+
+    run_query_patch = mock.patch(
+        "google.cloud.bigquery.magics.magics._run_query", autospec=True
+    )
+
+    ip.user_ns.pop("custom_query", None)  # Make sure the variable does NOT exist.
+    cell_body = "$custom_query"
+
+    with pytest.raises(
+        NameError, match=r".*custom_query does not exist.*"
```
loferris, line 1772: Maybe comment to indicate we're testing the first kind of error, as I got the regex string match from context of the other file.

tests/unit/test_magics.py
```diff
@@ -1710,6 +1710,117 @@ def test_bigquery_magic_with_improperly_formatted_params():
...
+        "google.cloud.bigquery.magics.magics._run_query", autospec=True
+    )
+
+    ipython_ns_cleanup.append((ip, "custom_query"))
+
+    ip.user_ns["custom_query"] = object()
+    cell_body = "$custom_query"
+
+    with pytest.raises(
+        TypeError, match=r".*must be a string or a bytes-like.*"
```
loferris, line 1817: Might be good to have another comment here to match it to the other `else` condition in `run_query_magic`

tests/unit/test_magics.py
```diff
@@ -1710,6 +1710,117 @@ def test_bigquery_magic_with_improperly_formatted_params():
...
+@pytest.mark.usefixtures("ipython_interactive")
+@pytest.mark.skipif(pandas is None, reason="Requires `pandas`")
+def test_bigquery_magic_empty_query_variable_name():
+    ip = IPython.get_ipython()
+    ip.extension_manager.load_extension("google.cloud.bigquery")
+    magics.context.credentials = mock.create_autospec(
+        google.auth.credentials.Credentials, instance=True
+    )
+
+    cell_body = "$"
```
loferris, line 1788: Might be good to have a comment here explaining what this test is catching as it's not explicit in the method. Also, makes me wonder wonder why it wouldn't also trigger the `Unknown variable` error.

loferris (approved):
LGTM - could do with more comments on test cases, perhaps.

tests/unit/test_magics.py
plamut: The name of the test explains which case it covers ("non-existing query variable"), do you think that's not sufficient by itself?

I always look at the name of the test first, although it might be just how my brain is wired...

tests/unit/test_magics.py
plamut: > Also, makes me wonder wonder why it wouldn't also trigger the `Unknown variable` error.

Fair point, I will try to remember what the reasoning (or the lack of it :slightly_smiling_face: ) was and adjust accordingly.

**Update:** It might actually be a fall-through that coincidentally looked reasonable, as `''` is not an identifier. But an `unknown variable` type of error might indeed make more sense.

plamut: Looked at this again and I agree. The missing variable name (i.e. a naked `$`) is now handled explicitly.

I also threw in a few extra comments to explain which case a particular `cell_body` value targets.

loferris:
LGTM - appreciate the changes!!

tests/unit/test_magics.py
loferris: LGTM - I really like how you added the "non-existing query variable" condition.

loferris (approved):

PR #1061, opened 2021-11-15
feat: default to DATETIME type when loading timezone-naive datetimes from Pandas

### LINKED ISSUE #985: v3: default to loading timezone-less pandas datetime columns into a DATETIME BigQuery
Previously, this wasn't supported, but now it is. See: https://github.com/googleapis/python-bigquery/pull/895

To avoid breaking changes, we continue to upload these columns as TIMESTAMP, but we should probably load them as DATETIME for correctness in v3.

plamut:
Closes #985.

This proved to be more tricky than expected, because manual introspection is needed when augmenting the schema - `pyarrow` attaches the UTC timezone to naive datetimes, making it problematic to distinguish these from timezone-aware datetimes.

**PR checklist:**


google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -482,6 +512,19 @@ def augment_schema(dataframe, current_bq_schema):
...
+            # We thus additionally check the actual data to see if we need to overrule
+            # that and choose DATETIME instead.
+            # Note that this should only be needed for datetime values inside a list,
+            # since scalar datetime values have a proper Pandas dtype that allows
+            # distinguishing between timezone-naive and timezone-aware values before
+            # even requiring the additional schema augment logic in this method.
+            if detected_type == "TIMESTAMP":
+                valid_item = _first_array_valid(dataframe[field.name])
+                if isinstance(valid_item, datetime) and valid_item.tzinfo is None:
+                    detected_type = "DATETIME"
```
plamut, line 527: I was thinking of doing this check for all detected TIMESTAMP values, but it turned out it's only necessary for datetimes inside an array, because that's when we need to use `pyarrow` to help.

For datetime values _outside_ of arrays, we can already distinguish between naive and aware ones based on Pandas dtypes, meaning that we do not even enter `augment_schema()` for them.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -379,6 +379,36 @@ def _first_valid(series):
...
+
+    valid_array = series.at[first_valid_index]
+    valid_item = next((item for item in valid_array if not pandas.isna(item)), None)
+
+    if valid_item is not None:
+        return valid_item
+
+    # Valid item is None because all items in the "valid" array are invalid. Try
+    # to find a true valid array manually.
+    for array in islice(series, first_valid_index + 1, None):
```
plamut, line 400: I was not sure if slicing the series results in an unnecessary copy (Pandas docs say it's context-dependent), thus played it safe and just used `islice`.

tswast (approved):
Thanks! Good catch in identifying the additional logic needed for arrays of DATETIME

PR #1062, opened 2021-11-15
chore: update doc links from googleapis.dev to cloud.google.com

dandhlee:
Updating the reference documentation website from googleapis.dev to cloud.google.com for the index pages. Also updating references in the README.

### DIFF
--- a/.repo-metadata.json
+++ b/.repo-metadata.json
@@ -2,7 +2,7 @@
     "name": "bigquery",
     "name_pretty": "Google Cloud BigQuery",
     "product_documentation": "https://cloud.google.com/bigquery",
-    "client_documentation": "https://googleapis.dev/python/bigquery/latest",
+    "client_documentation": "https://cloud.google.com/python/docs/reference/bigquery/latest",
     "issue_tracker": "https://issuetracker.google.com/savedsearches/559654",
     "release_level": "ga",
     "language": "python",
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@ processing power of Google's infrastructure.
 .. |versions| image:: https://img.shields.io/pypi/pyversions/google-cloud-bigquery.svg
    :target: https://pypi.org/project/google-cloud-bigquery/
 .. _BigQuery: https://cloud.google.com/bigquery/what-is-bigquery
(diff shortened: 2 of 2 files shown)


plamut (approved):

plamut:
The change looks good, but there seems to be an unrelated error:
```
Cannot parse: 1:58: def cancel_job(job_id, project=None, location=None, retry=<google.api_core.retry.Retry object>, timeout=None): pass
```

Is that related to the recent change in the generator?

The problem is that `repr()` of the default `timeout` object is not `eval()`-uable, causing this error. We unfortunately cannot simply take the function signature and wrap it into `def ... : pass`.

dandhlee:
>Is that related to the recent change in the generator?

Yes. Two parts, 1) not handling the exception gracefully and 2) unable to resolve some references by the time this signature is extracted. The first problem is handled in https://github.com/googleapis/sphinx-docfx-yaml/pull/152, however I'm unsure whether I can try and format the signature after things are resolved.

plamut:
> however I'm unsure whether I can try and format the signature after things are resolved for the second part of the problem.

When all you have is a string, this might indeed be difficult or even impossible in the general case, as `__repr__()` can return just about anything, even strings that cannot be used to re-construct the instance they are representing.

Even the default `object.__repr__()` does not return an evaluable representation, so... we might be out of luck here. :upside_down_face: 
```py
>>> class Foo: pass
>>> repr(Foo())
'<__main__.Foo object at 0x7fd95d087d00>'
```

dandhlee:
Interestingly, this is how it currently gets shown for googleapis.dev docs as well:
```
cancel_job(job_id: str, project: Optional[str] = None, location: Optional[str] = None, retry: google.api_core.retry.Retry = <google.api_core.retry.Retry object>, timeout: Optional[float] = None)
```
I was hoping it was something that's plugin specific and the docs we have right now would handle it well, seems like I'll need to dig in deeper :S

plamut:
I will nevertheless go ahead and merge this, as the docs-presubmit check is not required and the error is unrelated to the link changes here.

dandhlee:
Thanks for helping look into this! I'll take the findings here and log it to better understand this problem to hopefully be able to solve it 😓

PR #1068, opened 2021-11-17
cleanup: silence non-relevant system test warnings

### LINKED ISSUE #1060: Fix warnings caused by library code when running system tests
Running system tests locally produces the following (under Python 3.8, if it matters, on the `v3` branch):

```
UserWarning: Geometry is in a geographic CRS. Results from 'area' are likely incorrect. Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.
```

The tests should be refactored to not cause these warnings.

FWIW, there is one other Shapely deprecation warning, but that is caused by Pandas core, no our own code.

plamut:
Fixes #1060.

Instead of fighting with geo data and how to correctly compute the area, I opted for the same fix we used in #973.

**PR checklist:**

### DIFF
--- a/tests/system/test_pandas.py
+++ b/tests/system/test_pandas.py
@@ -20,6 +20,7 @@ import decimal
 import json
 import io
 import operator
+import warnings
 
 import google.api_core.retry
 import pkg_resources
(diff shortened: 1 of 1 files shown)


tseaver (approved):

loferris (approved):
LGTM

PR #1070, opened 2021-11-18
fix: apply timeout to all resumable upload requests

### LINKED ISSUE #1067: Provided timeout should be applied to all resumable upload requests
I investigated `ReadTimeout` error that is thrown in `Client.load_table_from_json`.

The error occurred at `ResumableUpload.transmit_next_chunk` in `_do_resumable_upload` in `client.py`.
Lookig at [the code](https://github.com/googleapis/python-bigquery/blob/3a681e046819df18118aa0b2b5733416d004c9b3/google/cloud/bigquery/client.py#L2848-L2853), `timeout` argument is passed to `_initiate_resumable_upload` but not to `transmit_next_chunk`, and [the default timeout](https://github.com/googleapis/google-resumable-media-python/blob/d0827725085ab064169cda004a4fde65e92813f0/google/resumable_media/requests/_request_helpers.py#L35) (which is 60 sec) will be used.
Is this intentional, or does anyone plan to fix it?
Thank you.

plamut:
Fixes #1067.

**PR checklist:**

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -31,9 +31,10 @@ import tempfile
 import typing
 from typing import (
     Any,
-    BinaryIO,
     Dict,
+    IO,
     Iterable,
+    Mapping,
     List,
     Optional,
     Sequence,
@@ -112,10 +113,14 @@ from google.cloud.bigquery import _helpers
 pyarrow = _helpers.PYARROW_VERSIONS.try_import()
 
 TimeoutType = Union[float, None]
+ResumableTimeoutType = Union[
+    None, float, Tuple[float, float]
+]  # for resumable media methods
 
 if typing.TYPE_CHECKING:  # pragma: NO COVER
     # os.PathLike is only subscriptable in Python 3.9+, thus shielding with a condition.
     PathType = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
+    import pandas  # type: ignore
 
 _DEFAULT_CHUNKSIZE = 100 * 1024 * 1024  # 100 MB
 _MAX_MULTIPART_SIZE = 5 * 1024 * 1024
@@ -2348,7 +2353,7 @@ class Client(ClientWithProject):
 
     def load_table_from_file(
         self,
-        file_obj: BinaryIO,
+        file_obj: IO[bytes],
         destination: Union[Table, TableReference, TableListItem, str],
         rewind: bool = False,
         size: int = None,
@@ -2358,7 +2363,7 @@ class Client(ClientWithProject):
         location: str = None,
         project: str = None,
         job_config: LoadJobConfig = None,
-        timeout: TimeoutType = DEFAULT_TIMEOUT,
+        timeout: ResumableTimeoutType = DEFAULT_TIMEOUT,
     ) -> job.LoadJob:
         """Upload the contents of this table from a file-like object.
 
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/client.py
```diff
@@ -2850,7 +2850,7 @@ def _do_resumable_upload(
         )
 
         while not upload.finished:
-            response = upload.transmit_next_chunk(transport)
+            response = upload.transmit_next_chunk(transport, timeout=timeout)
```
tseaver, line 2853: FWIW, the [`timeout` argument](https://github.com/googleapis/google-resumable-media-python/blob/d0827725085ab064169cda004a4fde65e92813f0/google/resumable_media/_upload.py#L753-L759) can actually be a 2-tuple as well:

```
            timeout (Optional[Union[float, Tuple[float, float]]]):
                The number of seconds to wait for the server response.
                Depending on the retry strategy, a request may be repeated
                several times using the same timeout each time.

                Can also be passed as a tuple (connect_timeout, read_timeout).
                See :meth:`requests.Session.request` documentation for details.
```

Maybe we should expose that in the docstring for this method, and its one caller (`Client.load_table_from_file`).

tswast:
Test failure:

```
____________________________ test_upload_chunksize _____________________________

client = <google.cloud.bigquery.client.Client object at 0x7f73ab1c0710>

    def test_upload_chunksize(client):
        with mock.patch("google.cloud.bigquery.client.ResumableUpload") as RU:
            upload = RU.return_value

            upload.finished = False

            def transmit_next_chunk(transport):
                upload.finished = True
                result = mock.MagicMock()
                result.json.return_value = {}
                return result

            upload.transmit_next_chunk = transmit_next_chunk
            f = io.BytesIO()
>           client.load_table_from_file(f, "foo.bar")

tests/unit/test_client.py:8469:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
google/cloud/bigquery/client.py:2443: in load_table_from_file
    file_obj, job_resource, num_retries, timeout, project=project
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <google.cloud.bigquery.client.Client object at 0x7f73ab1c0710>
stream = <_io.BytesIO object at 0x7f73ab14f048>
metadata = {'configuration': {'load': {'destinationTable': {'datasetId': 'foo', 'projectId': 'PROJECT', 'tableId': 'bar'}}}, 'jobReference': {'jobId': 'c2ad6c61-6202-4c0e-a6af-8310c4aabd61', 'projectId': 'PROJECT'}}
num_retries = 6, timeout = None, project = 'PROJECT'

    def _do_resumable_upload(
        self, stream, metadata, num_retries, timeout, project=None
    ):
        """Perform a resumable upload.

        Args:
            stream (IO[bytes]): A bytes IO object open for reading.

            metadata (Dict): The metadata associated with the upload.

            num_retries (int):
                Number of upload retries. (Deprecated: This
                argument will be removed in a future release.)

            timeout (float):
                The number of seconds to wait for the underlying HTTP transport
                before using ``retry``.

            project (Optional[str]):
                Project ID of the project of where to run the upload. Defaults
                to the client's project.

        Returns:
            requests.Response:
                The "200 OK" response object returned after the final chunk
                is uploaded.
        """
        upload, transport = self._initiate_resumable_upload(
            stream, metadata, num_retries, timeout, project=project
        )

        while not upload.finished:
>           response = upload.transmit_next_chunk(transport, timeout=timeout)
E           TypeError: transmit_next_chunk() got an unexpected keyword argument 'timeout'

google/cloud/bigquery/client.py:2853: TypeError
=============================== warnings summary ===============================
```

I think we'll have to bump the minimum version of the resumable media library.

tseaver:
@tswast The feature was added (by @plamut) in June 2020, and released with version 0.6.0 of `google-resumable-media`, which is our minimum version.

The test failure comes from the [broken stub version in the testcase](https://github.com/googleapis/python-bigquery/blob/1b5dc5ce1c6419233d9564aa32edbdee523f0f10/tests/unit/test_client.py#L8445-L8449).

google/cloud/bigquery/client.py
plamut: Makes sense, done.

google/cloud/bigquery/client.py
```diff
@@ -2358,50 +2363,50 @@ def load_table_from_file(
...
                 Extra configuration options for the job.
-            timeout (Optional[float]):
+            timeout:
                 The number of seconds to wait for the underlying HTTP transport
-                before using ``retry``.
+                before using ``retry``. Depending on the retry strategy, a request
+                may be repeated several times using the same timeout each time.
+
+                Can also be passed as a tuple (connect_timeout, read_timeout).
+                See :meth:`requests.Session.request` documentation for details.
```
tswast, line 2409: I'm curious, does the same apply to our other API requests? I know we use `requests` library for those too, but via `google-cloud-core`.

google/cloud/bigquery/client.py
```diff
@@ -2819,60 +2827,77 @@ def load_table_from_json(
...
 
     def _do_resumable_upload(
-        self, stream, metadata, num_retries, timeout, project=None
+        self,
+        stream: IO[bytes],
+        metadata: Mapping[str, str],
+        num_retries: int,
+        timeout: Optional[ResumableTimeoutType],
+        project: Optional[str] = None,
     ):
```
tswast, line 2836: Missing return type, which has been removed from the docstring.

google/cloud/bigquery/client.py
plamut: From what I can see, it does. The `timeout` argument gets passed down to `google.cloud._http.JSONConnection`, which [accepts](https://github.com/googleapis/python-cloud-core/blob/24b7de49943a49e8235f9dbee6b32693deed8c1f/google/cloud/_http/__init__.py#L442-L447) a two-tuple.

I _think_ we did not advertise this second option in BigQuery and just went with `Optional[Union[int, float]]` to avoid leaking transport implementation details, but I'm not 100%.

tseaver: `google-cloud-core` [definitely](https://github.com/googleapis/python-cloud-core/blob/24b7de49943a49e8235f9dbee6b32693deed8c1f/google/cloud/_http/__init__.py#L375-L447) exposes both `float` and `tuple` for `timeout`.

Note that its current docs link (under `cloud.google.com`) doesn't show those details.

tswast (approved):

PR #1075, opened 2021-11-30
feat: support authorized dataset entity

### LINKED ISSUE #1064: Support Authorized Dataset
https://github.com/googleapis/python-bigquery/blob/705205430527f4c736fa4ad94b58e8cbb625fc11/google/cloud/bigquery/dataset.py#L134-L144

Not sure about the reason to have restricted entities here.
We need to use authorized dataset, and got ValueError.
image

steffnay:
Removes the client-side validation of entity types and replaces them with enum values.

Fixes #1064  🦕


google/cloud/bigquery/enums.py
```diff
@@ -232,6 +232,19 @@ def _make_sql_scalars_enum():
 StandardSqlDataTypes = _make_sql_scalars_enum()
 
 
+class EntityTypes:
```
tswast, line 235: I think we can make this a true enum like we did some of the other classes in this module.

```suggestion
class EntityTypes(str, enum.Enum):
```

tests/unit/test_dataset.py
```diff
@@ -141,18 +141,35 @@ def test_to_api_repr_routine(self):
...
+            self._make_one("READER", "dataset", dataset)
+
     def test_from_api_repr(self):
         resource = {"role": "OWNER", "userByEmail": "salmon@example.com"}
         entry = self._get_target_class().from_api_repr(resource)
         self.assertEqual(entry.role, "OWNER")
         self.assertEqual(entry.entity_type, "userByEmail")
         self.assertEqual(entry.entity_id, "salmon@example.com")
 
-    def test_from_api_repr_w_unknown_entity_type(self):
```
tswast, line 151: Could we update this to check that it successfully creates an AccessEntry? I'm a bit worried still that unknown types will cause runtime errors if the backend ever adds another access type.

We may want to cache the resource from `from_api_repr` in a `_properties` value as a stepping stone towards https://github.com/googleapis/python-bigquery/issues/1077 even if we don't use it for all the properties (yet).

google/cloud/bigquery/dataset.py
```diff
@@ -77,28 +77,29 @@ def _get_routine_reference(self, routine_id):
 class AccessEntry(object):
```
tswast, line 77: I notice this class uses an older pattern of hidden properties such as `_role` and `_entity_type`. I'd much prefer if it followed our `_properties` pattern that most of our other classes use. That way it'll be flexible to new properties in the future (such as during a private preview before we have a chance to implement the feature).

Perhaps best left as a follow-up PR so as not to block the customers who need it shorter term. https://github.com/googleapis/python-bigquery/issues/1077 Let me know if you're interested in that somewhat larger refactor.

tswast (changes requested):

tswast:
Let's make sure we update the code samples with the new enum too.

https://github.com/googleapis/python-bigquery/blob/679e8edc27f24a312243f5587e1c71798902fce8/samples/snippets/authorized_view_tutorial.py#L109

https://github.com/googleapis/python-bigquery/blob/effd6734a08914216482ba370d9433364d278f3a/samples/snippets/update_dataset_access.py#L40

google/cloud/bigquery/dataset.py
```diff
@@ -241,7 +235,10 @@ def from_api_repr(cls, resource: dict) -> "AccessEntry":
         entity_type, entity_id = entry.popitem()
         if len(entry) != 0:
             raise ValueError("Entry has unexpected keys remaining.", entry)
-        return cls(role, entity_type, entity_id)
+
+        config = cls(role, entity_type, entity_id)
+        config._properties = copy.deepcopy(resource)
```
steffnay, line 240: @tswast does it make sense to do it this way? I can't tell if it's better to do something like
```
        entry = resource.copy()
        role = entry.pop("role", None)
        entity_type = entity_id = None       
        if entry:
            entity_type, entity_id = entry.popitem()
            if len(entry) != 0:
                raise ValueError("Entry has unexpected keys remaining.", entry)
        config = cls(role, entity_type, entity_id)
        return config
  ```

Comment (steffnay) [reply google/cloud/bigquery/dataset.py:240]: For some reason, fixing this issue in a way that anticipates the changes requested in #1077 is really throwing me off since I'm not familiar with this library.

Comment (tswast) [reply google/cloud/bigquery/dataset.py:240]: That works. I think it could be simpler than that and avoid the one of the two `deepcopy` calls by avoiding `pop` and `popitem`, but I'm okay leaving those as-is until we do the larger refactor.

Comment (tswast) [reply google/cloud/bigquery/dataset.py:240]: Note: we still need to update `to_api_repr` in this PR.

File: tests/unit/test_dataset.py [hunk 5]
```diff
@@ -150,8 +172,10 @@ def test_from_api_repr(self):
 
     def test_from_api_repr_w_unknown_entity_type(self):
```
tswast, line 173: Could we also add a `from_api_repr` -> `to_api_repr` round-trip test with unknown entity type (and even additional properties that aren't the entity type)? Could be an extra set of assertions on this test or a separate test.

`to_api_repr` is the most important piece, as that's what determines what actually gets sent to the API.

tswast (changes requested):

tests/unit/test_dataset.py
steffnay: Updated

google/cloud/bigquery/dataset.py
steffnay: Added

tswast (approved):

Issue #1074, opened 2021-11-23
Run mypy (and pytype?) on code samples


tswast:
So that we have some assurance that our customers can use mypy / pytype successfully from our library, I think we should run mypy / pytype on them as well.

Ideally we'd do this from the Samples Kokoro jobs, as that'll be more similar to the development environment that a customer has setup.

Step 0.

- Add type hints to our code samples. (Maybe include in our root mypy/pytype noxfile config?)

Step 1.

- Add optional pytype / mypy sessions to https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_samples/noxfile.py.j2

Step 2.

- Enable this feature in our samples, probably via the noxfile_config.py files.

tswast:
@plamut I kinda consider this a v3 blocker. I'd really like to make sure our type hints are working from the end-user perspective before we start publishing them.

plamut:
@tswast OK, I can start working on this as soon as I'm done with a similar task in Pub/Sub. It's quite tedious there due to a much more dynamic nature of the code and metaprogramming hacks, but I beleive I'll be done soon.

tswast:
TIL, there is a `enforce_type_hints` flag already https://github.com/GoogleCloudPlatform/python-docs-samples/blob/14533af57081783f3671fa6ac0386fea5f6ebc7f/noxfile_config.py#L26-L28

But it only checks that annotations are there, not that they actually work.

plamut:
> But it only checks that annotations are there, not that they actually work.

LGTM! :upside_down_face: :grin:

plamut:
Random observation: We need to annotate return types of our `__init__()` methods.

For example, when checking the samples with more strict options, the following innocent line resulted in a `mypy` error:
```py
table = bigquery.Table(table_id)
```
Error:
```
... Call to untyped function "Table" in typed context
```

The following seems to be required to appease `mypy`:
```patch
diff --git google/cloud/bigquery/table.py google/cloud/bigquery/table.py
index f434688..5b96369 100644
--- google/cloud/bigquery/table.py
+++ google/cloud/bigquery/table.py
@@ -369,7 +369,7 @@ class Table(_TableBase):
         "require_partition_filter": "requirePartitionFilter",
     }
 
-    def __init__(self, table_ref, schema=None):
+    def __init__(self, table_ref, schema=None) -> "Table":
         table_ref = _table_arg_to_table_ref(table_ref)
[...]

PR #1081, opened 2021-12-08
process: add mypy-valid type annotations to samples

plamut:
Towards #1074.

This PR adds type annotations to samples and fixes quite a few issues discovered along. It also adds a new nox session to run type checks with mypy on samples.

The `mypy.ini` files in `samples/*` subdirectories are added for the upcoming type-check sessions that will be added to the noxfiles in these directories (through the code generator).

**PR checklist:**


plamut:
> ImportError: cannot import name 'TypedDict'

This was only added to `typing` in Python 3.8. I actually fixed this, but not at all places, and I was only running the checks with Python 3.8, missing this. :upside_down_face:

samples/geography/insert_geojson.py
```diff
@@ -12,14 +12,21 @@
...
+) -> Sequence[Dict[str, object]]:
     # [START bigquery_insert_geojson]
     import geojson
     from google.cloud import bigquery
 
     bigquery_client = bigquery.Client()
 
+    if override_values is None:
+        override_values = {}
+
```
tswast, line 29: Let's move the `override_values` outside of the actual sample part.

```suggestion
    if override_values is None:
        override_values = {}

    # [START bigquery_insert_geojson]
    import geojson
    from google.cloud import bigquery

    bigquery_client = bigquery.Client()

```

Alternatively, we can put it after `# [END bigquery_insert_geojson]`, but GitHub won't let me suggest modifications including line 33.

Alternatively again, we could switch to the `table_id_original` pattern we've started using in some other samples.

samples/geography/insert_wkt.py
```diff
@@ -12,15 +12,22 @@
...
     # [START bigquery_insert_geography_wkt]
     from google.cloud import bigquery
     import shapely.geometry
     import shapely.wkt
 
     bigquery_client = bigquery.Client()
 
+    if override_values is None:
+        override_values = {}
+
```
tswast, line 30:
```suggestion
   if override_values is None:
        override_values = {}

    # [START bigquery_insert_geography_wkt]
    from google.cloud import bigquery
    import shapely.geometry
    import shapely.wkt

    bigquery_client = bigquery.Client()

```

samples/query_external_gcs_temporary_table.py
```diff
@@ -30,7 +31,7 @@ def query_external_gcs_temporary_table():
         bigquery.SchemaField("name", "STRING"),
         bigquery.SchemaField("post_abbr", "STRING"),
     ]
-    external_config.options.skip_leading_rows = 1
+    typing.cast(bigquery.CSVOptions, external_config.options).skip_leading_rows = 1
```
tswast, line 34: We now have a [`csv_options` property](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.external_config.ExternalConfig.html#google.cloud.bigquery.external_config.ExternalConfig.csv_options), where this type cast would be unnecessary.

```suggestion
    external_config.csv_options.skip_leading_rows = 1
```

Looks like we are careful to ensure that sub-setters are possible, since we use the same `_properties` reference. https://github.com/googleapis/python-bigquery/blob/44221f437f93fae2fac10d074b39a28008b9ad01/google/cloud/bigquery/external_config.py#L891

Edit: We might still need something since the type is annotated as `Optional`. I wonder if we should remove the `if self.source_format == ....` logic and always do `setdefault` ? It means we don't have client-side protections against setting values for multiple source formats, but I'm okay letting the server-side do that validation.

samples/query_external_sheets_permanent_table.py
```diff
@@ -56,8 +57,9 @@ def query_external_sheets_permanent_table(dataset_id):
         "/d/1i_QCL-7HcSyUZmIbP9E6lO_T5u3HnpLe7dnpHaijg_E/edit?usp=sharing"
     )
     external_config.source_uris = [sheet_url]
-    external_config.options.skip_leading_rows = 1  # Optionally skip header row.
-    external_config.options.range = (
+    options = typing.cast(bigquery.GoogleSheetsOptions, external_config.options)
```
tswast, line 60:
```suggestion
    options = external_config.google_sheets_options
```

samples/query_external_sheets_temporary_table.py
```diff
@@ -53,8 +54,9 @@ def query_external_sheets_temporary_table():
         bigquery.SchemaField("name", "STRING"),
         bigquery.SchemaField("post_abbr", "STRING"),
     ]
-    external_config.options.skip_leading_rows = 1  # Optionally skip header row.
-    external_config.options.range = (
+    options = typing.cast(bigquery.GoogleSheetsOptions, external_config.options)
```
tswast, line 57:
```suggestion
    options = external_config.options
```

samples/snippets/update_with_dml.py
```diff
@@ -60,10 +62,13 @@ def update_with_dml(
     query_job.result()
 
     print(f"DML query modified {query_job.num_dml_affected_rows} rows.")
-    return query_job.num_dml_affected_rows
+    return typing.cast(int, query_job.num_dml_affected_rows)
```
tswast, line 65: I assume this is because `num_dml_affected_rows` is `Optional`? Could we change the function return type instead and avoid the cast?

tswast (changes requested):
So cool!

A few nits. Plus I'd to see some improvements to the `ExternalConfig` class to see if we can avoid `casts` in those samples.

samples/geography/insert_geojson.py
loferris: +1

plamut: I can't find any references to `table_id_original`?

I will thus just move the `is None` check outside of the sample.

samples/snippets/update_with_dml.py
plamut: Yes, that's the reason that `mypy` (rightfully) complains.

Since `num_dml_affected_rows` can actually be `None`,  I don't think we can just change the return type. But there is another way, possibly better - if we explicitly assert that the target value is not `None`, mypy will understand it, and the cast will not be needed anymore.

```py
assert query_job.num_dml_affected_rows is not None
return query_job.num_dml_affected_rows  # passes the type check
```

samples/query_external_gcs_temporary_table.py
plamut: How about sticking in an explicit assertion `assert foo is not None`? Mypy understands that and we can get rid of casting the optional values, which cannot be `None` in a particular runtime context.

plamut:
@tswast Since there was a new sample added to `main`, I merged it into this branch. Had to resolve some conflicts, thus that commit should also be checked.

The rest of the comments have been addressed, and I got rid of several `cast()` calls by using explicit `is None` assertions instead - let me know how that looks to you.

tswast (approved):
Awesome!

It's a bit funny to have assertions in the code samples, but not too out of place. It looks more natural than the type casts.

plamut:
> It's a bit funny to have assertions in the code samples, but not too out of place. It looks more natural than the type casts.

True that, but I agree that's much better than those rather ugly casts.

PR #1090, opened 2021-12-16
test: skip failing copy table tests

tswast:
Workaround for internal issue 210907595

### DIFF
--- a/samples/tests/test_copy_table.py
+++ b/samples/tests/test_copy_table.py
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import pytest
+
 from .. import copy_table
 
 
 def test_copy_table(capsys, table_with_data_id, random_table_id, client):
+    pytest.skip("b/210907595: copy fails for shakespeare table")
 
     copy_table.copy_table(table_with_data_id, random_table_id)
     out, err = capsys.readouterr()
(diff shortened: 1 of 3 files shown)


stephaniewang526 (approved):

tswast:
Three test failures all in the same spot:

```
___________ TestBigQuery.test_dbapi_connection_does_not_leak_sockets ___________

self = <tests.system.test_client.TestBigQuery testMethod=test_dbapi_connection_does_not_leak_sockets>

    @unittest.skipIf(
        bigquery_storage is None, "Requires `google-cloud-bigquery-storage`"
    )
    def test_dbapi_connection_does_not_leak_sockets(self):
        current_process = psutil.Process()
        conn_count_start = len(current_process.connections())

        # Provide no explicit clients, so that the connection will create and own them.
        connection = dbapi.connect()
        cursor = connection.cursor()

        cursor.execute(
            """
            SELECT id, `by`, time_ts
            FROM `bigquery-public-data.hacker_news.comments`
            ORDER BY `id` ASC
            LIMIT 100000
        """
        )
        rows = cursor.fetchall()
        self.assertEqual(len(rows), 100000)

        connection.close()
        conn_count_end = len(current_process.connections())
>       self.assertEqual(conn_count_end, conn_count_start)
E       AssertionError: 3 != 7

tests/system/test_client.py:1728: AssertionError
```

Possibly just some flakiness with how long it takes to close a read API grpc connection?

PR #1099, opened 2022-01-06
test: make GC check resilient to broken weakrefs

msuozzo:
In some python deployments (I believe macOS is one), weakref objects aren't guaranteed to exist across gc collections. The proposed check pattern is also used in tensorflow ([link]) which ran into similar issues a few years ago.

[link]: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/errors_test.py#L33

### DIFF
--- a/tests/unit/test_dbapi_connection.py
+++ b/tests/unit/test_dbapi_connection.py
@@ -219,8 +219,14 @@ class TestConnection(unittest.TestCase):
         # Connections should not hold strong references to the Cursor instances
         # they created, unnecessarily keeping them alive.
         gc.collect()
-        cursors = [obj for obj in gc.get_objects() if isinstance(obj, Cursor)]
-        self.assertEqual(len(cursors), 2)
+        cursor_count = 0
+        for obj in gc.get_objects():
+          try:
+            if isinstance(obj, Cursor):
+              cursor_count += 1
+          except ReferenceError:
+            pass
+        self.assertEqual(cursor_count, 2)
 
     def test_commit(self):
         connection = self._make_one(client=self._mock_client())
--- a/tests/unit/test_dbapi_connection.py
+++ b/tests/unit/test_dbapi_connection.py
@@ -221,11 +221,11 @@ class TestConnection(unittest.TestCase):
         gc.collect()
         cursor_count = 0
         for obj in gc.get_objects():
-          try:
-            if isinstance(obj, Cursor):
-              cursor_count += 1
-          except ReferenceError:  # pragma: NO COVER
-            pass
+            try:
+                if isinstance(obj, Cursor):
+                    cursor_count += 1
+            except ReferenceError:  # pragma: NO COVER
+                pass
         self.assertEqual(cursor_count, 2)
 
     def test_commit(self):
--- a/tests/unit/test_dbapi_connection.py
+++ b/tests/unit/test_dbapi_connection.py
@@ -224,7 +224,7 @@ class TestConnection(unittest.TestCase):
           try:
             if isinstance(obj, Cursor):
               cursor_count += 1
-          except ReferenceError:
+          except ReferenceError:  # pragma: NO COVER
             pass
         self.assertEqual(cursor_count, 2)
 


tests/unit/test_dbapi_connection.py
```diff
@@ -219,8 +219,14 @@ def test_does_not_keep_cursor_instances_alive(self):
...
         # they created, unnecessarily keeping them alive.
         gc.collect()
-        cursors = [obj for obj in gc.get_objects() if isinstance(obj, Cursor)]
-        self.assertEqual(len(cursors), 2)
+        cursor_count = 0
+        for obj in gc.get_objects():
+          try:
+            if isinstance(obj, Cursor):
+              cursor_count += 1
+          except ReferenceError:
```
tswast, line 227: I think we might need some `# pragma: NO COVER` comments on lines/branches that aren't guaranteed to be hit each test run.

https://github.com/googleapis/python-bigquery/blob/82e0a7c231e823b025f7fa5b81c4695ced0f73ba/.coveragerc#L11

tswast (changes requested):
Thanks for the contribution! I hadn't noticed any flakes when running the test suite locally on my Mac, but could just be a rare thing.

tests/unit/test_dbapi_connection.py
msuozzo: done

msuozzo:
Any idea what the conventional commit error is?

plamut:
> Any idea what the conventional commit error is?

This is the output, accessible by clicking the "Details" link next to the failed check (is it displayed for you?):

```
Linting errors for "make gc check resilient to broken weakrefs"
    - subject may not be empty
    - type may not be empty
```

tests/unit/test_dbapi_connection.py
```diff
@@ -219,8 +219,14 @@ def test_does_not_keep_cursor_instances_alive(self):
...
-        cursors = [obj for obj in gc.get_objects() if isinstance(obj, Cursor)]
-        self.assertEqual(len(cursors), 2)
+        cursor_count = 0
+        for obj in gc.get_objects():
+          try:
+            if isinstance(obj, Cursor):
+              cursor_count += 1
+          except ReferenceError:  # pragma: NO COVER
+            pass
+        self.assertEqual(cursor_count, 2)
```
plamut, line 229: Looking at the linked approach TensorFlow took, this looks good to me.

Out of curiosity and because I cannot reproduce the test failure locally (I'm on Linux), I'd just like to better understand what the issue is here.

Specifically, why would the check `isinstance(obj, Cursor)` raise a `ReferenceError` if a weak reference to `obj` does not exist anymore? And it's only an issue with the test, not the `Connection` code itself?

plamut (commented):
LGTM, just have a clarification question, thanks!

**Edit:** Please just fix the indentation (4 spaces should be used) to conform to this project's style guide, which is why one of the checks failed:

```
nox > flake8 tests
tests/unit/test_dbapi_connection.py:224:11: E111 indentation is not a multiple of 4
tests/unit/test_dbapi_connection.py:226:15: E111 indentation is not a multiple of 4
tests/unit/test_dbapi_connection.py:227:11: E111 indentation is not a multiple of 4
nox > Command flake8 tests failed with exit code 1
nox > Session lint failed.
```

tests/unit/test_dbapi_connection.py
msuozzo: Check out [weakrefobject.c](https://github.com/python/cpython/blob/main/Objects/weakrefobject.c#L416-L437).

The weakref `obj` is unwrapped when calling `isinstance` and if the object has been purged, it will fail. So this should be unrelated to `Cursor` or any of the code here but can be triggered by dependent libraries doing funny things with objects (think protos).

plamut: > The weakref obj is unwrapped when calling isinstance

You mean `weakref.proxy`? Because when initially locally testing with a dead `weakref.weakref`, it simply returned `None` and `isinstance(wref, Foo)` returned `False` (no `ReferenceError`). However, the error is indeed reproducible when using a dead `weakref.proxy` instead.

I suppose `obj` returned by `gc.get_objects()` is actually a proxy, which explains the behavior. Thanks!

plamut (approved):

plamut:
@tswast I don't have write access anymore, please re-run the tests and merge the fix, thanks!

tswast (approved):

PR #1125, opened 2022-01-31
feat: refactor AccessEntry to use _properties pattern

### LINKED ISSUE #1077: refactor AccessEntry to use `_properties` pattern

Often there is a private preview feature (or just a new feature that we haven't implemented yet) and we want to provide our customers a workaround to be able to send a resource as represented by JSON. Then we could have provided a workaround to the user in https://github.com/googleapis/python-bigquery/issues/1064 like the following:

```
entry = AccessEntry.from_api_repr({
  "role": "READER",
  "dataset": {
    "projectId": "project-id",
    "datasetId": "dataset_id",
  }
})
```

or even

```
entry = AccessEntry("READER")
entry._properties["dataset"] = {
    "projectId": "project-id",
    "datasetId": "dataset_id",
}
```

and been confident that it would send the correct values when making the API request.

Likewise, one could read values from the API like `entry._properties["dataset"]` when checking values from the API.

The way we handle `entity_type` and `entity_id` is reminiscent of `ExternalConfig.options`, which I refactored in https://github.com/googleapis/python-bigquery/pull/994. It was clever at the time when there were only a few external data formats, but has diverged far the actual API representation.

`AccessEntry` should have separate properties for `view`, `routine`, `dataset`, etc. These could accept and return relevant types (e.g. TableReference) as well.

Existing `entity_type` and `entity_id` should be made optional in the constructor, but in the same order so that backwards compatibility is maintained.

Leaving the class as-is should continue to work, but it's a significant risk IMO.

 **Additional context**

As identified in https://github.com/googleapis/python-bigquery/pull/1075

See similar work in https://github.com/googleapis/python-bigquery/pull/994

steffnay:

Fixes #1077  🦕


google/cloud/bigquery/dataset.py
```diff
@@ -145,34 +288,139 @@ class AccessEntry(object):
...
+            if isinstance(value, (Dataset, DatasetListItem)):
+                value = value.reference
+
+            value = value.to_api_repr()
+
+        prop = {
+            "dataset": value,
+            "targetTypes": self._properties.get("targetTypes", None),
+        }
+        self._properties["dataset"] = prop
```
tswast, line 330: We have some helper methods (https://github.com/googleapis/python-bigquery/blob/8712b6f4f6c91b0f5aca3de98efe81deb2c7262f/google/cloud/bigquery/_helpers.py#L771) to make safe setting of sub-properties easier. I think we might have got `targetTypes` preservation wrong here.

```suggestion
        _helpers._set_sub_prop(self._properties, ["dataset", "dataset"], prop)
```

google/cloud/bigquery/dataset.py
```diff
@@ -145,34 +288,139 @@ class AccessEntry(object):
...
+
+    @property
+    def target_types(self) -> Optional[List[str]]:
+        """Which resources that the dataset in this entry applies to."""
+        return self._properties["dataset"]["targetTypes"]
```
tswast, line 335: We have a helper (https://github.com/googleapis/python-bigquery/blob/8712b6f4f6c91b0f5aca3de98efe81deb2c7262f/google/cloud/bigquery/_helpers.py#L722) to make getting sub-properties safer as well.

```suggestion
        return _helpers._get_sub_prop(self._properties, ["dataset", "targetTypes"])
```

google/cloud/bigquery/dataset.py
```diff
@@ -145,34 +288,139 @@ class AccessEntry(object):
...
+
+    @target_types.setter
+    def target_types(self, value):
+        if "dataset" not in self._properties:
+            self._properties["dataset"] = {}
+        self._properties["dataset"]["targetTypes"] = value
```
tswast, line 341: Let's use `_set_sub_prop` here, too.

google/cloud/bigquery/dataset.py
```diff
@@ -145,34 +288,139 @@ class AccessEntry(object):
...
+                value = DatasetReference.from_string(value)
+
+            if isinstance(value, (Dataset, DatasetListItem)):
+                value = value.reference
+
+            value = value.to_api_repr()
+
+        prop = {
+            "dataset": value,
+            "targetTypes": self._properties.get("targetTypes", None),
```
tswast, line 328: ~Maybe we want the default `targetTypes` to be `["VIEW"]` instead of `None`?~

Edit: Probably not. That might be confusing if/when more targetTypes are added.

tswast (changes requested):
Looking pretty good! I made a few suggestions to use `get(...)`, `_get_sub_prop`, and `_set_sub_prop`.

I think we'll need some more unit tests that exercise getting the various properties when they aren't set. Make sure they get `None` (or whatever the desired default is) instead of raising a `KeyError`.

google/cloud/bigquery/dataset.py
steffnay: Ah, got it! I was doing a very Node.js thing.

google/cloud/bigquery/dataset.py
```diff
@@ -145,34 +288,139 @@ class AccessEntry(object):
...
-                )
-        else:
-            if role is None:
-                raise ValueError(
-                    "Role must be set for entity " "type %r" % (entity_type,)
-                )
-        self._role = role
+        if entity_type is not None:
+            self._properties[entity_type] = entity_id
+        self._role = self._properties["role"] = role
```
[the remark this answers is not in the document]
steffnay: Nope, it's a goner.

google/cloud/bigquery/dataset.py
```diff
@@ -145,34 +288,139 @@ class AccessEntry(object):
...
         self._entity_type = entity_type
         self._entity_id = entity_id
```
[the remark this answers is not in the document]
steffnay: Yeah, that's way better.

google/cloud/bigquery/dataset.py
```diff
@@ -74,6 +76,147 @@ def _get_routine_reference(self, routine_id):
     )
 
 
+class DatasetReference(object):
```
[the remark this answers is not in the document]
steffnay: Yes, exactly.

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +217,281 @@ def __repr__(self):
...
+                value = RoutineReference.from_string(value)
+
+            if isinstance(value, Routine):
+                value = value.reference
+
+            value = value.to_api_repr()
+        self._properties["routine"] = value
+
+    @property
+    def view(self) -> Optional[dict]:
```
tswast, line 365: nit: we tend to use type `Dict` where we can.

```suggestion
    def view(self) -> Optional[Dict[str, Any]]:
```

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +217,281 @@ def __repr__(self):
...
+        """Construct the API resource representation of this access entry
+
+        Returns:
+            Dict[str, object]: Access entry represented as an API resource
+        """
+        resource = copy.deepcopy(self._properties)
+        if self._entity_type is not None:
+            resource[self.entity_type] = self.entity_id
+        if self.role is not None:
+            resource["role"] = self.role
```
tswast, line 464: I think this logic is unnecessary now that we've refactored, right?

```suggestion
```

tswast (changes requested):
Much better! A few small things

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,282 @@ def __repr__(self):
...
+        )
+
+    @target_types.setter
+    def target_types(self, value):
+        if "dataset" not in self._properties:
+            self._properties["dataset"] = {}
+        _helpers._set_sub_prop(self._properties, ["dataset", "targetTypes"], value)
+
+    @property
+    def routine(self) -> Optional[Dict[str, Any]]:
```
tswast, line 354: Likewise, perhaps we should be returning a `RoutineReference` here? If folks want the json-like dictionary, they can use `entity_id`.

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,282 @@ def __repr__(self):
...
+                value = RoutineReference.from_string(value)
+
+            if isinstance(value, Routine):
+                value = value.reference
+
+            value = value.to_api_repr()
+        self._properties["routine"] = value
+
+    @property
+    def view(self) -> Optional[Dict[str, Any]]:
```
tswast, line 375: `TableReference` here?

tswast (changes requested):
Looking great! A few suggestions on a perhaps more Pythonic design.

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,282 @@ def __repr__(self):
...
+
+        _helpers._set_sub_prop(self._properties, ["dataset", "dataset"], value)
+        _helpers._set_sub_prop(
+            self._properties,
+            ["dataset", "targetTypes"],
+            self._properties.get("targetTypes"),
+        )
+
+    @property
+    def target_types(self) -> Optional[List[str]]:
```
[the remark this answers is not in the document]
steffnay: I like that idea.

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,293 @@ def __repr__(self):
...
+
+    @role.setter
+    def role(self, value):
+        self._properties["role"] = value
+
+    @property
+    def dataset(self) -> Optional[DatasetReference]:
+        """API resource representation of a dataset reference."""
+        return typing.cast(
+            Optional[DatasetReference],
```
tswast, line 316: I'd actually like to see `_properties` store the REST API version and the property getter call `from_api_repr`. Same for the other properties.

tswast (changes requested):
Sweet! I think the behavior is right now. Just a nit re: what we're storing in `_properties`. I'd like us to stay consistent with the other resource classes.

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,286 @@ def __repr__(self):
...
+    def dataset(self, value):
+        if self.role is not None:
+            raise ValueError(
+                "Role must be None for a dataset. Current " "role: %r" % (self.role)
+            )
+
+        if isinstance(value, str):
+            value = DatasetReference.from_string(value).to_api_repr()
+
+        if isinstance(value, (Dataset, DatasetListItem)):
```
tswast, line 328: Are we missing a case for `DatasetReference`?

```suggestion
        if isinstance(value, DatasetReference):
            value = value.to_api_repr()

        if isinstance(value, (Dataset, DatasetListItem)):
```

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,286 @@ def __repr__(self):
...
+        """Which resources that the dataset in this entry applies to."""
+        return typing.cast(
+            Optional[List[str]],
+            _helpers._get_sub_prop(self._properties, ["dataset", "targetTypes"]),
+        )
+
+    @dataset_target_types.setter
+    def dataset_target_types(self, value):
+        if "dataset" not in self._properties:
+            self._properties["dataset"] = {}
```
tswast, line 349: Alternatively, you could use [`setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault)

```suggestion
        self._properties.setdefault("dataset", {})
```

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,286 @@ def __repr__(self):
...
+        if self.role is not None:
+            raise ValueError(
+                "Role must be None for a routine. Current " "role: %r" % (self.role)
+            )
+
+        if isinstance(value, str):
+            value = RoutineReference.from_string(value).to_api_repr()
+
+        if isinstance(value, Routine):
+            value = value.reference.to_api_repr()
```
tswast, line 369: Likewise, I think we want to handle RoutineReference here too.

```suggestion
        if isinstance(value, RoutineReference):
            value = value.to_api_repr()

        if isinstance(value, Routine):
            value = value.reference.to_api_repr()
```

google/cloud/bigquery/dataset.py
```diff
@@ -382,6 +219,286 @@ def __repr__(self):
...
+        if self.role is not None:
+            raise ValueError(
+                "Role must be None for a view. Current " "role: %r" % (self.role)
+            )
+
+        if isinstance(value, str):
+            value = TableReference.from_string(value).to_api_repr()
+
+        if isinstance(value, Table):
+            value = value.reference.to_api_repr()
```
tswast, line 390:
```suggestion
        if isinstance(value, TableReference):
            value = value.to_api_repr()

        if isinstance(value, Table):
            value = value.reference.to_api_repr()
```

tswast (approved):
Looking really good! I'll approve, but there are a few edge cases you'll want to handle before merging.

PR #1131, opened 2022-02-04
fix: return 403 when VPC-SC violation happens

Erroration2022:
This is fixing the issue where VPC-SC violation is not returning 403. Error message map does not recognize VPCSC policy violation error and will default to return an internal server error.


tswast (approved):

tswast:
Merging blocked by https://github.com/googleapis/google-resumable-media-python/pull/297

PR #1141, opened 2022-02-16
feat: add `--no_query_cache` option to `%%bigquery` magics to disable…

tswast:
… query cache

Fixes internal issue 215520997 🦕

### DIFF
--- a/google/cloud/bigquery/magics/magics.py
+++ b/google/cloud/bigquery/magics/magics.py
@@ -35,6 +35,8 @@
         A dataset and table to store the query results. If table does not exists,
         it will be created. If table already exists, its data will be overwritten.
         Variable should be in a format <dataset_id>.<table_id>.
+    * ``--no_query_cache`` (Optional[line argument]):
+        Do not use cached query results.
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/magics/magics.py
```diff
@@ -442,6 +444,12 @@ def _create_dataset_if_necessary(client, dataset_id):
         "this option's value in the context bqstorage_client_options."
     ),
 )
+@magic_arguments.argument(
+    "--no_query_cache",
+    action="store_true",
```
shollyman, line 449: I'm not understanding what this "store_true" is about, but it seems consistent with the other arguments.

shollyman (approved):

google/cloud/bigquery/magics/magics.py
tswast: That's behavior IPython/Jupyter inherits from argparse https://docs.python.org/3/library/argparse.html#action

The gist is that if this argument is present it'll set `args.no_query_cache = True` rather than requiring a boolean parameter.

PR #1166, opened 2022-03-10
fix(deps): require google-api-core >= 1.31.5, >= 2.3.2 on v1 release

busunkim96:
Allow google-api-core<3.0.0dev on previous release of library. This will be followed by a release PR.


nicain (approved):

tswast (approved):
LGTM. Test failures look like a conflict between new pandas and old pyarrow, but probably not worth investigating further.

PR #1182, opened 2022-03-29
Feat: adding async support for notebooks magics

loferris:


google/cloud/bigquery/magics/magics.py
```diff
@@ -512,7 +522,7 @@ def _cell_magic(line, query):
 
     Args:
         line (str): "%%bigquery" followed by arguments as required
-        query (str): SQL query to run
+        cell_body (str): SQL query or other command to run TODO
```
shollyman, line 525: Can you clarify here?

shollyman (commented):
Unclear on the completeness here, given the TODO comments and test failures.

google/cloud/bigquery/magics/magics.py
loferris: Will update soon to resolve test failures - the TODOs are typos/oversights.

tests/unit/test_magics.py
```diff
@@ -1970,3 +1970,19 @@ def test_bigquery_magic_create_dataset_fails():
...
+    ip = IPython.get_ipython()
+    ip.extension_manager.load_extension("google.cloud.bigquery")
+
+    run_query_patch = mock.patch(
+        "google.cloud.bigquery.magics.magics._run_query", autospec=True
+    )
+
+    with run_query_patch as query_mock:
+        cell_magic_args = "--send_widget_job"
+        ip.run_cell_magic("bigquery", cell_magic_args, "SELECT 1")
```
brijk7, line 1987: It would be nice if we could test a query that takes longer than 10 sec to complete and verify that the widget output shows the 'job still running' and 'job is done' status messages if possible. At the least, the status messages to indicate that the widget really kicked in and ran its code.

tests/unit/test_magics.py
```diff
@@ -33,7 +33,7 @@
 try:
     from google.cloud.bigquery.magics import magics
 except ImportError:
-    magics = None
+    raise  # magics = None
```
shollyman, line 36: I'm assuming this is some temporary debugging thing that leaked into the PR.  The point here is to have a bool you can use for skipif annotations.  You'll want to do the same with ipywidgets, so you only execute the test if the optional dependency is present.

shollyman (commented):
You may want to articulate ipywidgets in the extra optional dependencies in setup.py, the same place where ipython is surfaced.

Most of the test failures currently are dependency related.  probably skipping the tests when the dependencies are insufficient should clear up most of the issues.

google/cloud/bigquery/magics/magics.py
```diff
@@ -89,15 +89,16 @@
...
 
 try:
     import IPython  # type: ignore
     from IPython import display  # type: ignore
     from IPython.core import magic_arguments  # type: ignore
 except ImportError:  # pragma: NO COVER
-    raise ImportError("This module can only be loaded in IPython.")
-
+    raise  # ImportError("This module can only be loaded in IPython.")
+import ipywidgets as widgets
```
shollyman, line 101: Should this be in the try block as well?  Put another way, are there cases where importing ipywidgets has merit without ipython presence?

loferris: Very true - I'm wondering if I should change the exception flag at all, though.

loferris:
> You may want to articulate ipywidgets in the extra optional dependencies in setup.py, the same place where ipython is surfaced.
> 
> Most of the test failures currently are dependency related. probably skipping the tests when the dependencies are insufficient should clear up most of the issues.

It ended up being an issue that resolved with an update the `noxfile.py` at project root. However, I completely agree that nesting "ipywidgets" as a dependency of "IPython" is misleading, so I'm separating it out for readability's sake at least.

tests/unit/test_magics.py
loferris: addressed in latest PR!

tests/unit/test_magics.py
loferris: Added a longer-running query to start!

google/cloud/bigquery/magics/magics.py
```diff
@@ -720,6 +736,39 @@ def _cell_magic(line, query):
...
+        def thread_func(widget_job, out):
+            time_sec = 10.0
+            while widget_job.state != "DONE":
+                job_status = "Job is still running!"
+                out.append_stdout(f"{job_status}")
+                time.sleep(time_sec)
+                widget_job.reload()
+            else:
+                result = widget_job.to_dataframe()
+            out.append_stdout(f"{result}")
```
shollyman, line 761: I'm unfamiliar with the widget abstraction, but this seems odd to me.  the result dataframe is simply printed?  Don't users want to access the result programmatically?

noxfile.py
```diff
@@ -252,7 +254,7 @@ def cover(session):
     test runs (not system test runs), and then erases coverage data.
     """
     session.install("coverage", "pytest-cov")
-    session.run("coverage", "report", "--show-missing", "--fail-under=100")
+    session.run("coverage", "report", "--show-missing", "--fail-under=0")
```
shollyman, line 257: I'm guessing this will be resolved before submit?  This invalidates the point of coverage testing

google/cloud/bigquery/magics/magics.py
```diff
@@ -720,6 +736,39 @@ def _cell_magic(line, query):
     finally:
         close_transports()
 
+    if args.send_widget_job:
```
brijk7, line 739: i'm confused why this async query block is located here. the code already constructs a client and job_config object and calls client.query in the _run_query() method. i'd expect that the send_widget_job boolean be passed as an additional argument there, and the thread code to be inside there.
also, this cell_magic function is to return the results dataframe. not sure how that happens in the threaded case.

parthea:
The failure with `Kokoro Prerelease Dependencies` test only appears with `grpcio==1.49.0rc1` and may be due to https://github.com/grpc/grpc/issues/30640

google/cloud/bigquery/magics/magics.py
loferris: I'm running and tweaking in Jupyterlab!

google/cloud/bigquery/magics/magics.py
loferris: I'm running and tweaking in Jupyterlab! Currently it's supposed to display the widget while the job is running and then at the end print the result to dataframe.

loferris:
> The failure with `Kokoro Prerelease Dependencies` test only appears with `grpcio==1.49.0rc1` and may be due to [grpc/grpc#30640](https://github.com/grpc/grpc/issues/30640)

I asked it to ignore that specific version as a fix for now!

google/cloud/bigquery/magics/magics.py
loferris: You're right - I removed the redundant client instance! As for the dataframe, I'm tweaking the output in Jupyterlab currently and will update with what the current solution looks like for further feedback!

brijk7: for handling the output dataframe, it may be sufficient to populate the `args.destination_var` variable in your thread func (if provided) instead of returning it (since the magic function would return while the job is still running, returning the result is not possible, i'd think).

google/cloud/bigquery/magics/magics.py
```diff
@@ -89,14 +89,23 @@
...
+    from IPython.display import display, HTML, clear_output  # type: ignore
     from IPython.core import magic_arguments  # type: ignore
 except ImportError:  # pragma: NO COVER
     raise ImportError("This module can only be loaded in IPython.")
+try:
+    import ipywidgets as widgets  # type: ignore
+except ImportError:  # pragma: NO COVER
+    raise ImportError(
+        "This module can only be loaded in IPython with ipywidgets installed."
+    )
```
brijk7, line 107: with this, we are effectively saying that the BQ magic can only be used with ipywidgets installed. there could be enterprise users for whom existing notebooks will break because they don't have ipywidgets installed, and they don't have permissions to install it (say they are running in hosted notebooks). even if such users don't care about the ability to run async long running cells, they won't be able to use their notebooks. i think we should try import only if args.send_long_job is true.

google/cloud/bigquery/magics/magics.py
```diff
@@ -593,6 +616,37 @@ def _cell_magic(line, query):
 
     close_transports = functools.partial(_close_transports, client, bqstorage_client)
 
+    if args.send_long_job:
```
brijk7, line 619: i'm still unclear why this whole block of the async job handling is here, before other common processing happens. i'd expect all of this logic to be inside `_run_query()`: 
```
if send_long_job:
  start thread and use widget to output
else:
  use blocking while loop
return query_job # either complete or running
```

loferris: I'm currently investigating how to best update and use `_run_query()` atm!

google/cloud/bigquery/magics/magics.py
```diff
@@ -679,13 +724,18 @@ def _cell_magic(line, query):
...
             _handle_error(ex, args.destination_var)
             return
 
         if not args.verbose:
-            display.clear_output()
+            clear_output()
+
+        if args.send_long_job:
+            if args.destination_var:
+                IPython.get_ipython().push({args.destination_var: result})
```
brijk7, line 737: execution can reach here while job thread is still running in the async case. i think you'd want to refactor this so that there is one copy of the result post-processing code, and it gets called in both scenarios (sync and async) after completion of the job (maybe as a separate function).

tswast:
We're moving the magics code to a new repo. Closing this and tracking support here: https://github.com/googleapis/python-bigquery-magics/issues/17

PR #1199, opened 2022-04-01
feat: add str method to table

### LINKED ISSUE #1174: Add __str__ method to Table class
*TableReference* class already has an implementation of the `__str__` method.

It would be convenient to have the same implementation for the *Table* class.

abecerrilsalas:

Fixes #1174 🦕


tests/unit/test_table.py
```diff
@@ -1410,7 +1412,11 @@ def test___repr__(self):
         )
         self.assertEqual(repr(table1), expected)
 
-
+    def test___str__(self):
+        dataset = DatasetReference("project1", "dataset1")
```
loferris, line 1416: I would follow the convention in the other tests and name this var `dataset_ref`

steffnay (approved):

PR #1217, opened 2022-04-14
fix: export bigquery.HivePartitioningOptions

jli:
(I didn't file an issue for this because it's small) 🦕

My team programmatically creates externally partitioned tables and they hold some of our most core deliverables - they're working great! We followed the sample code here: https://cloud.google.com/bigquery/docs/samples/bigquery-create-table-external-hivepartitioned#bigquery_create_table_external_hivepartitioned-python

But Pylance complains about this line:
```python
hive_partitioning_opts = bigquery.external_config.HivePartitioningOptions()
```
> "external_config" is not a known member of module 

It seems like most of the other `*Options` classes are exported directly, so this does the same for HivePartitioningOptions.


shollyman (approved):
Thanks for catch, and the contribution!

Looks like there's a test failure in our prerelease testing, but it's unrelated to this PR.  I'll go ahead and merge.

PR #1222, opened 2022-04-19
feat: support using BIGQUERY_EMULATOR_HOST environment variable

### LINKED ISSUE #1193: Allow setting of api_endpoint via environment variable like google-cloud-storage

I'd like a way to have a bigquery emulator running as a local docker container which I can use for local integration testing.

I'm able to achieve something similar with GCS using https://github.com/oittaa/gcp-storage-emulator and setting the STORAGE_EMULATOR_HOST environment variable which is read during the storage Client constructor and used to set the api_endpoint in the client_info see [here](https://github.com/googleapis/python-storage/blob/c7bf615909a04f3bab3efb1047a9f4ba659bba19/google/cloud/storage/_helpers.py#L66) and [here](https://github.com/googleapis/python-storage/blob/c7bf615909a04f3bab3efb1047a9f4ba659bba19/google/cloud/storage/client.py#L133).   I've found I can construct a BQ Client using something like `bigquery.Client(client_options={"api_endpoint": "http://localhost:8080"})` and the calls will be redirected to that host/port.  However, that requires changing code which creates clients.  Whereas with the environment variable in storage I can redirect without changing code.  I'd like this library to implement similar logic to check for the existence of a BIGQUERY_EMULATOR_HOST or some similar environment variable and use that to set the api_endpoint in the same way as the STORAGE_EMULATOR_HOST does so that I could then stub out a mock BigQuery emulator to use in testing.

As described above I've looked at changing any constructors in my code base to include the api_endpoint and that is an okay work around, but it seems like having a consistent way to specify an alternate via an environment variable would be better.

 **Additional context**
Hopefully the links above are sufficient extra context.

abecerrilsalas:
## What?
Implemented a BIGQUERY_EMULATOR_HOST environment variable to be used to set the api_endpoint in the python-bigquery client constructor.

## Why?
Users have requested this implementation (#1193). Currently there is no logic to check for the existence of a BIGQUERY_EMULATOR_HOST or some similar environment variable

## How?
Logic was applied and code was changed in the _helpers.py and client.py files.

Fixes #1193 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -231,6 +231,7 @@ class Client(ClientWithProject):
         )
 
         kw_args = {"client_info": client_info}
+        # 
         storage_host = _get_storage_host()
         kw_args["api_endpoint"] = (
             storage_host if storage_host != _DEFAULT_STORAGE_HOST else None
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -66,7 +66,7 @@ from google.cloud.bigquery._helpers import _get_sub_prop
 from google.cloud.bigquery._helpers import _record_field_to_json
 from google.cloud.bigquery._helpers import _str_or_none
 from google.cloud.bigquery._helpers import _verify_job_config_type
-from google.cloud.bigquery._helpers import _get_host
+from google.cloud.bigquery._helpers import _get_bigquery_host
 from google.cloud.bigquery._helpers import _DEFAULT_HOST
 from google.cloud.bigquery._http import Connection
 from google.cloud.bigquery import _pandas_helpers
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/client.py
```diff
@@ -230,6 +231,11 @@ def __init__(
         )
 
         kw_args = {"client_info": client_info}
+        #
+        storage_host = _get_host()
```
steffnay, line 235: Variable name should be changed for bigquery.

steffnay (changes requested):
Looking good! Just needing changes to variable naming and then the additional tests you have planned.

tests/unit/test__helpers.py
```diff
@@ -1288,3 +1288,52 @@ def test_decimal_as_float_api_repr():
...
+        client = self._make_one()
+
+        self.assertEqual(client._connection.API_BASE_URL, _DEFAULT_HOST)
+
+    def test_bigquery_emulator_host(self):
+        client_options = {"api_endpoint": "https://www.foo-googleapis.com"}
+        client = self._make_one(client_options=client_options)
+        self.assertEqual(
+            client._connection.API_BASE_URL, "https://www.foo-googleapis.com"
+        )
```
steffnay, line 1339:
```suggestion
```
I think our coverage is good with the first round of tests you added 👍

steffnay (approved):

PR #1235, opened 2022-04-29
feat: add support for table clones

### LINKED ISSUE #1243: Implement table clones
Feature request for tracking implementation of Table Clones functionality in BigQuery client. Table Clone feature allows customers to create a second copy of a table that is mutable.

Additional information has been provided to implementors separately.

abecerrilsalas:
Implemented the CLONE table copy mechanism  to support creation of  delta storage (clone) tables.

Fixes #1243 🦕


tests/system/test_client.py
```diff
@@ -2190,3 +2190,72 @@ def test_table_snapshots(dataset_id):
...
     assert rows == [(1, "one"), (2, "two")]
+
+
+def test_table_clones(dataset_id):
+    from google.cloud.bigquery import CopyJobConfig
+    from google.cloud.bigquery import OperationType
+
+    client = Config.CLIENT
+
+    source_table_path = f"{client.project}.{dataset_id}.test_table"
```
steffnay, line 2201: This table name needs to be different.

google/cloud/bigquery/table.py
```diff
@@ -1304,6 +1318,29 @@ def __init__(self, resource: Dict[str, Any]):
             )
 
 
+class CloneDefinition:
+    """Information about base table and clone time of the clone.
+
+    See https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#clonedefinition
```
leahecole, line 1324: This is a nit and I love your attention to details, but do we need this link here? Or do we run the risk that it will go obsolete? I defer to @steffnay and/or @shollyman for what's typical with this library

samples/snippets/requirements.txt
```diff
@@ -5,7 +5,7 @@ grpcio==1.46.0
 ipython===7.31.1; python_version == '3.7'
 ipython===8.0.1; python_version == '3.8'
 ipython==8.3.0; python_version >= '3.9'
-matplotlib==3.5.2
+matplotlib==3.5.1
```
leahecole, line 8:
```suggestion
matplotlib==3.5.2
```

Unless 3.5.1 is required for some reason, in which case just add a comment next to it with why it's needed

tests/system/test_client.py
```diff
@@ -2190,3 +2190,57 @@ def test_table_snapshots(dataset_id):
...
+    copy_job.result()
+
+    # List rows from the source table and compare them to rows from the clone.
+    rows_iter = client.list_rows(table_path_source)
+    rows = sorted(row.values() for row in rows_iter)
+    assert rows == [(1, "one"), (2, "two")]
+
+    rows_iter = client.list_rows(clone_table_path)
+    rows = sorted(row.values() for row in rows_iter)
+    assert rows == [(1, "one"), (2, "two")]
```
leahecole, line 2237: It might just be my lack of familiarity for this feature and the fact that I'm trying to do this on a train but should we be checking for the "foo" and "bar" also? (or is this feature supposed to be getting rid of them?)

tests/unit/test_table.py
```diff
@@ -841,6 +841,40 @@ def test_snapshot_definition_set(self):
             2010, 9, 28, 10, 20, 30, 123000, tzinfo=UTC
         )
 
+    def test_clone_definition_not_set(self):
```
leahecole, line 844: are there other conditions we should be checking for where it isn't set? like some kind of error? (or is that more in the unit tests? defer to BQ experts here)

tests/unit/test_table.py
```diff
@@ -1789,6 +1823,46 @@ def test_ctor_full_resource(self):
...
+    def _make_one(cls, *args, **kwargs):
+        klass = cls._get_target_class()
+        return klass(*args, **kwargs)
+
+    def test_ctor_empty_resource(self):
+        instance = self._make_one(resource={})
+        assert instance.base_table_reference is None
+        assert instance.clone_time is None
+
+    def test_ctor_full_resource(self):
```
leahecole, line 1843: are `klass` and `ctor` known BQ terms? If not and if they're abbreviations or acronyms, please lengthen them to be a more human readable. If known BQ or similar terms, plz disregard

tests/unit/test_table.py
steffnay: I think this is good and follows the pattern we used with snapshot implementation.

tests/unit/test_table.py
steffnay: These terms are used throughout the tests in this library, but I agree we could make them more human readable. Perhaps I can open a bug for doing that across all instances.

tests/system/test_client.py
steffnay: The "foo" and "bar" are the column names and there's a check to ensure the schema is correct on line 2243, so any discrepancy should be caught.

google/cloud/bigquery/table.py
steffnay: We have similar links to the docs in some other properties, including the `Table.snapshot` property, so I think it's okay to leave this as well.

steffnay (approved):

PR #1245, opened 2022-05-09
fix(deps): allow pyarrow v8

parthea:


leahecole (approved):

meredithslota:
@bnaul The reason that we use this approach is because a new major version of a dependency can break users who do not explicitly restrict them via `requirements.txt`. We know it is best practice to have a `requirements.txt` file but not all users do that. In addition, it may take time for our code or samples to be re-written/fixed in order to have compatibility with a new major version, and we'd prefer to err on the side of not accidentally breaking folks. We try to update as soon as possible. Thanks!

PR #1271, opened 2022-06-07
fix(deps): require protobuf<4.0.0 on v1 branch

parthea:


noxfile.py
```diff
@@ -206,7 +206,7 @@ def blacken(session):
 def docs(session):
     """Build the docs."""
 
-    session.install("ipython", "recommonmark", "sphinx<3.0.0", "sphinx_rtd_theme")
+    session.install("ipython", "recommonmark", "sphinx<3.0.0", "jinja2<3.1", "sphinx_rtd_theme")
```
Mariatta, line 209: This is a really old version of Sphinx. I would recommend updating it, but doesn't need to be in this PR.

Mariatta (approved):

PR #1309, opened 2022-08-11
docs(samples): uses function (create_job) more appropriate to the described sample intent

### LINKED ISSUE #1085: Python `bigquery_create_job` sample is inconsistent with other code samples
The https://github.com/googleapis/python-bigquery/blob/main/samples/create_job.py code sample appears here: https://cloud.google.com/bigquery/docs/running-jobs#running_jobs_using_client_libraries

This sample requires two improvements:

* [ ] That section is about any kind of job, not just queries. As such, it should use the [create_job](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.create_job) method instead of the more specific `query` method. There should be comments that it is recommended to use the corresponding method for query/copy/load/extract.
* [ ] Remove `job_id_prefix` logic, as labels cover similar functionality and are recommended over custom job IDs.
* [ ] Move the sample to the `samples/snippets` directory. Needs to be done in 2 separate PRs so that the existing samples in the docs don't 404.

chalmerlowe:
This is a work in progress.

The previous version simply used the `.query()` method, but the description and intent was to display the use of the `create_job()` method. This migrates the code to using the intended method.

Fixes #1085 🦕

BEGIN_COMMIT_OVERRIDE
docs(samples): uses function (create_job) more appropriate to the described sample intent 
END_COMMIT_OVERRIDE


noxfile.py
```diff
@@ -288,7 +288,7 @@ def prerelease_deps(session):
         "google-cloud-bigquery-storage",
         "google-cloud-core",
         "google-resumable-media",
-        "grpcio",
+        "grpcio!=1.49.0rc1",
```
dandhlee, line 291: If it can't be higher than 1.49.0rc1, wouldn't it be better to note it with `grpcio<=1.49.0rc1`?

samples/create_job.py
```diff
@@ -13,33 +13,46 @@
...
 import typing
+from typing import Union
 
 if typing.TYPE_CHECKING:
-    from google.cloud import bigquery
+    from google.cloud.bigquery import LoadJob, CopyJob, ExtractJob, QueryJob
 
 
-def create_job() -> "bigquery.QueryJob":
+def create_job() -> "Union[LoadJob, CopyJob, ExtractJob, QueryJob]":
```
dandhlee, line 22: This is definitely much better than `bigquery.QueryJob`!

samples/create_job.py
```diff
@@ -13,33 +13,46 @@
...
-        ),
-        # The client libraries automatically generate a job ID. Override the
-        # generated ID with either the job_id_prefix or job_id parameters.
-        job_id_prefix="code_sample_",
+    query_job = client.create_job(
+        # Specify a job configuration, providing both query
+        # and to set optional job resource properties, if needed.
+        # The job instance can be a LoadJob, CopyJob, ExtractJob, QueryJob
+        # Here, we demonstrate a "query" job.
+        # Reference: https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.create_job
```
dandhlee, line 35: This is not going to render well with the long line of text on Sample Browser or even seeing this on GitHub. I like the link to the documentation though, could you perhaps add this link on https://cloud.google.com/bigquery/docs/samples/bigquery-create-job page instead?

samples/create_job.py
```diff
@@ -13,33 +13,46 @@
...
+        #
+        # NOTE: the preferred approach is to use one of the dedicated API calls:
```
dandhlee, line 37: Could you elaborate on what "preferred approach" should be for?

If it's for executing the query it seems counterproductive to have a sample for `create_client` if this is not the preferred method.

samples/create_job.py
```diff
@@ -13,33 +13,46 @@
...
+                         FROM `bigquery-public-data.utility_us.country_code_iso`
+                         LIMIT 5
+                         """,
+            },
+            "labels": {"example-label": "example-value"},
+            "maximum_bytes_billed": 10000000,
+        }
     )  # Make an API request.
 
     print("Started job: {}".format(query_job.job_id))
```
dandhlee, line 55:
```suggestion
    print(f"Started job: {query_job.job_id}")
```

noxfile.py
chalmerlowe: It is allowed to be higher than `1.49` (presuming future releases fix the problem.)
For now, across a number of googleapis python packages, it appears that this selective approach of avoiding just this release candidate is the approach that is being used. In fact, it appears that this approach has already been applied to this repo (and this one line change will be rendered moot).
https://github.com/googleapis/python-bigquery/pull/1328/files

samples/create_job.py
chalmerlowe: I think I am confused: your link takes us to the page where this sample is displayed. 
Is that intentional? That page does not currently provide additional information regarding the four types of jobs available. 

Relatedly: in our renderings on the Sample Browser, is there a way to create standard hyperlinks within the code blocks displayed in the screen?

samples/create_job.py
chalmerlowe: I love `f-strings`. Thanks for suggesting it.

samples/create_job.py
chalmerlowe: @tswast can you elaborate for us on this comment you made in the issue, as it relates to @dandhlee's question above?

> "That section is about any kind of job, not just queries. As such, it should use the [create_job](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client.create_job) method instead of the more specific query method. There should be comments that it is recommended to use the corresponding method for query/copy/load/extract."

chalmerlowe (commented):
See specific comments and questions at various places in the code.

noxfile.py
dandhlee: I think it should be good to revert this then.

samples/create_job.py
dandhlee: It currently does not, but once your PR updates it the page will also be updated.

Tried sifting through https://googlecloudplatform.github.io/samples-style-guide/ and we currently don't provide any guidance on how hyperlinks can be should they be too long... Let me get back to you on this after I ask some folks around.

chalmerlowe: I looked at the style guide a bit. thanks for linking to it. good reminder for me that it exists.

This https://googlecloudplatform.github.io/samples-style-guide/#clients item has a Python snippet with a really long URL in the code sample, just like the one I included.
¯\_(ツ)_/¯

samples/create_job.py
tswast: One use-case this sample would be preferred to compared to one of the more specific examples is when folks try to retry failed jobs.

There are customers who use this method by iterating through a list of recent jobs and retrying any that have failed as a way to make their data pipelines a bit more robust.

One could also use this method as a way to create a job with an experimental API property that hasn't been added to the client library's manually written job configuration classes yet.

samples/create_job.py
tswast: We could remove the ending anchor (`#google.cloud...`) to make the URL a bit smaller.

I recall that we have a g.co/cloud URL shortener for cloud.google.com pages (e.g. cloud.google.com/bigquery becomes g.co/cloud/bigquery, which isn't all that much shorter but we occasionally used it) I wonder if we could get g.co/bqpython or something similar pointing to the latest API reference?

chalmerlowe: I submitted a request for a short link:

g.co/bqpython > https://googleapis.dev/python/bigquery/latest/

It has to go through an approval process. 

I would suggest that we not hold off on issuing this PR. Especially in light of the fact that even the style guide in code samples has examples of extremely long URLs, as noted in the comment above.

chalmerlowe:
@tswast  @dandhlee 

I added some justification for when/why `.create_jobs()` might be preferred over a direct call to `.query()`, etc.
I have also made a request (that needs to be approved) for a shorten URL, but I don't think that should hold up the processing of this PR.

Can you take a look at the changes I made and see if we are closer to the mark. Thanks.

samples/create_job.py
```diff
@@ -13,33 +13,53 @@
...
+        #      been added to one of the manually written job configuration
+        #      classes yet
+        #
+        # NOTE: unless it is necessary to create a job in this way, the
+        # preferred approach is to use one of the dedicated API calls:
+        #   client.query()
+        #   client.extract_table()
+        #   client.copy_table()
+        #   client.load_table_file(), client.load_table_from_dataframe(), etc
+        job_config={
```
tswast, line 49: I think a link to https://cloud.google.com/bigquery/docs/reference/rest/v2/Job would be quite helpful here.

tswast (approved):

samples/create_job.py
dandhlee: Thanks for looking into the shortlink process! I wasn't aware of such features. Hope it works out :)

I've asked the samples team for guidance, however it likely will take a long time for us to come up with a feasible solution, and will likely involve multiple teams. For now, it's adding more benefits so I'm happy to move forward as is.

dandhlee (approved):
LGTM. Please see Tim's comment!

samples/create_job.py
chalmerlowe: I added the link.

PR #1395, opened 2022-10-31
fix: corrects test for non-existent attribute

chalmerlowe:
A test was failing because we were attempting to access an attribute that does not exist (NOTE it is not supposed to exist for this test).

Our test was supposed to detect that the failing attribute did not exist by using an `assertIsNone()` test method, when we should have been asserting that the attribute does not exist, because within the `assertIsNone()` method, if the attribute does not exist when you call it, it raises an `AttributeError` instead of returning `None`.

### DIFF
--- a/tests/unit/test_table.py
+++ b/tests/unit/test_table.py
@@ -1971,9 +1971,9 @@ class Test_EmptyRowIterator(unittest.TestCase):
         self.assertIsInstance(df, geopandas.GeoDataFrame)
         self.assertEqual(len(df), 0)  # verify the number of rows
         if version_info.major == 3 and version_info.minor > 7:
-            assert not hasattr(df, "crs")  # used with Python == 3.7
+            assert not hasattr(df, "crs")  # used with Python > 3.7
         else:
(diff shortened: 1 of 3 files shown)


tests/unit/test_table.py
```diff
@@ -1969,7 +1970,10 @@ def test_to_geodataframe(self):
         df = row_iterator.to_geodataframe(create_bqstorage_client=False)
         self.assertIsInstance(df, geopandas.GeoDataFrame)
         self.assertEqual(len(df), 0)  # verify the number of rows
-        self.assertIsNone(df.crs)
+        if version_info.major == 3 and version_info.minor > 7:
+            assert not hasattr(df, "crs")  # used with Python == 3.7
```
tswast, line 1974: Are these comments swapped?

tswast (approved):

tests/unit/test_table.py
chalmerlowe: Looks like it.
Sigh.

Will go fix.

parthea (approved):

PR #1420, opened 2022-11-22
docs: adds snippet for creating table with external data config

### LINKED ISSUE #773: Document how to construct external tables

Currently, the document on creating a table is clear on how to initialize a Table object with a specific fields schema. For example, here is a code snippet from the docs:

image

However, a challenge I faced with these docs is knowing how to create a Table with, e.g., an ExternalConfig set.

Based on the source for Table / unit tests, is the correct approach using a property setter?:

```python
client bigquery.Client()

table = bigquery.Table(...)

# use property setter
table.external_data_configuration = bigquery.ExternalConfig(...)

# edit: it looks like ExternalConfig also needs to use setters, so might include, e.g..
external_config = bigquery.ExternalConfig("NEWLINE_DELIMITED_JSON")
external_config.source_uris = [...]
```

Can we add an example to the docs so that readers learn these two pieces?:

* Table attributes like schema can be set on `__init__` (or via setters)
* Other Table attributes must be set using setters

It seems like two places users may look for this info is in the Table API doc, and in the [usage/tables - creating a table](https://googleapis.dev/python/bigquery/latest/usage/tables.html#creating-a-table) doc. Also, apologies if it's already somewhere in the docs and I missed it 😅.

Since Table can be constructed using `.from_api_repr()`, I wondered if this is a more common approach for construction? In that case it seems like having a nudge in the docs on using this path for e.g. constructing an external table would be super helpful!

aribray:
Adds snippets and tests for `create_table_external_data_configuration` and `create_external_table_definition`

Fixes #773  🦕


samples/create_table_external_data_configuration.py
```diff
@@ -0,0 +1,66 @@
...
+    # [START bigquery_create_external_table_definition]
+    from google.cloud import bigquery
+
+    # Construct a BigQuery client object.
+    client = bigquery.Client()
+
+    """
+    TODO(developer):
+    Set table_id to the ID of the table to create.
+    table_id = "your-project.your_dataset.your_table_name"
```
tswast, line 35: I've been using a different pattern for this. I got the occasional complaint from folks who didn't like the fact that necessary code was included in comments.

Now I reset variables like Mario does here: https://github.com/googleapis/python-bigquery/pull/1436/files#diff-933d0a59ab13b3ed5238456246f4e3fd3566082b4c2d0733949f7f885c358df8R29

chalmerlowe: I agree with @tswast on this one.

As I have grown more familiar with our code samples and based on my experiences with students...
Having code in comments is harder to parse.
Folks tend to glaze over the comments entirely / not read them and just enter the code portions and then get perplexed when the code doesn't work at all. 

Feeding in "working code" using placeholders for things the folks need to replace at least causes the system to respond with an error code that I feel guides them more clearly to what they need to replace.

chalmerlowe (approved):
LGTM, with the exception of the comment related to producing more user friendly code (ie. removing important code from comment strings and putting it into the body of the sample with placeholders).

tswast:
Once internal CL 505774967 is merged, I think we'll want to revert it. Turns out we already had such a code sample, it was just hard to find because the description didn't contain the word "external". https://cloud.google.com/bigquery/docs/samples/bigquery-query-external-gcs-perm

Issue #752, opened 2021-07-12
Unify the way how unit tests are conditionally skipped


plamut:
Following the [comment](https://github.com/googleapis/python-bigquery/pull/750#discussion_r667878253), we currently use two different ways of conditionally skipping unit tests that require particular minimum dependency version.

One approach parses the installed version manually:
```py
PANDAS_MINIUM_VERSION = pkg_resources.parse_version("1.0.0")

if pandas is not None:
    PANDAS_INSTALLED_VERSION = pkg_resources.get_distribution("pandas").parsed_version
else:
    # Set to less than MIN version.
    PANDAS_INSTALLED_VERSION = pkg_resources.parse_version("0.0.0")

...

@unittest.skipIf( 
    pandas is None or PANDAS_INSTALLED_VERSION < PANDAS_MINIUM_VERSION, 
    "Only `pandas version >=1.0.0` supported", 
) 
```

The alternative approach is using `pytest.importorskip`, which is more concise:
```py
pytest.importorskip(
        "pandas",
        minversion="1.0.0",
        reason="Requires `pandas version >= 1.0.0` which introduces pandas.NA",
)
```

We should use `pytest.importorskip` combined with the `minversion` argument. That feature relies on the imported module's `__version__` attribute, but all our major dependencies support it, so we should be fine.

plamut:
Hmm, this issue might already be obsolete, since we are promoting some of the extras to required dependencies in #757. After that is done, there might not be much left to unify in the first place. :slightly_smiling_face:

tswast:
:-)

Pandas is still gonna be optional. But yeah, `importorskip` is so much nicer.

Linchin:
@tswast Just went through the code base, and I don't see any `@unittest.skipIf(...)` with a version check in it. I do see a lot of `@unittest.skipIf(pandas is None, "Requires `pandas`")` though. Can we mark this issue as resolved, or we should replace all `@unittest.skipIf(...)` with `importorskip` here?

tswast:
> Can we mark this issue as resolved, or we should replace all @unittest.skipIf(...) with importorskip here?

Using `importorskip` everywhere would be my preference.

shollyman:
Moving this to @chalmerlowe as he's been cleaning this up, but I'm not sure if its fully completed.

chalmerlowe:
There are no more references to `@unittest.skipif`.
We have solidified on an approach that uses `pytest.importorskip()`

[...]

PR #1473, opened 2023-01-18
testing: simplify optional dependency testing

shollyman:
This library leverages multiple, independent optional dependencies.  Previously we used an indirect way of conditionally importing things and registering global imports per file.  Then, tests would use `@unittest.skipIf` conditionals to check the state of those globals to govern whether a test was skipped or run.

This PR changes how we exercise optional dependencies  in test files.  In this PR, we move to the `pytest.importorskip` to skip tests where unsatisfied optional dependencies exist.  If a test references an import directly, we assign it to the import name.  importorskip can also be supplied with minimum version to test dependencies, but it doesn't appear to be something we've used much so this PR doesn't enforce min-versions in current tests.

## examples

To demonstrate the change, we often directly reference pandas types in a test.  Thus, the invocation looks like this:

```
    def test_something(self):
       pandas=pytest.importorskip("pandas")
      ...
```

However, for something that relies on the BigQuery Storage API has a more indirect namespace, and this we can just use importorskip without assignment.  Because of the optional dependency, we include the import inside the test after the importorskip:

```
    def test_something_else(self):
       pytest.importorskip("google-cloud-bigquery-storage")
      ...
```

Towards: https://github.com/googleapis/python-bigquery/issues/752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@
 [1]: https://pypi.org/project/google-cloud-bigquery/#history
 
 
+## [3.5.0](https://github.com/googleapis/python-bigquery/compare/v3.4.2...v3.5.0) (2023-01-31)
+
+
+### Features
+
+* Add __str__ method to DatasetReference ([#1477](https://github.com/googleapis/python-bigquery/issues/1477)) ([f32df1f](https://github.com/googleapis/python-bigquery/commit/f32df1fb74e4aea24cd8a4099040ad2f7436e54d))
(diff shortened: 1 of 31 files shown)


tswast (approved):
Much cleaner. Thanks!

tswast:
Regarding the pyarrow failures in the kokoro session, we actually have some tests that intentionally install an older incompatible version of pyarrow. We might need to add min version to some `importorskips`.

chalmerlowe:
Moving this content to a new PR that is based off the most recent version of main. Trying to deal with a year's worth of merge conflict made my head hurt.

shollyman:
sounds good.  closing this one.

PR #1487, opened 2023-02-07
fix: annotate optional integer parameters with optional type

### LINKED ISSUE #509: increase minimum version of google-cloud-core to 1.6.0
Once enough time has passed to give people time to upgrade (TBD how long that is), we should increase the minimum version and clean up and logic that switches based on package version that was added in https://github.com/googleapis/python-bigquery/pull/492.

tswast:
I noticed in a project that uses this client that the type annotations for `result` are incorrect.

```
bigframes/dataframe.py:64: error: Argument "max_results" to "result" of "QueryJob" has incompatible type "Optional[int]"; expected "int"  [arg-type]
```

Fixes https://github.com/googleapis/python-bigquery/issues/509

### DIFF
--- a/google/cloud/bigquery/_http.py
+++ b/google/cloud/bigquery/_http.py
@@ -15,9 +15,7 @@
 """Create / interact with Google BigQuery connections."""
 
 import os
-import packaging.version
 
-import google.cloud.version
 from google.cloud import _http  # type: ignore  # pytype: disable=import-error
 from google.cloud.bigquery import __version__
 
--- a/google/cloud/bigquery/_http.py
+++ b/google/cloud/bigquery/_http.py
@@ -15,19 +15,11 @@
 """Create / interact with Google BigQuery connections."""
 
 import os
-import pkg_resources
+import packaging.version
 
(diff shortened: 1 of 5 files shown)


chalmerlowe:
General question:

what was the criteria for deciding which parameters should get the enhanced Type Annotation?

Noticed some other `int` parameters in various methods that did not get the same enhancement and was curious.

chalmerlowe (approved):
Added a general question to the comment thread.
Beyond that question, this LGTM.

tswast:
> what was the criteria for deciding which parameters should get the enhanced Type Annotation?
> Noticed some other int parameters in various methods that did not get the same enhancement and was curious.

Our approach was: when you touch a method, add type annotations.

At the time, that meant we could also remove the type information from the docstring because it was then redundant information from what Sphinx generated and what we put in the docstring. Looks to be the same with https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client but I'm not 100% sure how the linking works over there since they don't use intersphinx.

PR #1497, opened 2023-02-11
feat: Add checksum option that defaults to md5 

### LINKED ISSUE #1496: Add checksumming for file uploads
Any time network transfer occurs there is a possibility of bit corruption along the way.  Using checksumming to detect corruption can help avoid issues downstream.  We should checksum final uploaded file contents to help avoid corruption.

emkornfield:
See: https://github.com/googleapis/python-bigquery/issues/1496

Add default checksum for crc32

Fixes #1496


emkornfield:
hmm, looks like a bad merge will try to fix in a little bit.

emkornfield:
Looking into what is going wrong with samples.

google/cloud/bigquery/client.py
```diff
@@ -2838,6 +2871,7 @@ def _do_resumable_upload(
         num_retries: int,
         timeout: Optional[ResumableTimeoutType],
         project: Optional[str] = None,
+        checksum: Optional[str] = None,
```
andrewsg, line 2874: The intention here is to have a default for the public method, but not the private methods, correct?

andrewsg (approved):

google/cloud/bigquery/client.py
emkornfield: yes. that is what I was thinking.

I looked into this and this will require probably some changes to the google media library and likely the backend.

chalmerlowe:
@emkornfield 
Checking in to see the status on this. Looks like it has gone a little stale.
Anything that you need to help get this over the finish line?

emkornfield:
Going to close this until I have more bandwidth to work on it.

PR #1521, opened 2023-03-10
feat: expose configuration property on CopyJob, ExtractJob, LoadJob, QueryJob

tswast:
Note for google-cloud-bigquery developers: This also refactors these classes so that `_set_properties` does not modify the `_properties` dictionary in-place. Doing so was also mutating the request object, making it difficult to debug what request was _actually_ sent. Before this change, many tests hallucinated that the request was always equal to the response.

🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -1976,15 +1976,8 @@ class Client(ClientWithProject):
             )
             destination = _get_sub_prop(job_config, ["copy", "destinationTable"])
             destination = TableReference.from_api_repr(destination)
-            sources = []
-            source_configs = _get_sub_prop(job_config, ["copy", "sourceTables"])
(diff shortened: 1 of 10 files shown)


google/cloud/bigquery/client.py
```diff
@@ -1976,15 +1976,8 @@ def create_job(
             )
             destination = _get_sub_prop(job_config, ["copy", "destinationTable"])
             destination = TableReference.from_api_repr(destination)
-            sources = []
```
tswast, line 1979: There was a unit test failing because we were actually transforming `"sourceTable"` to `"sourceTables"` without this change.

tswast:
P.S. Thanks @chelsea-lin for identifying this bug!

rosiezou (approved):
LGTM

Issue #1345, opened 2022-09-08
Optional switch to non-nullable dtypes in to_dataframe


derHeinzer:
Release 3.0.0 introduced the use of nullable Int64 and boolean dtypes (pandas extension dtypes) (#786)

However the pandas extension dtypes are not widely supported across the pandas api. This might lead to issues using numba in pandas operations.

Please see my bug report in the pandas project for example:
https://github.com/pandas-dev/pandas/issues/46867

I would propose to introduce an optional switch to pre-3.0.0 behaviour.

derHeinzer:
To make things worse, not all widely-used machine learning libraries can handle pandas extension dtypes. I just figured out, the training of a lightgbm Regressor ist not possible with such input. 
I did not check every library I know, but I am pretty sure there must be others.

tswast:
Thank you so much for the feedback! I hope that the nullable dtypes can stay as the default, as they more accurately reflect the BigQuery data types, but you make some good points that there are still use cases where the old behavior would be desirable.

tswast:
Related: https://github.com/googleapis/python-bigquery/issues/1345

I'm thinking we add `int_dtype`, `bool_dtype`, `string_dtype`, `float_dtype`, `timestamp_dtype`, `datetime_dtype` as arguments to `to_dataframe`.

Alternative 1: expose typemapper directly, but doubtful that a lambda function that inspects an arrow schema would be all that understandable to the average pandas user.

Alternative 2: expose `use_nullable_dtypes` argument to match https://pandas.pydata.org/docs/reference/api/pandas.read_parquet.html and similar methods. It defaults to True to avoid lossy conversions as we have attempted to do in google-cloud-bigquery v3 -- technically another breaking change though since we'll want to change to use Float64Dtype and StringDtype too, which we aren't currently.

chelsea-lin:
The changes have been merged.

PR #1529, opened 2023-03-22
feat: add bool, int, float, string dtype to to_dataframe

### LINKED ISSUE #954: Support for string[pyarrow] dtype
[Pandas 1.3](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.3.0.html#pyarrow-backed-string-data-type) added a new `string[pyarrow]` dtype which can be considerably more [memory-efficient](https://www.youtube.com/watch?v=_zoPmQ6J1aE).

I'm not sure what all would be involved but obviously it would be nice to support this natively since presumably(?) we already communicate the data in the appropriate format for the pyarrow string type before converting it back to python string objects. Maybe an option like was introduced in #848 for geography types could be used to determine the behavior?

chelsea-lin:
feat: add 'bool_dtype', 'int_dtype', 'float_dtype' and 'string_dtype' to the 'to_dataframe' APIs

This change introduces  'bool_dtype', 'int_dtype', 'float_dtype' and 'string_dtype' parameters to the 'to_dataframe' APIs, so that users can choose the custom ExtensionDtype for data conversions. 

Fixes #954 and #1345 🦕

### DIFF
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -21,6 +21,7 @@ from itertools import islice
 import logging
 import queue
 import warnings
+from typing import Any, Union
 
 from packaging import version
 
@@ -283,7 +284,13 @@ def bq_to_arrow_schema(bq_schema):
     return pyarrow.schema(arrow_fields)
 
 
-def default_types_mapper(date_as_object: bool = False):
+def default_types_mapper(
+    date_as_object: bool = False,
+    bool_dtype: Union[Any, None] = None,
+    int_dtype: Union[Any, None] = None,
+    float_dtype: Union[Any, None] = None,
+    string_dtype: Union[Any, None] = None,
+):
     """Create a mapping from pyarrow types to pandas types.
 
     This overrides the pandas defaults to use null-safe extension types where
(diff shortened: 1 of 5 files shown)


google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -283,7 +284,13 @@ def bq_to_arrow_schema(bq_schema):
     return pyarrow.schema(arrow_fields)
 
 
-def default_types_mapper(date_as_object: bool = False):
+def default_types_mapper(
+    date_as_object: bool = False,
+    bool_dtype: Union[Any, None] = pandas.BooleanDtype(),
```
tswast, line 289: Pandas is an optional dependency. Pretty sure we can't do this. Let's use a "sentinel" value as in https://stackoverflow.com/a/60605919/101923 to avoid initializing pandas objects at import time.

google/cloud/bigquery/job/query.py
```diff
@@ -1620,6 +1625,10 @@ def to_dataframe(
         create_bqstorage_client: bool = True,
         max_results: Optional[int] = None,
         geography_as_object: bool = False,
+        bool_dtype: Union[Any, None] = pandas.BooleanDtype(),
```
tswast, line 1628: Likewise, we can't initialize pandas objects at import time since pandas is an optional dependency.

tswast (commented):
Love it! Just some nits regarding how we handle the default values.

google/cloud/bigquery/_pandas_helpers.py
chelsea-lin: Good points. Fixed with your suggestions.

google/cloud/bigquery/job/query.py
chelsea-lin: Addressed. Thanks!

tswast (approved):
Love it!

I don't think the prerelease failures could be caused by this change. Filed https://github.com/googleapis/python-bigquery/issues/1530 to investigate.

PR #1551, opened 2023-04-14
bug: handle case when expirationMs is None

abdelmegahedgoogle:
Right now, unsetting ` ['timePartitioning']['expirationMs'] ` on BigQuery table is not supported and raises `google.api_core.exceptions.BadRequest: 400 PATCH https://bigquery.googleapis.com/bigquery/v2/projects/abdelrahmanm-bd-dev/datasets/partition_expiry_test/tables/sample_data_partitioned?prettyPrint=false: Invalid value at 'table.time_partitioning.expiration_ms.value' (TYPE_INT64), "None"`

This happens since the backend doesn't treat "None" (string) as a null. This PR addresses this issue from the client side.


abdelmegahed:
Sample code that fails with the above exception:
```
client = bigquery.Client(project='abdelrahmanm-bd-dev')
table_ref = client.dataset('partition_expiry_test').table('sample_data_partitioned')
table = client.get_table(table_ref)
table.partition_expiration = None # ISSUE HERE!! This get cast to string and the backend returns a 400 as a result.
client.update_table(table, ['partition_expiration'])
```

abdelmegahed:
Re-opened this PR with an account that's approved for CLA. https://github.com/googleapis/python-bigquery/pull/1553

PR #1582, opened 2023-06-08
docs: add/reformat return types for cloud RAD docs

### LINKED ISSUE #1438: Missing return types in some methods on cloud.google.com documentation
Note that `schema_from_json` returns a list of SchemaField objects, but there's no indication of that here: https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_schema_to_json

leahecole:
- In order for the Return table to show up in the docs, the type of the return object must be in the Returns docstring
- I also removed two examples from docstrings - they don't format properly in our docs and I felt they were encompassed in other samples. If we want to keep them we can, I'll just need to reach out to the Cloud RAD team to understand how they need to be formatted

Fixes #1438 once the Cloud RAD job runs after this is merged 😄

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -1934,7 +1934,7 @@ class Client(ClientWithProject):
 
         Returns:
             Union[job.CopyJob, job.ExtractJob, job.LoadJob, job.QueryJob, job.UnknownJob]:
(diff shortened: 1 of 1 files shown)


tswast (approved):
Thanks!

PR #1587, opened 2023-06-19
fix: add missing handler for deserializing json value

### LINKED ISSUE #1500: Deserializing JSON subfields within structs fails
#### Environment details

  - OS type and version: MacOS Ventura 13.1
  - Python version: 3.9.5
  - pip version: 22.3.1
  - `google-cloud-bigquery` version: 3.5.0

#### Steps to reproduce

1. Create table with record field containing a JSON subfield
2. Query record field and try to read row from query result

#### Code example

```python
bigquery.Client().query("SELECT record_with_json_subfield.* FROM <table>").result().__next__()
```

#### Stack trace
```
  File "<stdin>", line 1, in <listcomp>
  File "lib/python3.9/site-packages/google/api_core/page_iterator.py", line 209, in _items_iter
    for item in page:
  File "lib/python3.9/site-packages/google/api_core/page_iterator.py", line 131, in __next__
    result = self._item_to_value(self._parent, item)
  File "lib/python3.9/site-packages/google/cloud/bigquery/table.py", line 2710, in _item_to_row
    _helpers._row_tuple_from_json(resource, iterator.schema),
  File "lib/python3.9/site-packages/google/cloud/bigquery/_helpers.py", line 435, in _row_tuple_from_json
    row_data.append(_field_from_json(cell["v"], field))
  File "lib/python3.9/site-packages/google/cloud/bigquery/_helpers.py", line 411, in _field_from_json
    return converter(resource, field)
  File "lib/python3.9/site-packages/google/cloud/bigquery/_helpers.py", line 368, in _record_from_json
    converter = _CELLDATA_FROM_JSON[subfield.field_type]
KeyError: 'JSON'
```

tomwojcik:
Fixes #1500


tomwojcik:
@tswast bump

tswast (commented):
Thanks!

One concern I have is the other direction. If someone supplies a parsed JSON object to an API like `insert_rows`, do we correctly serialize that?

tswast (approved):

tswast:
Looks like everything is working well except for the type annotations:

```
nox > Running session mypy
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/mypy
nox > python -m pip install -e '.[all]'
nox > python -m pip install mypy==1.6.1
nox > python -m pip install types-protobuf types-python-dateutil types-requests types-setuptools
nox > mypy -p google --show-traceback
google/cloud/bigquery/query.py:472: error: Cannot call function of unknown type  [operator]
google/cloud/bigquery/query.py:629: error: Cannot call function of unknown type  [operator]
google/cloud/bigquery/query.py:778: error: Cannot call function of unknown type  [operator]
Found 3 errors in 1 file (checked 54 source files)
nox > Command mypy -p google --show-traceback failed with exit code 1
nox > Session mypy failed.
```

MartijnvanElferen:
This change is causing the following error for `insert_rows`:

> TypeError: _json_from_json() missing 1 required positional argument: 'field'

Is there a recommended fix for this?

tswast:
Thanks @MartijnvanElferen for the report. I've filed https://github.com/googleapis/python-bigquery/issues/1756 to investigate further.

tswast:
As a workaround [`insert_rows_json`](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_insert_rows_json) should avoid this type conversion code, but you're responsible for putting things in a format the BigQuery REST API understands.

tswast:
@MartijnvanElferen Fix pending: https://github.com/googleapis/python-bigquery/pull/1757

Note: this PR will change the behavior of JSON data in `insert_rows` to call `json.dumps`, which is consistent with the new behavior of calling `json.loads` when reading row values. Previously, JSON support was happening via a generic fallback that didn't do any conversion of values passed in.

MartijnvanElferen:
Thanks, @tswast. I'll refactor a couple of things to avoid the type conversion. Looking forward to the fix at the same time!

kcooney:
@tswast This change should probably be listed as a potentially braking change in the release notes, since before this change reading the entire column would return a value of type `str`:

```sql
CREATE TABLE mydataset.table(
  id INT64,
  proto_as_json JSON
);
```

```python
from google.protobuf import json_format
from com.example.my_proto_pb2 import MyProto

client = bigquery.Client()
proto = MyProto(first_name="Jane", last_name="Doe") 
proto_as_json = json_format.MessageToJson(
    proto, use_integers_for_enums=True, indent=None)
rows_to_insert = [
    {"id": 1, "proto_as_json": proto_as_json},
]
errors = client.insert_rows("mydataset.table", rows_to_insert)
assert not errors

row = client.query("SELECT proto_as_json FROM mydataset.table").result().__next__()
json_str = row["proto_as_json]
# Prior to this change, json_str was of type str, so we could do this:
read_proto = MyProto()
json_format.Parse(json_str, read_proto)
assert proto == read_proto
```

Luckily, in our project we created a common helper method for the `json_format.Parse()` call, so we can update it to use `json_format.ParseDict()`

tswast:
Yeah, it's a gray area with this since we never said we support the JSON type until this change.

We still don't support JSON in some code paths like to_dataframe()

kcooney:
@tswast We saw the breakage when we tried to update a number of python libraries to newer versions. The release notes were very unhelpful in determining which library and version could have caused the problem (because the PR and bug listed in the release notes suggested that it only affected selecting subfields). It wasn't until we found a work around that we were able to track the issue to this change. Even if there were no prior promises of support of the JSON type, a short blurb in the release notes would be very much appreciated.

PR #1594, opened 2023-06-27
chore: update noxfile for docfx job

dandhlee:
`docfx` _should_ run at `>=3.9`, however I've had some difficulties with some libraries with Sphinx version at 4 and Python 3.10 and above. Pinning to 3.9 for now, while I'm working on fixing this for 3.10 and above. 

Fixes b/289001353 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -425,13 +425,16 @@ def docs(session):
     )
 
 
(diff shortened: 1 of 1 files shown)


chalmerlowe (approved):
LGTM

PR #1601, opened 2023-06-29
fix: add types to DatasetReference constructor

### LINKED ISSUE #1598: Type annotations missing on bigqueryDatasetReference

#### Environment details

  - OS type and version: Windows 11
  - Python version: 3.9.6
  - pip version:  23.1.2
  - `google-cloud-bigquery` version:  3.11.3
  - `mypy` version: 1.4.1

#### Steps to reproduce

  1. Create a file `test.py` with the following contents:
```
ref = bigquery.DatasetReference("my-project", "my_dataset")
```

  2. Run `mypy test.py --disallow-untyped-calls`

#### Code example
See above

#### Stack trace
```
test.py:3: error: Call to untyped function "DatasetReference" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```
#### Solution
To solve the problem, type annotations need to be added to the constructor of bigquery.DatasetReference

kserruys:

Fixes #1598  🦕


chalmerlowe (dismissed):
LGTM

chalmerlowe:
@kserruys let's run kokoro and see what kinda results we get from the tests.

chalmerlowe:
@kserruys 

The two types of `prerelease` dependency test failures are a known issue that I need to tackle and we can safely ignore them in this case.

The `Kokoro` test failure relates to the overall test coverage.
It looks like enough has changed in that portion of code by adding a new conditional branch that the Python library `coverage` feels our tests are not comprehensive enough.

Are you comfortable diving into the test suite to see what needs to be added to the tests to provide coverage for the new conditional?

google/cloud/bigquery/dataset.py
```diff
@@ -166,22 +166,24 @@ def from_string(
...
+            else:
+                raise ValueError(
+                    "When default_project is not set, dataset_id must be a "
+                    "fully-qualified dataset ID in standard SQL format, "
+                    'e.g., "project.dataset_id" got {}'.format(dataset_id)
+                )
         elif len(parts) == 2:
             output_project_id, output_dataset_id = parts
-        elif len(parts) > 2:
+        else:
```
kserruys, line 182: Here `else:` covers the same cases as `elif len(parts) > 2:`

Parameter `dataset_id` is a string, because of that list `parts` will always have at least 1 item.
By using `else` we can reassure `pytest-cov` that everything is ok.

kserruys:
> @kserruys
> 
> The two types of `prerelease` dependency test failures are a known issue that I need to tackle and we can safely ignore them in this case.
> 
> The `Kokoro` test failure relates to the overall test coverage. It looks like enough has changed in that portion of code by adding a new conditional branch that the Python library `coverage` feels our tests are not comprehensive enough.
> 
> Are you comfortable diving into the test suite to see what needs to be added to the tests to provide coverage for the new conditional?

@chalmerlowe 

Sure :). I just took a while to find some time to do this.
Please have a look at the updated code. Coverage on my local machine return 100% now.

Regards

tswast (approved):

tswast:
Thanks @kserruys so much for the contribution and for your patience on this!

PR #1613, opened 2023-07-19
fix: updates typing in function definitions

### LINKED ISSUE #1605: Type annotation incorrect for `job_config` argument(s)
To make mypy happy when I have an optional `job_config`, I'm having to write code like this:

```python
        if job_config is not None:
            query_job = self.bqclient.query(sql, job_config=job_config)
        else:
            query_job = self.bqclient.query(sql)
```

This if/else should be unnecessary, but mypy complains that `job_config` argument to `query` can't be `None`.

The root cause is that the type annotation is only `QueryJobConfig`

https://github.com/googleapis/python-bigquery/blob/130450a3a5d866d2d48ef0c396e9f37d22d0b1d5/google/cloud/bigquery/client.py#L3274

It should be `Optional[QueryJobConfig]`.

This same issue occurs in `copy_table` and `extract_table`.

chalmerlowe:
Adds type annotations for three functions.

@tswast 

Fixes #1605 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -3070,7 +3070,7 @@ class Client(ClientWithProject):
         job_id_prefix: Optional[str] = None,
         location: Optional[str] = None,
         project: Optional[str] = None,
-        job_config: CopyJobConfig = None,
+        job_config: Optional[CopyJobConfig] = None,
         retry: retries.Retry = DEFAULT_RETRY,
         timeout: TimeoutType = DEFAULT_TIMEOUT,
     ) -> job.CopyJob:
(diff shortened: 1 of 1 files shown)


chalmerlowe:
The pre-release fails are due to an inconsistency between `ms` and `ns` with an upstream library. These fails are a known issue being addressed separately.

They should not affect the merge of this PR.

parthea (approved):
LGTM, the errors only appear with newer versions of mypy. Please create a tracking bug to upgrade mypy here: https://github.com/googleapis/python-bigquery/blob/344b7246931e76ea5f507533aa3b81f42ab8c28c/noxfile.py#L25

chalmerlowe:
Created a tracking bug to upgrade mypy.

https://github.com/googleapis/python-bigquery/issues/1614

PR #1616, opened 2023-07-20
feat: search statistics

chalmerlowe:
Adds a feature to display search statistics in support of the following customer request: https://buganizer.corp.google.com/issues/290679162

Includes several classes to encapsulate the data found here: 
https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#searchstatistics

Also includes a function to access the data in the classes.


google/cloud/bigquery/job/query.py
```diff
@@ -858,6 +890,13 @@ def priority(self):
         """
         return self.configuration.priority
 
+    @property
+    def search_stats(self):
+        """docstring"""
+
+        stats = self._job_statistics().get("searchStatistics")
+        return SearchStats.from_api_repr(stats)
```
shollyman, line 898: need to handle the not-present case, e.g. `if stats is not None:`

google/cloud/bigquery/job/query.py
```diff
@@ -198,6 +198,39 @@ def from_api_repr(cls, stats: Dict[str, str]) -> "DmlStats":
...
+
+class SearchStats(typing.NamedTuple):
+    """docstring"""
+
+    mode: str
+    reason: list
+
+    @classmethod
+    def from_api_repr(cls, stats: Dict[str, Any]):
+        mode = stats.get("indexUsageMode", "INDEX_USAGE_MODE_UNSPECIFIED")
```
shollyman, line 227: Generally, we want to avoid the library injecting values not present in the API response, so returning the default value here may not be more desirable than empty/none.

shollyman (commented):
Good start on the basic structure.  This is an output-only stat so no need for a setter, but we'll need all the unit testing and a similar IT test to java using a query that invokes the `search()` sql function.

tests/unit/job/test_query_stats.py
```diff
@@ -108,6 +108,75 @@ def test_from_api_repr_full_stats(self):
...
+        from google.cloud.bigquery.job.query import SearchStats
+
+        return SearchStats
+
+    def _make_one(self, *args, **kwargs):
+        return self._get_target_class()(*args, **kwargs)
+
+    def test_ctor_defaults(self):
+        search_stats = self._make_one()
+        assert search_stats.mode == "INDEX_USAGE_MODE_UNSPECIFIED"
```
chalmerlowe, line 123:
```suggestion
        assert search_stats.mode == ''
```

tests/unit/job/test_query_stats.py
```diff
@@ -108,6 +108,75 @@ def test_from_api_repr_full_stats(self):
...
+        return SearchReason
+
+    def _make_one(self, *args, **kwargs):
+        return self._get_target_class()(*args, **kwargs)
+
+    def test_ctor_defaults(self):
+        search_reason = self._make_one()
+        assert search_reason.code == ""
+        assert search_reason.message == ""
+        assert if search_reason.baseTable is None
```
chalmerlowe, line 151:
```suggestion
        assert search_reason.baseTable is None
```

tests/unit/job/test_query_stats.py
```diff
@@ -108,6 +108,75 @@ def test_from_api_repr_full_stats(self):
...
+
+        assert isinstance(result, klass)
+        assert result.code == "INDEX_CONFIG_NOT_AVAILABLE"
+        assert result.message == "There is no search index..."
+        assert result.baseTable == {
+            "projectId": "bigquery-public-data",
+            "datasetId": "usa_names",
+            "tableId": "usa_1910_current",
+        }
+        assert if result.indexName is None
```
chalmerlowe, line 177:
```suggestion
        assert result.indexName is None
```

shollyman (commented):
This PR appears to include pandas fixes which seem out of scope for the PR at hand.  Is it based on another PR yet to be merged?

google/cloud/bigquery/job/query.py
```diff
@@ -198,6 +198,39 @@ def from_api_repr(cls, stats: Dict[str, str]) -> "DmlStats":
         return cls(*args)
 
 
+class SearchReason(typing.NamedTuple):
```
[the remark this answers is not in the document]
chalmerlowe: Regarding naming:
I simply mirrored naming conventions that were already present in the code: 

`BiEngineStats`
`BiEngineReason`

My take was that `SearchReason` and `SearchStats` specifically highlights the broader purpose/use of the Reason (i.e. this reason is related to the Search function):

If we think we get greater clarity by calling the class IndexUnusedReason, I am game to do that.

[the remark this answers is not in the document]
chalmerlowe: Can you elaborate on what you mean by property definitions?

Are to talking about expanding on the docstring and parameter definitions? If yes, that is yet to be done. At this second, I am focused on does the code work/seem reasonable/have adequate tests.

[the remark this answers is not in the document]
shollyman: I mostly mean getters with the "@property" annotation for referencing fields in the message types.  As these are primarily readonly fields setters shouldn't be necessary.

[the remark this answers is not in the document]
chalmerlowe: @shollyman I got the pandas references out.
I will see about renaming the underlying type.
Did you come across any ideas on whether we need/want to add @properties when using a `typing.NamedTuple`

tests/unit/job/test_query_stats.py
```diff
@@ -108,6 +108,75 @@ def test_from_api_repr_full_stats(self):
...
+        result = klass.from_api_repr(
+            {"indexUsageMode": "INDEX_USAGE_MODE_UNSPECIFIED", "indexUnusedReasons": []}
+        )
+
+        assert isinstance(result, klass)
+        assert result.mode == "INDEX_USAGE_MODE_UNSPECIFIED"
+        assert result.reason == []
+
+
+class TestSearchReasons:
```
chalmerlowe, line 137:
```suggestion
class TestIndexUnusedReason:
```

tests/unit/job/test_query_stats.py
```diff
@@ -108,6 +108,75 @@ def test_from_api_repr_full_stats(self):
...
+    @staticmethod
+    def _get_target_class():
+        from google.cloud.bigquery.job.query import SearchReason
+
+        return SearchReason
```
chalmerlowe, line 142:
```suggestion
        from google.cloud.bigquery.job.query import IndexUnusedReason

        return IndexUnusedReason
```

google/cloud/bigquery/job/query.py
```diff
@@ -198,6 +198,59 @@ def from_api_repr(cls, stats: Dict[str, str]) -> "DmlStats":
         return cls(*args)
 
 
+class IndexUnusedReason(typing.NamedTuple):
+    """Reason about why no search index was used in the search query (or sub-query).
+
+    https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#indexunusedreason
+    """
+
+    code: Optional[str] = ""
```
shollyman, line 207: Should these attribute type hints be`Optional[string] = None`?

google/cloud/bigquery/job/query.py
```diff
@@ -198,6 +198,59 @@ def from_api_repr(cls, stats: Dict[str, str]) -> "DmlStats":
...
+
+    mode: str = ""
+    """Indicates the type of search index usage in the entire search query."""
+
+    reason: List[IndexUnusedReason] = []
+    """Reason about why no search index was used in the search query (or sub-query)"""
+
+    @classmethod
+    def from_api_repr(cls, stats: Dict[str, Any]):
+        mode = stats.get("indexUsageMode", "")
```
shollyman, line 246: Should the default be None here as well, rather than empty string?

google/cloud/bigquery/job/query.py
```diff
@@ -858,6 +910,14 @@ def priority(self):
         """
         return self.configuration.priority
 
+    @property
+    def search_stats(self):
```
shollyman, line 914: This needs typing, and docstring should likely some variation of "returns" vs "generates" when describing the property.

shollyman (commented):
After some more reading, I think the attribute approach is sufficient vs explicit property annotations.

Main concern is about attribute defaults.  Typically None is better when users are trying to disambiguate between presence of a value vs the empty value, but happy to talk about this more.  Swapping the default impacts typing, unit tests, etc.

google/cloud/bigquery/job/query.py
```diff
@@ -198,6 +198,59 @@ def from_api_repr(cls, stats: Dict[str, str]) -> "DmlStats":
...
+
+    mode: Optional[str] = None
+    """Indicates the type of search index usage in the entire search query."""
+
+    reason: List[IndexUnusedReason] = []
+    """Reason about why no search index was used in the search query (or sub-query)"""
+
+    @classmethod
+    def from_api_repr(cls, stats: Dict[str, Any]):
+        mode = stats.get("indexUsageMode", "")
```
chalmerlowe, line 246:
```suggestion
        mode = stats.get("indexUsageMode", None)
```

shollyman (approved):

Issue #1623, opened 2023-07-24
Extend the OTEL traces to track bytes billed and bytes processed


MrwanBaghdad:
Extend OTEL traces to track bytes billed and bytes processed. Allowing developers to see how much data and cost was consumed as part of BigQueryJob trace

PR #1624, opened 2023-07-24
feat: Update opentel to track bytes billed and processed for job queries

MrwanBaghdad:

Update opentel tracing tracking bytes billed and processed for job queries
Link to [feature request](https://github.com/googleapis/python-bigquery/issues/1623)


chalmerlowe:
@tswast 

In looking at this, as best I can tell, this looks like it relies upon [AsyncJob](https://github.com/googleapis/python-bigquery/blob/21714e18bad8d8d89ed5642dbdb61d14e97d5f33/google/cloud/bigquery/job/base.py#L380), which we define elsewhere but when we define it, we don't initialize the attributes listed in this PR: namely `total_bytes_billed` and `total_bytes_processed`.

I am considering going in an seeing about adding those attributes to AsyncJob, so that we can access them here.

Am I missing something?
Is there a better approach?

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -152,5 +152,11 @@ def _set_job_attributes(job_ref):
 
     if job_ref.num_child_jobs is not None:
         job_attributes["num_child_jobs"] = job_ref.num_child_jobs
+    
+    if job_ref.total_bytes_billed is not None:
+        job_attributes["total_bytes_billed"] = int(job_ref.total_bytes_billed)
```
tswast, line 157: Since this property only applies query jobs (https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobstatistics2), but this code runs for all job types, let's protect this with [getattr](https://docs.python.org/3.12/library/functions.html#getattr) with a default.

Also, I don't think the conversion to `int` is required, right?

```suggestion
    total_bytes_billed = getattr(job_ref, "total_bytes_billed", None)
    if total_bytes_billed is not None:
        job_attributes["total_bytes_billed"] = total_bytes_billed
```

google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -152,5 +152,11 @@ def _set_job_attributes(job_ref):
...
+
+    if job_ref.total_bytes_processed is not None:
+        job_attributes["total_bytes_processed"] = int(job_ref.total_bytes_processed)
```
tswast, line 160: Same here

```suggestion
    total_bytes_processed = getattr(job_ref, "total_bytes_processed", None)
    if total_bytes_processed is not None:
        job_attributes["total_bytes_processed"] = total_bytes_processed
```

tswast (commented):
We'll need some unit test coverage before merging this, too.

tswast:
> I am considering going in an seeing about adding those attributes to AsyncJob, so that we can access them here.

I left some suggestions. I don't want to mislead folks by having these properties available on LoadJob for example, which is free and doesn't have such properties.

chalmerlowe:
This has been migrated to a [new PR](https://github.com/googleapis/python-bigquery/pull/1889).
Closing.

PR #1656, opened 2023-09-12
Expose jobTimeoutMs in JobConfig

### LINKED ISSUE #1421: Expose jobTimeoutMs field in JobConfigurationQuery
At job creation time I want to set a timeout for the job if it takes too long to run. Too many long running queries eat up the slot reservation and significantly deteriorates query performance. Currently we have to manually cancel the queries in the console. I noticed that there is a `jobTimeoutMs` field in the [JobConfiguration](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobconfiguration) object. This is not exposed by the Python client but is exposed by other Bigquery clients ([Golang](https://github.com/googleapis/google-cloud-go/pull/5707)). Can we add this to the Python client as well?

The alternative would be to have a timer on the client side and manually cancel the request. This feels wasteful considering the feature already exists on the server side and it's just a question of exposing it in the client.

lbk-fishtown:
Fixes #1421

Exposes job timeout setting in Python job timeout config.
        https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfiguration.FIELDS.job_timeout_ms

```
# python3 test.py 5000
Client creating using default project: demo-project-xyz
Job config: {'query': {}, 'jobTimeoutMs': '5000'}
Trying to execute long (30s) running query: call `demo-project-xyz`.`dbt_lbondkennedy`.`sleep_seconds_approximate`(30);
An exception of type GoogleAPICallError occurred. Arguments:
('Job execution was cancelled: Job timed out after 5 sec',)
```

image

---

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -172,7 +172,7 @@ class _JobConfig(object):
         super(_JobConfig, self).__setattr__(name, value)
 
     @property
-    def jobtimeout(self):
+    def job_timeout_ms(self):
         """ Optional parameter. Job timeout in milliseconds. If this time limit is exceeded, BigQuery might attempt to stop the job.
 
         https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfiguration.FIELDS.job_timeout_ms
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -192,8 +192,8 @@ class _JobConfig(object):
             return self._properties["jobTimeoutMs"]
         return None
 
-    @jobtimeout.setter
-    def jobtimeout(self, value):
+    @job_timeout_ms.setter
+    def job_timeout_ms(self, value):
         if not isinstance(value, int):
             raise ValueError("Pass an int for jobTimeoutMs, e.g. 5000")
         
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -179,7 +179,6 @@ class _JobConfig(object):
 
         e.g.
             job_config = bigquery.QueryJobConfig( jobtimeout = 5000 )
-            job_config = bigquery.QueryJobConfig()
             
                 or
 
@@ -188,7 +187,7 @@ class _JobConfig(object):
             ValueError: If ``value`` type is invalid.
         """
         
-        # None as this does is an optional parameter.
+        # None as this is an optional parameter.
         return None
 
     @jobtimeout.setter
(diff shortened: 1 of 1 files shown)


chalmerlowe:
@lbk-fishtown 

Thanks for submitting this PR.
In terms of tests, a fairly straightforward test similar to this one ought to be a good place to start.

`labels` are an exposed property, much like the `jobTimeoutMs` property and are set in a similar way, so mirroring this test should work for our purposes.

https://github.com/googleapis/python-bigquery/blob/03194e0156ed9201cb36301967c5af117d7ef29c/tests/unit/model/test_model.py#L343C1-L348C42

I suspect that we might want to put your tests in the following file: `python-bigquery/tests/unit/job/test_query.py`

We may want to go further than just checking to see if we can set the `jobTimeoutMs` value, but for now, let's just get that functionality working while we come up with the list of cases to include as tests.

google/cloud/bigquery/job/base.py
```diff
@@ -171,6 +171,33 @@ def __setattr__(self, name, value):
             )
         super(_JobConfig, self).__setattr__(name, value)
 
+    @property
+    def jobtimeout(self):
```
tswast, line 175: nit: normally we use snake_case to transform from rest API to python.

I'd prefer `job_timeout_ms` unless we change this to accept a `datetime.timedelta`, in which case `job_timeout` may make more sense.

tswast: FWIW: We use `default_table_expiration_ms` here: https://github.com/googleapis/python-bigquery/blob/54a77694afcd80be4ba469c6ebb7ca8be112b04e/google/cloud/bigquery/dataset.py#L660 so probably makes the most sense to stay consistent with that.

chalmerlowe, line 175:
```suggestion
    def job_timeout_ms(self):
```

google/cloud/bigquery/job/base.py
```diff
@@ -171,6 +171,33 @@ def __setattr__(self, name, value):
...
+            job_config = bigquery.QueryJobConfig( jobtimeout = 5000 )
+            
+                or
+
+            job_config.jobtimeout = 5000
+        Raises:
+            ValueError: If ``value`` type is invalid.
+        """
+        
+        # None as this is an optional parameter.
```
chalmerlowe, line 190:
```suggestion
        # None as this is an optional parameter.
        if self._properties.get("jobTimeoutMs"):
            return self._properties["jobTimeoutMs"]
```

google/cloud/bigquery/job/base.py
```diff
@@ -171,6 +171,33 @@ def __setattr__(self, name, value):
...
+        return None
+
+    @jobtimeout.setter
+    def jobtimeout(self, value):
```
chalmerlowe, line 194:
```suggestion
    @job_timeout_ms.setter
    def job_timeout_ms(self, value):
```

chalmerlowe:
This PR was superceded by [PR 1675](https://github.com/googleapis/python-bigquery/pull/1675).
Closing.

PR #1643, opened 2023-08-16
feat: implementing Storage Billing model property to BQ Datasets

sriramsridhar:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes Issue #1639 🦕

PR #1665, opened 2023-10-02
fix: allow `storage_billing_model` to be explicitly set to `None` to use project default value

tswast:
Follow-up to https://github.com/googleapis/python-bigquery/pull/1643

Fixes internal issue 302750597 🦕


chalmerlowe:
This does not account for the fourth option that is available:

`STORAGE_BILLING_MODEL_UNSPECIFIED`

That should at least be called out in the comment.

@tswast 
@Linchin

tswast:
> That should at least be called out in the comment.

Added!

IMO, the less client-side logic the better, but adding this in the docs makes sense to me.

chalmerlowe (approved):
LGTM

PR #1680, opened 2023-10-11
chore: clean up BigQueryStorage import checks

### LINKED ISSUE #1430: cleanup: Use logic from BQStorageVersions and PyarrowVersions instead of comparing version strings outside of _helpers
Ideally we'd never need to compare PyArrow versions outside of the [PyarrowVersions](https://github.com/googleapis/python-bigquery/blob/40e4da78bb690ff4c94832321377bb1590e2eeaf/google/cloud/bigquery/_helpers.py#L97) class. We might need to add some additional methods / properties to do this.

Follow-up to https://github.com/googleapis/python-bigquery/pull/1282/files#discussion_r927998418

Linchin:
Follow up of #1679

Also part of fixing #1430, this PR cleans up code that verifies BigQueryStorage extra can be successfully imported. It also:

* Modifies some lines to comply with PEP-8 line length limit
* Adds unit test for the added method

Fixes #1430 🦕

### DIFF
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -35,6 +35,7 @@ import packaging.version
 from google.cloud.bigquery.exceptions import (
     LegacyBigQueryStorageError,
     LegacyPyarrowError,
+    BigQueryStorageNotFoundError,
 )
 
 _RFC3339_MICROS_NO_ZULU = "%Y-%m-%dT%H:%M:%S.%f"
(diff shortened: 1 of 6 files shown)


google/cloud/bigquery/_helpers.py
```diff
@@ -120,11 +121,34 @@ def verify_version(self):
...
             msg = (
-                "Dependency google-cloud-bigquery-storage is outdated, please upgrade "
-                f"it to version >= {_MIN_BQ_STORAGE_VERSION} (version found: {self.installed_version})."
+                "Dependency google-cloud-bigquery-storage is outdated, "
+                f"please upgrade it to version >= {_MIN_BQ_STORAGE_VERSION} "
+                f"(version found: {self.installed_version})."
             )
             raise LegacyBigQueryStorageError(msg)
 
+    def try_import(self) -> Any:
```
tswast, line 130: Love it, but let's move to the same `_version_helpers.py` module as we're doing in https://github.com/googleapis/python-bigquery/pull/1679#discussion_r1358762921

google/cloud/bigquery/_helpers.py
```diff
@@ -120,11 +121,34 @@ def verify_version(self):
...
+        try:
+            from google.cloud import bigquery_storage  # type: ignore
+        except ImportError:
+            msg = (
+                "Package google-cloud-bigquery-storage not found. "
+                "Install google-cloud-bigquery-storage version >= "
+                f"{_MIN_BQ_STORAGE_VERSION}."
+            )
+            raise BigQueryStorageNotFoundError(msg)
+        return bigquery_storage
```
tswast, line 150: Let's do the same kind of version checks as we do for pyarrow. We can `raise_if_error: bool = False` like we have there too and remove the redundant `verify_version()` method.

google/cloud/bigquery/client.py
```diff
@@ -564,8 +567,8 @@ def _ensure_bqstorage_client(
             A BigQuery Storage API client.
         """
         try:
-            from google.cloud import bigquery_storage  # type: ignore
-        except ImportError:
+            bigquery_storage = BQ_STORAGE_VERSIONS.try_import()
+        except BigQueryStorageNotFoundError:
```
tswast, line 571: If we implement `try_import(raise_if_error=True)`, I think we can combine this try/except block and the one where we call `verify_version()` into one.

tswast (changes requested):

google/cloud/bigquery/_helpers.py
Linchin: Done, changed made at #1699

google/cloud/bigquery/_helpers.py
Linchin: Done, change made at #1699

google/cloud/bigquery/client.py
Linchin: Done, change made at #1699

Linchin:
Closed due to too much merge conflicts, use #1699 instead.

PR #1690, opened 2023-10-18
docs: removed a letter from docstring

kiraksi:


google/cloud/bigquery/table.py
```diff
@@ -301,7 +301,7 @@ def to_bqstorage(self) -> str:
         Install the ``google-cloud-bigquery-storage`` package to use this
         feature.
 
-        If the ``table_id`` contains a partition identifier (e.g.
```
dandhlee, line 304: This is the correct form.

PR #1698, opened 2023-10-23
fix: load_table_from_dataframe method for issue 1692

### LINKED ISSUE #1692: load_table_from_dataframe does not error out when `nan` in a required column - Million dollar bug
When loading a pandas dataframe into Bigquery that contains a nan in a required column, the upload succeeds but the resulting table is not representative of the dataframe. The values in the column containing the `nan` are unordered and the user is not aware of it.

#### Environment details

  - OS type and version:
  - Python version: Python 3.11.4
  - pip version: pip 23.2.1
  - `google-cloud-bigquery` version: 
  - Name: google-cloud-bigquery
  ```
Version: 3.12.0
Summary: Google BigQuery API client library
Home-page: https://github.com/googleapis/python-bigquery
Author: Google LLC
Author-email: googleapis-packages@google.com
License: Apache 2.0
Requires: google-api-core, google-cloud-core, google-resumable-media, grpcio, grpcio, packaging, proto-plus, protobuf, python-dateutil, requests
Required-by: google-cloud-aiplatform, pandas-gbq
  ```

#### Steps to reproduce

  1. Run the code below

#### Code example

```python
import pandas as pd
import numpy as np

df = pd.DataFrame([["hello", "string"], ["hello2", np.nan], ["hello3", "valid"], ["hello4", "valid2"]], columns=["image_uri", "phash"])

client = bigquery.Client(project="project")
job_config = bigquery.LoadJobConfig(
    schema=[
        bigquery.SchemaField("image_uri", "STRING", mode="REQUIRED"),
        bigquery.SchemaField("phash", "STRING", mode="REQUIRED"),
    ]
)
job = client.load_table_from_dataframe(
    df, "foo.foo_bar", job_config=job_config
)
job.result()

df_read = pd.read_gbq("foo.foo_bar", project_id="project")
```
image
image

Gaurang033:
Fixes #1692


dandhlee:
Please keep titles less than 50 characters.

dandhlee (changes requested):

tests/unit/test_client.py
```diff
@@ -8689,6 +8689,32 @@ def test_load_table_from_dataframe_w_nulls(self):
...
+            SchemaField("name", "STRING"),
+            SchemaField("age", "INTEGER", mode="REQUIRED"),
+        ]
+        job_config = job.LoadJobConfig(schema=schema)
+        with pytest.raises(ValueError) as e:
+            client.load_table_from_dataframe(
+                dataframe, self.TABLE_REF, job_config=job_config, location=self.LOCATION
+            )
+
+            assert str(e.value) == "required field age can not be nulls"
```
dandhlee, line 8716: When the exception is thrown, it returns to the context manager (pytest) and assertion is never reached. See the test below for an example for testing the message value.

Gaurang033: oh yes.. fixed it.

Gaurang033:
@dandhlee  could you please review the changes.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -302,6 +302,20 @@ def bq_to_arrow_array(series, bq_field):
     return pyarrow.Array.from_pandas(series, type=arrow_type)
 
 
+def _check_nullability(arrow_fields, dataframe):
+    """Throws error if dataframe has null values and column doesn't allow nullable"""
+    if dataframe.index.name:
```
Linchin, line 307: Could you please help me understand what these two lines are for?

Gaurang033: It's hard to let you know which exact code as I am on vacation. But when dataframe with index is used, index name is transformed as airow column name. There were two way to fix it. One was to put exception for this case or the create another column with index name. I choose the second option as it's easier. Without this the dataframe unit test case where they use index names will fail.

Linchin: Thank you, hope you are having a good time in vacation! I played with dataframe's index a little bit, and I think there are several corner cases (which are likely non-comprehensive) that we need to cover:
- Index doesn't have a name, does it still get converted into arrow?
- Multiple index
- Index with the same name as columns, which is possible with dataframes
- Index columns have the same names (possible too)
- multiindex

chalmerlowe: @gaurang033 Thanks for offering up this PR.
@Linchin I appreciate this summary of additional edge cases that may not be covered by this solution.

I too worry about the edge cases, but more importantly, I worry about spending too much time and energy trying to create a work around for what we all agree is a problem in `pyarrow`. This feels like it creates greater complexity in our code, increased fragility, and a higher maintenance burden in the long run. Am I missing something?

Linchin: I am also uncertain if we should add logic in our repo to correct an issue with pyarrow. I have been thinking about this PR as more of a temporary patchwork that maybe reverted later, but for now does help our customers. However if the logic covering the corner cases get too convoluted with the behaviors of pyarrow, I agree that perhaps it's a better idea to open an issue with pyarrow instead.

tswast:
Let's just let the server-side determine if we aren't matching the correct schema. I propose https://github.com/googleapis/python-bigquery/pull/1735 instead.

PR #1708, opened 2023-10-30
feat: support data_governance_type

### LINKED ISSUE #1707: support data_governance_type when using routine

Linchin:

Fixes #1707 🦕

### DIFF
--- a/google/cloud/bigquery/routine/routine.py
+++ b/google/cloud/bigquery/routine/routine.py
@@ -68,6 +68,12 @@ class Routine(object):
         "description": "description",
         "determinism_level": "determinismLevel",
         "remote_function_options": "remoteFunctionOptions",
+        "data_governance_type": "dataGovernanceType",
+    }
+
+    _DATA_GOVERNANCE_TYPE = {
+        "DATA_MASKING": None,
+        "DATA_GOVERNANCE_TYPE_UNSPECIFIED": None,
     }
 
     def __init__(self, routine_ref, **kwargs) -> None:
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/routine/routine.py
```diff
@@ -329,6 +335,27 @@ def remote_function_options(self, value):
...
+        Raises:
+            ValueError:
+                If the value is not :data:`string` or :data:`None`, or the value
+                is not in `self._DATA_GOVERNANCE_TYPE`.
+        """
+        return self._properties.get(self._PROPERTY_TO_API_FIELD["data_governance_type"])
+
+    @data_governance_type.setter
+    def data_governance_type(self, value):
+        if value is not None and value not in self._DATA_GOVERNANCE_TYPE:
```
shollyman, line 352: Generally we try to avoid this kind of setter enforcement in the client library itself, as the library may not be aware of all valid values.  Allow the methods to raise errors if there's issues (e.g at routine creation/update/etc).

shollyman (commented):
Is it worth adding the new field to an existing system/integration test related to routines, or are there extensive dependencies here to properly setup masking?

google/cloud/bigquery/routine/routine.py
Linchin: Thank you, this is really good to know! I will only check for type instead.

Linchin:
> Is it worth adding the new field to an existing system/integration test related to routines, or are there extensive dependencies here to properly setup masking?

Data governance feels like a new CUJ that is not covered by the current system tests, with a unique set of requirements for the routine itself (e.g., can have only one argument, input and output must be of the same type, etc.). So I added a test case for it.

shollyman (approved):

PR #1699, opened 2023-10-23
chore: refactor BigQueryStorageVersions

Linchin:
Follow up of #1679
Replaces #1680, due to too much merge conflicts

Also part of fixing #1430, this PR cleans up code that verifies BigQueryStorage extra can be successfully imported. It also:

* Modifies some lines to comply with PEP-8 line length limit
* Adds unit test for the added method

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #1430 🦕

PR #1711, opened 2023-11-02
chore: refactor version checks for pandas library

### LINKED ISSUE #769: refactor version checks for pandas, pyarrow
We're using `pkg_resources.parse_version` from setuptools, which is just a thin wrapper over [packaging](https://packaging.pypa.io/en/latest/version.html).

- parse_version implementation: https://github.com/pypa/setuptools/blob/a4dbe3457d89cf67ee3aa571fdb149e6eb544e88/pkg_resources/__init__.py#L112
- Our usage: https://github.com/googleapis/python-bigquery/search?q=pkg_resources

Also, we have some version comparisons for features. I think it'd be cleaner to implement these as properties like we do in `BQStorageVersions` that I propose adding in https://github.com/googleapis/python-bigquery/pull/748

```python
import packaging.version


_MIN_BQ_STORAGE_VERSION = packaging.version.Version("2.0.0")
_BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION = packaging.version.Version("2.6.0")


class BQStorageVersions:
    def __init__(self):
        self._installed_version = None

    @property
    def installed_version(
        self,
    ) -> Union[packaging.version.LegacyVersion, packaging.version.Version]:
        if self._installed_version is None:

            self._installed_version = packaging.version.parse(
                getattr(bigquery_storage, "__version__", "legacy")
            )

        return self._installed_version

    @property
    def is_read_session_optional(self) -> bool:
        return self.installed_version >= _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION

    def verify_version(self):
        """Verify that a recent enough version of BigQuery Storage extra is installed.

        The function assumes that google-cloud-bigquery-storage extra is installed, and
        should thus be used in places where this assumption holds.

        Because `pip` can install an outdated version of this extra despite the constraints
        in setup.py, the the calling code can use this helper to verify the version
        compatibility at runtime.

        Raises:
            LegacyBigQueryStorageError: If google-cloud-bigquery-storage is outdated.
        """
        if self.installed_version < _MIN_BQ_STORAGE_VERSION:
            msg = (
                "Dependency google-cloud-bigquery-storage is outdated, please upgrade "
                f"it to version >= 2.0.0 (version found: {self.installed_version})."
            )


BQ_STORAGE_VERSIONS = BQStorageVersions()
```

kiraksi:
Builds off of #1699 

Fixes #769 🦕

### DIFF
--- a/google/cloud/bigquery/_versions_helpers.py
+++ b/google/cloud/bigquery/_versions_helpers.py
@@ -223,7 +223,7 @@ class PandasVersions:
         if self.installed_version < _MIN_PANDAS_VERSION:
             if raise_if_error:
                 msg = (
-                    "Dependency pyarrow is outdated, please upgrade"
+                    "Dependency pandas is outdated, please upgrade"
                     f" it to version >= {_MIN_PANDAS_VERSION}"
                     f" (version found: {self.installed_version})."
                 )
--- a/google/cloud/bigquery/_versions_helpers.py
+++ b/google/cloud/bigquery/_versions_helpers.py
@@ -24,6 +24,8 @@ from google.cloud.bigquery import exceptions
 _MIN_PYARROW_VERSION = packaging.version.Version("3.0.0")
 _MIN_BQ_STORAGE_VERSION = packaging.version.Version("2.0.0")
 _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION = packaging.version.Version("2.6.0")
+_MIN_PANDAS_VERSION = packaging.version.Version("1.1.0")
+_MAX_PANDAS_VERSION = packaging.version.Version("2.0.0")
 
 
 class PyarrowVersions:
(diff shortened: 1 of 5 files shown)


Linchin:
Thank you @kiraksi for opening your first PR! Looks like the mypy tests are failing:
```
nox > * mypy: failed
nox > * mypy_samples: failed
```
Please try to run the same nox sessions locally, and fix the errors (likely some format issues).

google/cloud/bigquery/_versions_helpers.py
```diff
@@ -24,6 +24,8 @@
 _MIN_PYARROW_VERSION = packaging.version.Version("3.0.0")
 _MIN_BQ_STORAGE_VERSION = packaging.version.Version("2.0.0")
 _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION = packaging.version.Version("2.6.0")
+_MIN_PANDAS_VERSION = packaging.version.Version("1.1.0")
+_MAX_PANDAS_VERSION = packaging.version.Version("2.0.0")
```
tswast, line 28: I would prefer we don't enforce this. It will break BigQuery DataFrames, as we use pandas 2.0+ with this library just fine.

google/cloud/bigquery/client.py
```diff
@@ -2488,7 +2488,7 @@ def load_table_from_file(
 
     def load_table_from_dataframe(
         self,
-        dataframe: "pandas.DataFrame",
+        dataframe: "pandas.DataFrame", # type: ignore
```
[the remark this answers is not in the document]
tswast: If we do, add `pandas-stubs` to our pandas extra.

tests/unit/test__versions_helpers.py
```diff
@@ -173,3 +178,64 @@ def test_bqstorage_is_read_session_optional_false():
...
+        with pytest.raises(exceptions.LegacyPandasError):
+            versions.try_import(raise_if_error=True)
+
+
+@pytest.mark.skipif(pandas is None, reason="pandas is not installed")
+def test_try_import_returns_none_w_newer_pandas():
+    versions = _versions_helpers.PandasVersions()
+    with mock.patch("pandas.__version__", new="2.0.0"):
+        pandas = versions.try_import()
+        assert pandas is None
```
tswast, line 215: Let's actually test that it's not `None` :-)

tswast (changes requested):

google/cloud/bigquery/client.py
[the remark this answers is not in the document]
kiraksi: On this, I attempted to add pandas-stubs as a pandas extra in setup.py, however it did not fix the mypy issue, is there an added step to this?

tests/unit/test__versions_helpers.py
kiraksi: I am not sure what you meant, I have a test above this one that tests if it raises no error if its a supported version f pandas already, if that is what you meant

chalmerlowe (approved):
LGTM.
Confirm with Tim regarding the is not None comment.

google/cloud/bigquery/client.py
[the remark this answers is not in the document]
tswast: The nox session for mypy might not actually be using our extras. Edit: it is, hmm... https://github.com/googleapis/python-bigquery/blob/main/noxfile.py#L129

You might need to also add this to the top of the module:

```
if typing.TYPE_CHECKING:
    import pandas
```

tests/unit/test__versions_helpers.py
tswast: Yes. I want a test where we check for some 2.x version.

kiraksi: Implemented fix through pytest.parameterize in newest commit

tswast (approved):
Thanks!

PR #1720, opened 2023-11-13
feat: Introduce compatibility with native namespace packages

### LINKED ISSUE #313: use native support for namespace packages
Per: https://github.com/googleapis/google-auth-library-python/issues/595#issuecomment-683903062

> It looks like different Google API packages have different approaches to using or not using pkg_resources here; e.g. https://github.com/googleapis/python-cloudbuild uses native NS packages and https://github.com/googleapis/python-bigquery uses pkg_resources with a pkgutil fallback.

Now that Python 2.7 and 3.5 support has been dropped, we may be able to follow the example in python-cloudbuild

parthea:
Similar to https://github.com/googleapis/google-auth-library-python/pull/1205
Also see https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages

Fixes #313 🦕


parthea:
Adding `do not merge` as there is an upcoming release freeze. This should be merged after the release freeze ends on November 29th.

tests/unit/test_packaging.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_namespace_package_compat(tmp_path):
+    # The ``google`` namespace package should not be masked
+    # by the presence of ``google-cloud-bigquery``.
+    google = tmp_path / "google"
+    google.mkdir()
+    google.joinpath("othermod.py").write_text("")
+    env = dict(os.environ, PYTHONPATH=str(tmp_path))
+    cmd = [sys.executable, "-m", "google.othermod"]
+    subprocess.check_call(cmd, env=env)
```
kiraksi, line 28: Were you able to run this test with subprocess without specifying to do it in the shell? It wasn't able to find the tmp path without me specifying to run in the shell, which is the small edit I did in the other PR for this

kiraksi (approved):
LGTM

parthea:
Superseded by https://github.com/googleapis/python-bigquery/pull/1726

Issue #790, opened 2021-07-21
standardize samples directory layout


tswast:
We have an odd mix of sample locations:

- `docs/snippets.py` -- the original location, not very copy-pasteable / browsable
- lots of samples in `samples/*.py`, but the test drivers are in `samples/tests/`
- `samples/snippets/*.py`
- `samples/geography/*.py`

To align with the infrastructure in https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/python_samples, the goal is to have:

- `samples/snippets/` -- these samples just require the `google-cloud-bigquery` library 
- `samples/geography/` -- require extra dependencies for working with geography data
- `sample/pandas/` -- require extra dependencies for working with pandas dataframes
- more??? -- possibly a directory for the user-based authentication samples

To avoid breaking cloud.google.com docs, this needs to be done in several steps:

1. Copy samples to correct location
2. Update docs (Googler must do this)
3. Delete duplicate samples in old location

Towards: https://github.com/googleapis/python-bigquery/issues/366 (use templated noxfile) and ~https://github.com/googleapis/python-bigquery/issues/765 (blacken samples using templated noxfile(s))~

PR #1721, opened 2023-11-13
fix: move grpc, proto-plus and protobuf packages to extras

### LINKED ISSUE #1431: Make compiled packages such as "grpc" and "proto-plus" optional extras
I believe proto-plus / protobuf / grpc packages were only needed for `google.cloud.bigquery_v2`. These modules have been deprecated and disconnected from the `google.cloud.bigquery` modules in version 3.x. I believe this means we can move those dependencies to "extras" (a.k.a. optional dependencies) and make our package lighter-weight by default.

Follow-up to https://github.com/googleapis/python-bigquery/pull/1282/files#discussion_r928022130 and https://github.com/googleapis/python-bigquery/pull/1282/files#discussion_r928022317

kiraksi:
This PR:
- Makes the package lighter weight by moving dependencies grpc, proto-plus and protobuff, which are needed for bigquery_v2, into optional extras
- This will throw an import error if a user tries to access parts of bigquery_v2 without installing these packages, which they can then fix by importing them, as this version is deprecated

Fixes #1431 🦕


setup.py
```diff
@@ -29,20 +29,16 @@
...
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
-    "grpcio >= 1.47.0, < 2.0dev",  # https://github.com/googleapis/python-bigquery/issues/1262
-    "grpcio >= 1.49.1, < 2.0dev; python_version>='3.11'",
     # NOTE: Maintainers, please do not require google-api-core>=2.x.x
     # Until this issue is closed
     # https://github.com/googleapis/google-cloud-python/issues/10566
-    "google-api-core[grpc] >= 1.31.5, <3.0.0dev,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.0",
-    "proto-plus >= 1.15.0, <2.0.0dev",
+    "google-api-core >= 1.31.5, <3.0.0dev,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.0",
```
chalmerlowe, line 35: Is there a reason we are not referencing the `[grpc]` install of `google-api-core`?

chalmerlowe (commented):
Thanks for the hard work. Appreciate the time and effort.
I added several comments.

setup.py
```diff
@@ -29,20 +29,16 @@
 # 'Development Status :: 5 - Production/Stable'
 release_status = "Development Status :: 5 - Production/Stable"
 dependencies = [
-    "grpcio >= 1.47.0, < 2.0dev",  # https://github.com/googleapis/python-bigquery/issues/1262
-    "grpcio >= 1.49.1, < 2.0dev; python_version>='3.11'",
     # NOTE: Maintainers, please do not require google-api-core>=2.x.x
```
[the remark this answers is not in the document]
kiraksi: I think we mentioned this at our last meeting, but it was because grpcio was both in dependencies and bqstorage optional extras, I believe it was supposed to be moved there but was not until the referenced issue above it was closed

setup.py
kiraksi: Per my conversation with Tim Swast, there was no longer a reason to since it is now an optional dependency? Maybe he can answer this question better @tswast

tswast: `grpc` will pull in protobuf and grpc libraries. I don't think we actually need those for the core REST API stuff.

tswast (commented):
LGTM!

We might actually want to add a `system_noextras` or move out the samples from `samples/snippets` that need extras (https://github.com/googleapis/python-bigquery/issues/790#issue-949975519) so that we have a little more confidence that these dependencies are actually optional.

I'm pretty sure they are, but it'd be good to verify with a bit more testing.

parthea (commented):
Consider changing the conventional commit message to `fix:` instead of `chore:` so that it's visible in the public release notes

chalmerlowe:
> LGTM!
> 
> We might actually want to add a `system_noextras` or move out the samples from `samples/snippets` that need extras ([#790 (comment)](https://github.com/googleapis/python-bigquery/issues/790#issue-949975519)) so that we have a little more confidence that these dependencies are actually optional.
> 
> I'm pretty sure they are, but it'd be good to verify with a bit more testing.

@tswast was the intent to have this done as part of this PR OR create/piggyback off another Issue, such as 790?

tswast:
@chalmerlowe I have synced with @kiraksi about this. I would like https://github.com/googleapis/python-bigquery/pull/1732 merged first, which makes sure we have some sample tests in samples/snippets with no other dependencies installed.

tswast (approved):
LGTM, assuming the `samples/snippets` tests succeed.

kiraksi:
> LGTM, assuming the `samples/snippets` tests succeed.

They succeeded when I tested before merge

PR #1723, opened 2023-11-15
perf: use the first page a results when `query(api_method="QUERY")`

### LINKED ISSUE #589: Provide support for synchronous queries through the v2/projects/{projectId}/queries endpoint

tswast:

This is a restoration of part of https://github.com/googleapis/python-bigquery/pull/374

TODO:

- [x] Unit tests to verify that no extra `jobs.getQueryResults` or `jobs.get` calls happen when iterating over rows that are fully returned from `jobs.query`.
- [x] Ensure `to_pandas` and `to_arrow` work with this code path.

Closes #589 🦕


google/cloud/bigquery/table.py
```diff
@@ -1582,16 +1586,31 @@ def __init__(
...
         mechanisms.
         """
-        if self._first_page_response is None or self.next_page_token:
+        if self._first_page_response is None:
             return False
 
-        return self._first_page_response.get(self._next_token) is None
+        total_cached_rows = len(self._first_page_response.get(self._items_key, []))
+        if self.max_results is not None and total_cached_rows >= self.max_results:
+            return True
```
tswast, line 1600: Need unit test coverage of this line.

google/cloud/bigquery/table.py
```diff
@@ -1582,16 +1586,31 @@ def __init__(
...
+
+        if (
+            self.next_page_token is None
+            and self._first_page_response.get(self._next_token) is None
+        ):
+            return True
+
+        if self._total_rows is not None:
```
tswast, line 1608: Need unit test for the branch where this is not true.

Linchin:
I'm not sure I understand. Could you explain why doing this will improve query performance, and only applies when `query(api_method="QUERY")`?

tswast:
> Could you explain why doing this will improve query performance, and only applies when query(api_method="QUERY")?

Great question @Linchin.

There are two APIs for issuing a query in BigQuery: [jobs.insert](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/insert) and [jobs.query](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query). There are some key differences between these APIs.

* `jobs.insert`: creates and returns a query job unless it's considered a "bad request" or insufficient permissions or similar. In general this API returns relatively quickly, but only includes job metadata, no query results.

  When running a query this way, we call [jobs.getQueryResults](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) with the page size set to 0 results. This is purely to find out when the job finishes. The reason for 0 results is that for large rows, `getQueryResults` can actually hang well past the expected 10 seconds, causing issues such as https://github.com/googleapis/python-bigquery-pandas/issues/343.

  Once the query has finished, we call [jobs.get](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get) to refresh the job metadata. This ensures we have a destination table, which can be used with [tabledata.list](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list) or the BQ Storage Read API to fetch the results. Many years ago, it used to be much faster to use `tabledata.list` instead of `jobs.getQueryResults` to download data, but that changed in late 2019 when many optimizations were made to `getQueryResults` on the backend. As of https://github.com/googleapis/python-bigquery/pull/363, we use either `jobs.getQueryResults` or the BQ Storage Read API to download results.
  
  In general, this path works best for large query results, which we expect from folks using pandas, for example.
* `jobs.query`: starts the query and waits. If it finishes within about 10 seconds, it returns the results in that same API call. If the query has not finished, we use the job ID from the response to call `jobs.getQueryResults`. The subsequent steps are the same as with `jobs.insert`.

  The key difference here is that `jobs.query` can actually return results in the same request/response as the one to start the query, whereas `jobs.insert` always requires a subsequent API call to fetch the results.

This PR does a few things: (1) if the job has finished, don't make the unnecessary call to `getQueryResults` (2), actually use the rows returned from `jobs.query` in `RowIterator`, and (3) avoid using the BQ Storage Read API if we know only 1 or 2 more API requests are needed to download the rest of the data.

google/cloud/bigquery/_job_helpers.py
```diff
@@ -197,14 +198,9 @@ def _to_query_job(
...
         query_job._properties["status"]["state"] = "DONE"
-        # TODO: https://github.com/googleapis/python-bigquery/issues/589
-        # Set the first page of results if job is "complete" and there is
-        # only 1 page of results. Otherwise, use the existing logic that
-        # refreshes the job stats.
-        #
-        # This also requires updates to `to_dataframe` and the DB API connector
-        # so that they don't try to read from a destination table if all the
-        # results are present.
+        query_job._query_results = google.cloud.bigquery.query._QueryResults(
```
Linchin, line 201: Does this change mean we also load the query results when job is complete, when `query(api_method="INSERT")`?

Linchin (approved):
Thank you for helping me understand this PR :)

google/cloud/bigquery/_job_helpers.py
tswast: Good question. I just double-checked that this method is only called from `query_jobs_query`, so it won't affect when `api_method="INSERT"`

PR #1726, opened 2023-11-16
feat: removed pkg_resources from all test files and moved importlib into pandas extra

### LINKED ISSUE #313: use native support for namespace packages

### LINKED ISSUE #1725: Remove pkg_resource dependency after native namespace compatibility is implemented
Similar to: https://github.com/googleapis/python-bigquery-storage/pull/706
Pkg_resources can be removed from tests after https://github.com/googleapis/python-bigquery/pull/1720 is merged, importlib can be used instead.

kiraksi:
In this PR
- removed all instances of pkg_resources in tests and introduced importlib instead
- similar to https://github.com/googleapis/python-bigquery-storage/pull/706
- should resolve any further issues we had stemming from pkg_resources

Fixes #1725 🦕
Fixes #313 🦕


tests/system/test_pandas.py
```diff
@@ -41,12 +45,12 @@
...
-    PANDAS_INSTALLED_VERSION = pkg_resources.get_distribution("pandas").parsed_version
-else:
-    PANDAS_INSTALLED_VERSION = pkg_resources.parse_version("0.0.0")
+if pandas is not None:  # pragma: NO COVER
+    PANDAS_INSTALLED_VERSION = metadata.version("pandas")
+else:  # pragma: NO COVER
+    PANDAS_INSTALLED_VERSION = "0.0.0"
 
-PANDAS_INT64_VERSION = pkg_resources.parse_version("1.0.0")
+PANDAS_INT64_VERSION = "1.0.0"
```
parthea, line 53:
```suggestion
```


Instead of 
```
@pytest.mark.skipif(
    PANDAS_INSTALLED_VERSION < PANDAS_INT64_VERSION,
    reason="Only `pandas version >=1.0.0` is supported",
)
```
we could have
```
@pytest.mark.skipif(
    PANDAS_INSTALLED_VERSION[0:2] in ["0."],
    reason="Only `pandas version >=1.0.0` is supported",
)
```

tests/unit/test__pandas_helpers.py
```diff
@@ -57,13 +61,12 @@
...
+PANDAS_MINIUM_VERSION = "1.0.0"
 
-if pandas is not None:
-    PANDAS_INSTALLED_VERSION = pkg_resources.get_distribution("pandas").parsed_version
-else:
-    # Set to less than MIN version.
-    PANDAS_INSTALLED_VERSION = pkg_resources.parse_version("0.0.0")
+if pandas is not None:  # pragma: NO COVER
+    PANDAS_INSTALLED_VERSION = metadata.version("pandas")
+else:  # pragma: NO COVER
```
parthea, line 68: Previously we didn't have `# pragma: NO COVER`. Is it needed?

kiraksi: When I ran the kokoro build test it was giving me a failure on coverage in a test file (not this one), so I thought to add the tag everywhere just in case. It may have just been an issue with that test file, so I can remove it here and other places if needed!

parthea (approved):
LGTM. Added one observation for potentially unnecessary version checks in test files.

tests/system/test_pandas.py
```diff
@@ -311,7 +313,7 @@ def test_load_table_from_dataframe_w_automatic_schema(bigquery_client, dataset_i
 
 
 @pytest.mark.skipif(
-    PANDAS_INSTALLED_VERSION < PANDAS_INT64_VERSION,
+    PANDAS_INSTALLED_VERSION[0:2] in ["0."],
     reason="Only `pandas version >=1.0.0` is supported",
```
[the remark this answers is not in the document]
chalmerlowe: I tend to agree, we probably could simplify some of this. I am not sure that should hold up this PR.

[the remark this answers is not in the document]
kiraksi: I can make this change today as well before I planned on merging

tests/system/test_pandas.py
```diff
@@ -1043,9 +1042,7 @@ def test_list_rows_max_results_w_bqstorage(bigquery_client):
     assert len(dataframe.index) == 100
 
 
-@pytest.mark.skipif(
-    PANDAS_INSTALLED_VERSION >= pkg_resources.parse_version("2.0.0"), reason=""
-)
+@pytest.mark.skipif(PANDAS_INSTALLED_VERSION[0:2] not in ["0.", "1."], reason="")
```
chalmerlowe, line 1045: This has several elements in the list so it is a bit easier to grok what is happening here.

chalmerlowe (approved):
LGTM

tests/system/test_pandas.py
```diff
@@ -342,7 +341,7 @@ def test_load_table_from_dataframe_w_nullable_int64_datatype(
 
 
 @pytest.mark.skipif(
-    PANDAS_INSTALLED_VERSION < PANDAS_INT64_VERSION,
+    PANDAS_INSTALLED_VERSION[0:2].startswith("0.")
     reason="Only `pandas version >=1.0.0` is supported",
```
parthea, line 345:
```suggestion
    PANDAS_INSTALLED_VERSION[0:2].startswith("0."),
    reason="Only `pandas version >=1.0.0` is supported",
```

parthea:
Good catch @tswast ! I can produce the issue when using an editable install using egg as it appears in the [build log](https://fusion2.corp.google.com/invocations/2be1e8d5-99f1-41aa-97e4-4cabb94bd441/targets/bigframes%2Fpresubmit%2Fe2e/log) 

When I install using an editable install using egg, the import fails
```
pip install -e git+https://github.com/googleapis/python-bigquery.git@1f4ebb1eca4f9380a31172fc8cb2fae125f8c5a2#egg=google_cloud_bigquery --ignore-installed
```

When I install without an editable install, the import succeeds
```
pip install https://github.com/googleapis/python-bigquery/archive/main.zip --ignore-installed
```

I can produce the same issue with google auth. The import of `google.auth` fails when I use
```
pip install -e git+https://github.com/googleapis/google-auth-library-python.git@7ab0fced0008ad9283c955b7dd7e2b7db55ae5e7#egg=google_auth --ignore-installed`
```

but not with
```
pip install https://github.com/googleapis/google-auth-library-python/archive/main.zip
```

The change to move to native namespace packages was [released in July](https://github.com/googleapis/google-auth-library-python/releases/tag/v2.22.0) for `google-auth`. 

We should investigate whether we want to block on this issue before reverting the change as we do want to support python 3.12 without using `pkg_resources`.  

We could also consider the change to move to native namespace packages as a potential breaking change and bump the major version.

tswast:
Thanks for the investigation! I think we can remove the "-e" in the BigFrames tests.

PR #1722, opened 2023-11-15
feat: add `Client.query_and_wait` which directly returns a `RowIterator` of results

tswast:
Set the `QUERY_PREVIEW_ENABLED=TRUE` environment variable to use this with the new JOB_CREATION_OPTIONAL mode (currently in preview).

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Note: Currently in DRAFT mode as I realized that some of the necessary work to speed things up overlaps with the `query(api_method="QUERY")` work (https://github.com/googleapis/python-bigquery/pull/1723), so I'll work on modifying that first before moving on to the next step of a "jobless" API.

Towards #589 🦕

PR #1733, opened 2023-11-17
feat: add `job_id`, `location`, `project`, and `query_id` properties on `RowIterator`

tswast:
These can be used to recover the original job metadata when `RowIterator` is the result of a `QueryJob`. This will be especially important after https://github.com/googleapis/python-bigquery/pull/1722 where `query_and_wait()` won't return an actual `QueryJob` object.

Towards internal issue 305260214 and related to https://github.com/googleapis/python-bigquery/issues/589
🦕

### DIFF
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -931,6 +931,15 @@ class QueryJob(_AsyncJob):
         )
 
     @property
+    def query_id(self) -> Optional[str]:
+        """[Preview] ID of a completed query.
+
+        This ID is auto-generated and not guaranteed to be populated.
+        """
+        query_results = self._query_results
+        return query_results.query_id if query_results is not None else None
+
+    @property
     def query_parameters(self):
         """See
         :attr:`google.cloud.bigquery.job.QueryJobConfig.query_parameters`.
@@ -1525,7 +1534,12 @@ class QueryJob(_AsyncJob):
                 provided and the job is not retryable.
         """
         if self.dry_run:
-            return _EmptyRowIterator()
+            return _EmptyRowIterator(
+                project=self.project,
+                location=self.location,
+                # Intentionally omit job_id and query_id since this doesn't
+                # actually correspond to a finished query job.
+            )
         try:
             retry_do_query = getattr(self, "_retry_do_query", None)
             if retry_do_query is not None:
@@ -1594,7 +1608,12 @@ class QueryJob(_AsyncJob):
         # indicate success and avoid calling tabledata.list on a table which
         # can't be read (such as a view table).
         if self._query_results.total_rows is None:
-            return _EmptyRowIterator()
+            return _EmptyRowIterator(
+                location=self.location,
+                project=self.project,
+                job_id=self.job_id,
+                query_id=self.query_id,
+            )
 
         rows = self._client._list_rows_from_query_results(
             self.job_id,
@@ -1608,6 +1627,7 @@ class QueryJob(_AsyncJob):
             start_index=start_index,
             retry=retry,
             timeout=timeout,
+            query_id=self.query_id,
         )
         rows._preserve_order = _contains_order_by(self.query)
         return rows
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1590,7 +1590,7 @@ class RowIterator(HTTPIterator):
         self._project = project
 
     @property
-    def _bqstorage_project(self) -> Optional[str]:
+    def _billing_project(self) -> Optional[str]:
         """GCP Project ID where BQ Storage API will bill to (if applicable)."""
         client = self.client
         return client.project if client is not None else None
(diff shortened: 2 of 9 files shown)


google/cloud/bigquery/job/query.py
```diff
@@ -930,6 +930,15 @@ def query(self):
...
         )
 
+    @property
+    def query_id(self) -> Optional[str]:
+        """[Preview] ID of a completed query.
+
+        This ID is auto-generated and not guaranteed to be populated.
+        """
+        query_results = self._query_results
+        return query_results.query_id if query_results is not None else None
```
Linchin, line 940: Is attribute `query_results.query_id` guaranteed to exist? I see that `query_results` is of class `_QueryResults`, but I don't see this attribute in it.

google/cloud/bigquery/table.py
```diff
@@ -1723,7 +1766,7 @@ def to_arrow_iterable(
 
         bqstorage_download = functools.partial(
             _pandas_helpers.download_arrow_bqstorage,
-            self._project,
+            self._bqstorage_project,
```
Linchin, line 1769: Looks like we now have two possible project values, does this mean a `RowIterator` can correspond to different projects for storage APIs compared to other APIs?

google/cloud/bigquery/table.py
```diff
@@ -1575,12 +1579,51 @@ def __init__(
...
         self._first_page_response = first_page_response
+        self._location = location
+        self._job_id = job_id
+        self._query_id = query_id
+        self._project = project
+
+    @property
+    def _bqstorage_project(self) -> Optional[str]:
+        """GCP Project ID where BQ Storage API will bill to (if applicable)."""
+        client = self.client
```
Linchin, line 1595: I wonder what's the difference between `self.client.project` and `self._project`. When the client isn't of storage type, does `self.client.project` still reflect the storage project id?

google/cloud/bigquery/job/query.py
tswast: I just added it in this PR. `google/cloud/bigquery/query.py`, below.

google/cloud/bigquery/table.py
tswast: The difference in other contexts is usually described as "billing project" versus "data project". This is easiest to understand in the context of public datasets. I've renamed this private property to reflect that.

I have permission to query tables in the `bigquery-public-data` project and even download the data in bulk with the BigQuery Storage Read API, but I _don't_ have permission to run queries in that project or start a BigQuery Storage Read API session from that project. Instead, I have start the query or bq storage read session in my own project and reference the tables in the other project.

Linchin: Wow I see! so `self._project` is the project for the data/table, and `self.client.project` a.k.a. billing project is the user's own project. So I guess this billing project doesn't just apply to storage?

google/cloud/bigquery/table.py
```diff
@@ -1575,12 +1579,51 @@ def __init__(
...
         self._total_rows = total_rows
         self._first_page_response = first_page_response
+        self._location = location
+        self._job_id = job_id
+        self._query_id = query_id
+        self._project = project
+
+    @property
+    def _billing_project(self) -> Optional[str]:
+        """GCP Project ID where BQ Storage API will bill to (if applicable)."""
```
Linchin, line 1594:
```suggestion
        """GCP Project ID where BQ API will bill to (if applicable)."""
```

google/cloud/bigquery/table.py
tswast: That's true, though for queries we're using `jobs.getQueryResults`, so the project will always be the billing project since that's where temp results for queries are stored.

tswast: `tabledata.list` is free, but technically we should be overriding the [x-goog-user-project header](https://cloud.google.com/apis/docs/system-parameters#definitions) so that the quota is charged to the billing project. (I don't think we're actually doing that, though)

google/cloud/bigquery/job/query.py
Linchin: Thank you!

Linchin (approved):

tests/unit/test_table.py
```diff
@@ -2113,6 +2113,38 @@ def test_constructor_with_dict_schema(self):
         ]
         self.assertEqual(iterator.schema, expected_schema)
 
+    def test_job_id_missing(self):
+        rows = self._make_one()
+        self.assertIsNone(rows.job_id)
```
chalmerlowe, line 2118: @tswast 

Just curious about testing philosophy.
For this collection of tests that are pretty much the same (i.e. all the tests that .assertIsNone()) is there a benefit in your mind to having each test broken out separately versus providing a parameterized list of inputs and expected outputs?

tswast: I matched the existing pattern in this test. I do think we could probably combine a lot of these into a single parametrized test.

I worry slightly about the required `getattr(property_name)` being too complex for new contributors to understand. In general tests should avoid anything even as complex as an if statement. https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html In this case, `getattr` is a grey area. I wouldn't necessarily count it as "logic"

PR #1741, opened 2023-12-04
chore: cleanup resources at startup

### LINKED ISSUE #590: test: cleanup resources at startup time in system and sample tests
1. Update fixtures to use a consistent prefix, so we can tell which resources are created by `python-bigquery` (and possibly disambiguate between samples and system tests). Also include a timestamp in the prefix.
2. List all resources in the project, filter by our prefix, delete those resources that were created >= 24 hours ago.

This will prevent resource leaks from getting too bad. Teardown doesn't always get a chance to run because of intermittent Kokoro errors and timeouts.

kiraksi:
This PR:
- Uses prefixer function in conftest from python-test-utils to assign IDs to datasets in samples/snippets
- If dataset IDs already exist with the same name, those get deleted at start time
- Primarily, this ensures that any tests that are stopped or crash do not leave dataset resources behind

Fixes #590  🦕

### DIFF
--- a/samples/snippets/authorized_view_tutorial_test.py
+++ b/samples/snippets/authorized_view_tutorial_test.py
@@ -39,12 +39,8 @@ def test_authorized_view_tutorial(
     client: bigquery.Client, datasets_to_delete: List[str]
 ) -> None:
     override_values = {
-        "source_dataset_id": f"{prefixer.create_prefix()}_authorized_view_tutorial".format(
-            str(uuid.uuid4()).replace("-", "_")
-        ),
-        "shared_dataset_id": "shared_views_{}".format(
-            str(uuid.uuid4()).replace("-", "_")
-        ),
+        "source_dataset_id": f"{prefixer.create_prefix()}_authorized_view_tutorial",
+        "shared_dataset_id": f"{prefixer.create_prefix()}_authorized_view_tutorial_shared_views",
     }
     source_dataset_ref = "{}.{}".format(
         client.project, override_values["source_dataset_id"]
--- a/samples/snippets/authorized_view_tutorial_test.py
+++ b/samples/snippets/authorized_view_tutorial_test.py
@@ -22,16 +22,6 @@ import authorized_view_tutorial  # type: ignore
 from conftest import prefixer  # type: ignore
 
 
-@pytest.fixture
-def table_id(
-    bigquery_client: bigquery.Client, project_id: str, dataset_id: str
-) -> Iterator[str]:
-    table_id = f"{prefixer.create_prefix()}_authorized_view_tutorial"
-    yield table_id
-    full_table_id = f"{project_id}.{dataset_id}.{table_id}"
-    bigquery_client.delete_table(full_table_id, not_found_ok=True)
-
-
 @pytest.fixture(scope="module")
 def client() -> bigquery.Client:
     return bigquery.Client()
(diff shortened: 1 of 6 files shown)


samples/snippets/authorized_view_tutorial_test.py
```diff
@@ -19,6 +19,17 @@
 import pytest
 
 import authorized_view_tutorial  # type: ignore
+from conftest import prefixer  # type: ignore
+
+
+@pytest.fixture
+def table_id(
```
chalmerlowe, line 26: We use the function name as a variable name below.

'table_id = f"...'
and
'yield table_id...`

is this intentional?
NOTE: the same issue applies to the function in the `view_test.py` file.

chalmerlowe:
@kiraksi Per our discussion:

* please figure out whether we need individual mechanisms to create prefixes in our tests OR if we can rely upon the mechanisms in conftest to do it globally.

samples/snippets/authorized_view_tutorial_test.py
```diff
@@ -38,7 +39,7 @@ def test_authorized_view_tutorial(
     client: bigquery.Client, datasets_to_delete: List[str]
 ) -> None:
     override_values = {
-        "source_dataset_id": "github_source_data_{}".format(
+        "source_dataset_id": f"{prefixer.create_prefix()}_authorized_view_tutorial".format(
             str(uuid.uuid4()).replace("-", "_")
         ),
```
chalmerlowe, line 44: We should be able to rely upon `prefixer.create_prefix` to create a value for these datasets.
The value will start with something the looks like the `repo` name and the `folder directory` (`self._prefix`).

The cleanup script will look for any datasets that `.startswith()` the value stored as `self._prefix`.

```suggestion
        "source_dataset_id": f"{prefixer.create_prefix()}_authorized_view_tutorial",
```
We don't need to create a universally unique identifier (using `uuid`), since the `prefixer.create_prefix()` function already does that using a function from the `random` library and the `hex()` function. Also `prefixer.create_prefix()` doesn't inadvertantly inject any undesirable separators, so the `replace()` function is not needed.

Note: there are additional places in this code that create dataset_ids, such as the `shared_dataset_id` below. 

Please confirm whether they will also need to be deleted.

samples/snippets/dataset_access_test.py
```diff
@@ -22,6 +22,7 @@
     import pytest
 
 
+# is it possible to cleanup these datasets if they are passed in?
```
chalmerlowe, line 25: There may be a way, while we are examining and testing this process, to use the `capsys` module in pytest to capture outputs to stdout/stderr and examine them (ie. as a list of items that got deleted OR something).

See line 38 below.

https://docs.pytest.org/en/7.1.x/reference/reference.html#capsys

https://docs.pytest.org/en/7.1.x/_modules/_pytest/capture.html

samples/snippets/natality_tutorial_test.py
```diff
@@ -38,7 +39,7 @@ def test_natality_tutorial(
     client: bigquery.Client, datasets_to_delete: List[str]
 ) -> None:
     override_values = {
-        "dataset_id": "natality_regression_{}".format(
+        "dataset_id": f"{prefixer.create_prefix()}_natality_tutorial".format(
```
chalmerlowe, line 42: Revise this in the same manner as above.
Also revise any other dataset_id generators in the file.

samples/snippets/dataset_access_test.py
kiraksi: Will be removing this comment as I am prioritizing other workloads

chalmerlowe (approved):
LGTM

Issue #1724, opened 2023-11-16
bigquery Python: support for RANGE type


shollyman:
This issue tracks the work to enable support for the RANGE datatype in the BigQuery python client libraries.

Steps:
- [x] Schema Support: #1746
- [x] StandardSQL Data Type #1807
- [x] Query Parameterization #1827
- [x] Reading Support (already works)
- [x] Write API Support https://github.com/googleapis/python-bigquery-storage/pull/780
- [x] InsertAll Support #1909

PR #1746, opened 2023-12-07
feat: support RANGE in schema

Linchin:
part of #1724 
- add class `FieldElementType` to support `RANGE` type in schema
- add related unit tests


Linchin:
test failures seem to be api call timeout.

shollyman (approved):
Probably worth augmenting a system test to create a table with a RANGE type in the schema, but LGTM otherwise.

Linchin:
@shollyman After adding the system test, I also had to make a bunch of other changes. Please take another look, thank you :)

google/cloud/bigquery/schema.py
```diff
@@ -66,6 +66,46 @@ class _DefaultSentinel(enum.Enum):
...
+    Args:
+        element_type (str): The type of a field element.
+    """
+
+    def __init__(self, element_type: str):
+        self._properties = {}
+        self._properties["type"] = element_type.upper()
+
+    @property
+    def element_type(self):
```
shollyman, line 81: Do you also need a setter for this property?  Or you're only exposing a setter from the containing type?

google/cloud/bigquery/schema.py
```diff
@@ -252,6 +311,18 @@ def max_length(self):
...
+    def range_element_type(self):
+        """Optional[FieldElementType]: The subtype of the RANGE, if the
+        type of this field is RANGE.
+
+        Must be set when ``type`` is `"RANGE"`. Must be one of `"DATE"`,
+        `"DATETIME"` or `"TIMESTAMP"`.
+        """
+        if self._properties.get("rangeElementType"):
+            ret = self._properties.get("rangeElementType")
+            return FieldElementType(element_type=ret["type"])
```
shollyman, line 324: should this be something like `FieldElementType.from_api_repr(ret)`?  If the type gets more fields over time you'll possibly lose fidelity since you're only grabbing a single field.

tests/system/test_client.py
```diff
@@ -1804,6 +1799,11 @@ def test_dbapi_fetch_w_bqstorage_client_large_result_set(self):
                 ("id", 3),
                 ("timestamp", datetime.datetime(2006, 10, 9, 18, 40, 33, tzinfo=UTC)),
             ],
+            [
```
shollyman, line 1802: Not clear how the changes in this file relate to the feature.  Just a side fixup?

Linchin: The expected value was changed because the public table somehow lost the first row of the table (`id=1`) around 12/31, so the system test has been failing.

google/cloud/bigquery/schema.py
Linchin: Good point! I'll change this

google/cloud/bigquery/schema.py
Linchin: I decided to not allow the element type to be altered in order to protect the table schema from being changed unintentionally. As a reference, most of the properties in class `SchemaField` don't have setters either.

google/cloud/bigquery/schema.py
Linchin: done

shollyman (approved):

PR #1748, opened 2023-12-08
fix: keep `RowIterator.total_rows` populated after iteration

### LINKED ISSUE #589: Provide support for synchronous queries through the v2/projects/{projectId}/queries endpoint

tswast:
Split out of https://github.com/googleapis/python-bigquery/pull/1722

This was being reset in some cases when the rows were all available in the first page of results.

Fixes #589  🦕

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -2997,7 +2997,7 @@ def _rows_page_start(iterator, page, response):
     page._columns = _row_iterator_page_columns(iterator._schema, response)
 
     total_rows = response.get("totalRows")
-    # Don't reset total_rows is it's not present in the next API response.
+    # Don't reset total_rows if it's not present in the next API response.
     if total_rows is not None:
         iterator._total_rows = int(total_rows)
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -2997,9 +2997,9 @@ def _rows_page_start(iterator, page, response):
     page._columns = _row_iterator_page_columns(iterator._schema, response)
 
     total_rows = response.get("totalRows")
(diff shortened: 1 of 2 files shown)


tswast:
Alternatively, we could populate `totalRows` here if present:

https://github.com/googleapis/python-bigquery/blob/ab737969c000683ac52c499ee16ac76bd4cbdc84/google/cloud/bigquery/table.py#L1703

Not sure which to prefer, but this way involved updating fewer tests.

google/cloud/bigquery/table.py
```diff
@@ -2997,9 +2997,9 @@ def _rows_page_start(iterator, page, response):
     page._columns = _row_iterator_page_columns(iterator._schema, response)
 
     total_rows = response.get("totalRows")
+    # Don't reset total_rows is it's not present in the next API response.
```
parthea, line 3000:
```suggestion
    # Don't reset total_rows if it's not present in the next API response.
```

parthea (approved):

PR #1751, opened 2023-12-14
feat: Support jsonExtension in LoadJobConfig

### LINKED ISSUE #1581: Support jsonExtension in LoadJobConfig
This would allow running the equivalent to:
```
bq load --source_format=NEWLINE_DELIMITED_JSON --json_extension=GEOJSON  <some>.<table>  <some_geojson_path>
```

kiraksi:
This PR:
- Adds jsonExtension as a parameter to LoadJobConfig when loading data from a specified file format
- It enable users to directly specify the extension within the job configuration and store that information in metadata
- It allows BQ to Interpret the source format correctly (?)

- [] Appropriate docs were updated (if necessary) (should autogenerate?)

Fixes #1581 🦕

### DIFF
--- a/google/cloud/bigquery/job/load.py
+++ b/google/cloud/bigquery/job/load.py
@@ -328,6 +328,23 @@ class LoadJobConfig(_JobConfig):
         self._set_sub_prop("ignoreUnknownValues", value)
 
     @property
+    def json_extension(self):
+        """Optional[str]: The extension to use for writing JSON data to BigQuery.
+
+        For example, if you set `jsonExtension` to "nested", BigQuery
+        writes the loaded data as nested rows, one row per top-level
+        JSON object.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/JobConfigurationLoad.FIELDS.json_extension
+
+        """
+        return self._get_sub_prop("jsonExtension")
+
+    @json_extension.setter
+    def json_extension(self, value):
+        self._set_sub_prop("jsonExtension", value)
+    
+    @property
     def max_bad_records(self):
         """Optional[int]: Number of invalid rows to ignore.
 
--- a/google/cloud/bigquery/job/load.py
+++ b/google/cloud/bigquery/job/load.py
@@ -343,7 +343,7 @@ class LoadJobConfig(_JobConfig):
     @json_extension.setter
     def json_extension(self, value):
         self._set_sub_prop("jsonExtension", value)
-    
+
     @property
     def max_bad_records(self):
         """Optional[int]: Number of invalid rows to ignore.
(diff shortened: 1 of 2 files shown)


kiraksi:
We are unable to find jsonExtension explicitly in BigQuery documentation, only some mentions of it. This PR is on hold until we can get clarity on this.

tests/unit/job/test_load_config.py
```diff
@@ -413,6 +413,45 @@ def test_ignore_unknown_values_setter(self):
         config.ignore_unknown_values = True
         self.assertTrue(config._properties["load"]["ignoreUnknownValues"])
 
+    def test_json_extension_missing(self):
+        config = self._get_target_class()()
+        self.assertIsNone(config.json_extension)
+
+    def test_json_extension_hit(self):
```
chalmerlowe, line 420: What is the intended difference between `test_*_hit` and `test_*_setter`

They go through very similar machinations.

kiraksi: From my understanding, they are testing different behaviors, although I was following the format of the other tests, test_*_hit checks if the property is properly initialized or retrieved and test_*_set verifies that the setter correctly updates the internal state and reflects the new value, so its more testing the behavior of setting the property rather than testing the intial or retrieved value.

google/cloud/bigquery/job/load.py
```diff
@@ -327,6 +327,23 @@ def ignore_unknown_values(self):
...
 
+    @property
+    def json_extension(self):
+        """Optional[str]: The extension to use for writing JSON data to BigQuery.
+
+        For example, if you set `jsonExtension` to "nested", BigQuery
+        writes the loaded data as nested rows, one row per top-level
+        JSON object.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/JobConfigurationLoad.FIELDS.json_extension
```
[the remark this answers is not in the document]
kiraksi: After talking to Seth, it was noted that this property was created for this field, however the person working on it left Google before they could remove the documentation restriction. Seth has lifted the restriction and now the new link accurately reflects where you can find the documentation on jsonExtension.

tests/unit/job/test_load_config.py
chalmerlowe: I asked a little bit about this issue and one of the responses was pretty interesting and describes an interesting philosophy (i paraphrased it a bit and formatted it):

"Incidentally, ... I see it omitted in tests all the time: there should be an 

`assertFalse(config.key1)` then `config._properties["key1"] = True` and then the `assertTrue(config.key1)`

to make sure it actually registers a change in state."

chalmerlowe (approved):
LGTM

PR #1762, opened 2024-01-08
fix: add detailed message in job error

Linchin:
Temp fix of internal issue b/310544564.

Starting Oct 2023, customers have found that the error message with jobs have lacked exact cause info. This seems to have been caused by detailed errors being moved to a different part of the job message. This PR makes a temp fix by combining both error sources into the exception we raise.

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -83,7 +83,7 @@ def _error_result_to_exception(error_result, errors=None):
         for err in errors:
             for key, value in err.items():
                 string += f"{key}: {value}, "
-        string = string[:-2]
+        string = string[:-2]  # strips off the last unneeded comma and space
 
     error_message = error_result.get("message", "") + string
 
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -82,7 +82,7 @@ def _error_result_to_exception(error_result, errors=None):
         string = "; "
         for err in errors:
             for key, value in err.items():
-                string += key + ": " + str(value) + ", "
+                string += f"{key}: {value}, "
         string = string[:-2]
 
     error_message = error_result.get("message", "") + string
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -55,7 +55,7 @@ _ERROR_REASON_TO_EXCEPTION = {
 }
 
 
-def _error_result_to_exception(error_result):
+def _error_result_to_exception(error_result, errors=None):
     """Maps BigQuery error reasons to an exception.
 
     The reasons and their matching HTTP status codes are documented on
@@ -66,6 +66,7 @@ def _error_result_to_exception(error_result):
 
     Args:
         error_result (Mapping[str, str]): The error result from BigQuery.
+        errors (Union[Iterable[str], None]): The detailed error messages.
 
     Returns:
         google.cloud.exceptions.GoogleAPICallError: The mapped exception.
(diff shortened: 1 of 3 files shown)


samples/snippets/authenticate_service_account.py
```diff
@@ -24,7 +24,7 @@ def main() -> "bigquery.Client":
 
     # [START bigquery_client_json_credentials]
     from google.cloud import bigquery
-    from google.oauth2 import service_account
+    from google.oauth2 import service_account  # type: ignore
```
chalmerlowe, line 27: @Linchin 

Thanks for trying to enable the work to continue moving forward.
I saw an error in mypy in my PR last night with this line of code.

I worry about this as a "fix" for the problem.
That line of code used to work and doesn't now... not only in this repo, but others.

I spoke with Anthonios last night. It appears that there is an issue with the google-auth library that causes the mypy error. A change was recently made in that library which broke some things downstream.

I am gonna open a separate PR to set our requirements for this repo to avoid the problematic version of google-auth and provide a comment as to why that version is being avoided.

I am hoping that will fix the issue in a slightly more robust way than simply ignoring the error.

chalmerlowe: PR 1764 was approved. I merged that code into this branch. Hopefully that should clear up the mypy-samples error and we can delete the `type: ignore` comment.

Linchin: Thank you for explaining the issue in detail! It was indeed strange, because there wasn't much change in the imported module `google.oauth2`. I'll remove the ignore here.

google/cloud/bigquery/job/base.py
```diff
@@ -74,8 +75,20 @@ def _error_result_to_exception(error_result):
     status_code = _ERROR_REASON_TO_EXCEPTION.get(
         reason, http.client.INTERNAL_SERVER_ERROR
     )
+    # Manually create error message to preserve both error_result and errors.
+    # Can be removed once b/310544564 and b/318889899 are resolved.
+    string = ""
```
chalmerlowe, line 80: `string` is a [Python module](https://docs.python.org/3/library/string.html) in the Standard library. I recommend that we rename this variable for clarity's sake.

perhaps `concatenated_errors`.

Please ensure that the remaining instances of `string` are similarly renamed.

```suggestion
    concatenated_errors = ""
```

google/cloud/bigquery/job/base.py
```diff
@@ -74,8 +75,20 @@ def _error_result_to_exception(error_result):
...
+    if errors:
+        string = "; "
+        for err in errors:
+            for key, value in err.items():
+                string += key + ": " + str(value) + ", "
```
chalmerlowe, line 85: Using an `f-string` here seems like it will simplify this line of code.
Ensure that the variable name matches the above.

```suggestion
                string += f"{key}: {value}, "
```

google/cloud/bigquery/job/base.py
```diff
@@ -74,8 +75,20 @@ def _error_result_to_exception(error_result):
...
+        string = string[:-2]
```
chalmerlowe, line 86: I feel like a comment here quickly makes it obvious why we are using a slice here.

```suggestion
        string = string[:-2]  # strips off the last unneeded comma and space
```

chalmerlowe (changes requested):
See some comments in the code and a thought:

I recommend that we add a test to ensure that this code works as desired AND to help ensure that in the future if we modify this code, that we don't break anything.

google/cloud/bigquery/job/base.py
Linchin: Good point, I made the change. I will also add a unit test.

chalmerlowe:
I made a few changes/suggestions via the web interface to Github and my personal email address got associated with them so that is why the CLA error is showing up.
Sigh.

Linchin:
It happened to me last time too! I just overrode it :)

chalmerlowe (approved):
LGTM

PR #1770, opened 2024-01-11
docs: update the query with no cache sample to use `query_and_wait` API

SalemJorden:

### DIFF
--- a/samples/query_no_cache.py
+++ b/samples/query_no_cache.py
@@ -26,8 +26,8 @@ def query_no_cache() -> None:
         FROM `bigquery-public-data.samples.shakespeare`
         GROUP BY corpus;
     """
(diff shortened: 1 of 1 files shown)


samples/query_no_cache.py
```diff
@@ -26,7 +26,7 @@ def query_no_cache() -> None:
         FROM `bigquery-public-data.samples.shakespeare`
         GROUP BY corpus;
     """
-    query_job = client.query(sql, job_config=job_config)  # Make an API request.
+    query_job = client.query_and_wait(sql, job_config=job_config)  # Make an API request.
```
tswast, line 29: Please rename this variable to `results` since it is no longer a QueryJob object.

tswast (approved):

PR #1771, opened 2024-01-11
feature: add query location for bigquery magic

### LINKED ISSUE #1760: Ability to specify a location for the BigQuery magics %%bigquery requests
When using the BigQuery magics command, I am unable to specify a location where the query will be executed.  When we look at the Cloud Console, we find that we can go into query settings > Advanced  and set the location type (after unchecking "Automatic location selection").  By naming a region/multi region we can explicitly declare where the query should run.

Prior to now (2024-01) the ability to specify where the query should execute was likely not that interesting ... however, now we have the concept of Cross Region Dataset replication.  This allows me to define a dataset and specify a named region where my dataset will be asynchronously replicated to.  Effectively, I now have the dataset existing in MULTIPLE regions and, in that case, it makes sense to specify where I want my query to run.

I reviewed the magics code and *believe* the change would be relatively simple.  Looking at the code I find

```
client = bigqueryClient(...)
```

and then reviewing [here](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client) ... I see that location is a parameter of the Client() constructor.  I get the feel that adding location as a parameter to Bigquery magics would not be too onerous.

I could imagine:
```
%%bigquery --location us-east1
SELECT * FROM X.Y;
```

Gaurang033:
Fixes #1760

### DIFF
--- a/google/cloud/bigquery/magics/magics.py
+++ b/google/cloud/bigquery/magics/magics.py
@@ -508,6 +508,15 @@ def _create_dataset_if_necessary(client, dataset_id):
         "Defaults to use tqdm_notebook. Install the ``tqdm`` package to use this feature."
     ),
 )
+@magic_arguments.argument(
+    "--location",
+    type=str,
+    default=None,
+    help=(
+        "Set the location to execute query."
+        "Defaults to location set in query setting in console."
+    ),
+)
 def _cell_magic(line, query):
     """Underlying function for bigquery cell magic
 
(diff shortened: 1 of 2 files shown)


Linchin:
lgtm, thank you for your contribution! Cheers.

Linchin (approved):

wsmolak:
After the merge this [doc](https://googleapis.dev/python/bigquery-magics/latest/magics.html#module-bigquery_magics.bigquery) is outdated. `--location` parameter works like a charm, but you can't find it in the doc:(

PR #1786, opened 2024-01-17
docs: Updates query to query and wait in `client_query_w_positional_params.py`

DevStephanie:


samples/client_query_w_positional_params.py
```diff
@@ -35,7 +35,9 @@ def client_query_w_positional_params() -> None:
             bigquery.ScalarQueryParameter(None, "INT64", 250),
         ]
     )
-    query_job = client.query(query, job_config=job_config)  # Make an API request.
+    query_job = client.query_and_wait(
```
tswast, line 38: Please rename this variable, as the return value is no longer a QueryJob object. Instead it is a RowIterator object. "rows" or "results" could be a good name.

tswast (changes requested):

samples/client_query_w_positional_params.py
DevStephanie: Yes, sounds good. Will correct that now.

samples/client_query_w_positional_params.py
```diff
@@ -35,7 +35,9 @@ def client_query_w_positional_params() -> None:
             bigquery.ScalarQueryParameter(None, "INT64", 250),
         ]
     )
-    query_job = client.query(query, job_config=job_config)  # Make an API request.
+    results = client.query_and_wait(
+        query, job_config=job_config
+    )  # Make an API request.
 
     for row in query_job:
```
tswast, line 42: This line needs to be updated too. The `query_job` variable doesn't exist anymore.

DevStephanie: Yes, will correct now.

tswast (approved):
Thanks!

PR #1787, opened 2024-01-17
docs: Updates `query` to `query and wait` in samples/desktopapp/user_credentials.py

DevStephanie:


kiraksi:
Looks like you forgot to remove query_job.result() here as I believe query_and_wait returns the result itself, as the cause of your sample errors. Also Tim mentioned in my PR that he wanted to query_job to be renamed to result as well which may also be helpful to you here!

tswast (approved):
Thanks!

Issue #1398, opened 2022-11-03
Add support for default_value_expression in SchemaField


JoaoPaes-at-Dynamox:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

 **Is your feature request related to a problem? Please describe.**
I'm developing a wrapper library to use `pydantic.BaseModel` as means to define the `TableSchema`. As part of this wrapper, I would like to provide "server default" values, such as `CURRENT_DATETIME`, `CURRENT_TIMESTAMP`, so on, to `datetime.datetime` fields;

 **Describe the solution you'd like**

```python
client = bigquery.Client("someproject")

bq_table_schema_field = SchemaField(..., default_value_expression="CURRENT_DATETIME()")

table_with_default_value = Table("my_table", schema=[schema_field])

client.create_table(table_with_default_value)
```

The field `"defaultValueExpression"` is a valid [REST API field](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema) and would be a direct mapping from the initializer argument `default_value_expression` to the API `"defaultValueExpression"` field.

 **Describe alternatives you've considered**

The alternative to this, would be construct a DDL statement for creating tables which columns can have default values via `DEFAULT` keyword

```python
base_create = f"CREATE TABLE {somedataset_name}.{table_with_default_value.name} ("
for field in fields:
    field_as_column_statement = f'{field.name} {as_bq_type(field.type)} {"DEFAULT" if field.has_server_default else ""} {field.server_default_statement if field.has_server_default else ""}, '
    base_create = base_create + field_as_column_statement
base_create = base_create + ");"
```

Which is very cumbersome and ugly, and useless, since the REST API already provides a field for default value expressions.

JoaoPaes-at-Dynamox:
Actually, a very much simpler, but extremely bad practice, way of solving this is

```python
...

bq_table_schema_field = SchemaField(...)
bq_table_schema_field._properties["defaultValueExpression"] =  "CURRENT_DATETIME()"

...
```

Which is what I'm currently using.

JoaoPaes-at-Dynamox:
Just figured out that you included the parameter within the `3.4` release. I'm closing this.

tswast:
Closed by https://github.com/googleapis/python-bigquery/pull/1408

PR #1793, opened 2024-01-19
fix: `query_and_wait` now retains unknown query configuration `_properties`

### LINKED ISSUE #1792: bug: `query_and_wait` drops unknown properties in the QueryJobConfig
In implementing pandas-gbq integration with the new, faster `query_and_wait` method, a test failure raised the issue that some request body details in a custom job configuration are being dropped. See: https://github.com/googleapis/python-bigquery-pandas/pull/722/files/7739f41989c5d0effbaf66070c6b94b6c3840506#r1459296431

#### Steps to reproduce

1. Create a `QueryJobConfig` with unknown or invalid parameters via `from_api_repr()`
2. Pass the `job_config` to `query_and_wait()`.
3. Observe that the extra parameters are not sent.

#### Code example

```python
import google.cloud.bigquery
client = google.cloud.bigquery.Client()
project_id = "swast-scratch"

job_config = google.cloud.bigquery.QueryJobConfig.from_api_repr(
        {
            "copy": {
                "sourceTable": {
                    "projectId": project_id,
                    "datasetId": "publicdata:samples",
                    "tableId": "wikipedia",
                },
                "destinationTable": {
                    "projectId": project_id,
                    "datasetId": "publicdata:samples",
                    "tableId": "wikipedia_copied",
                },
            }
        })
client.query_and_wait("select 1", job_config=job_config)
```

#### Stack trace

No stack trace is produced, even though one should be due to an invalid configuration.

tswast:
fix: raise `ValueError` in `query_and_wait` with wrong `job_config` type

Fixes #1792
🦕


Linchin:
Could you tell me more about the rationale of keeping unknown config fields?

tswast:
> Could you tell me more about the rationale of keeping unknown config fields?

Two reasons: (1) as you can see from the removal of this code:

```
    # These keys are top level in job resource and query resource.
    if "labels" in job_config_resource:
        request_body["labels"] = job_config_resource["labels"]
    if "dryRun" in job_config_resource:
        request_body["dryRun"] = job_config_resource["dryRun"]
```

it allows for a cleaner conversion from query job insert request to jobs.query request.

(2) It allows for customers who want access to features that are in preview an not yet implemented in the client library to set REST API properties via setting `QueryJobConfig._properties`. This is a very common practice (see: https://github.com/googleapis/python-bigquery/issues/1398#issuecomment-1324147401 for example) and is in fact a requirement for features that are in a "private" release. We don't support adding private features to our public client library, but this `_properties` feature allows such customers to use the public client library via this workaround.

Linchin:
That's great to know! Thank you

Linchin (approved):

PR #1796, opened 2024-01-23
fix: add pyarrow.large_string to the _ARROW_SCALAR_IDS_TO_BQ map

### LINKED ISSUE #1795: The _ARROW_SCALAR_IDS_TO_BQ mapping misses LargeStringArray type
#### Environment details

  - OS type and version: Linux
  - Python version: 3.11.7
  - pip version: 23.3.1
  - `google-cloud-bigquery` version: 3.16.0

#### Steps to reproduce

Call `bqclient.load_table_from_dataframe` with a dataframe containing a string type. Before pandas 2.2.0, the `pyarrow.array` would detect the type as `pyarrow.lib.StringArray`. After switching to pandas `2.2.0`, the `pyarrow.lib.LargeStringArray` would be returned. But it misses the BQ type mapping.

#### Stack trace

callstack

The left results are in `pandas 2.2.0` and the right result are from `pandas 2.1.3`

chelsea-lin:
- [X] Make sure to open an issue as a [bug](https://github.com/googleapis/python-bigquery/issues/1795) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #1795🦕

### DIFF
--- a/google/cloud/bigquery/_pyarrow_helpers.py
+++ b/google/cloud/bigquery/_pyarrow_helpers.py
@@ -91,6 +91,7 @@ if pyarrow:
         pyarrow.date64().id: "DATETIME",  # because millisecond resolution
         pyarrow.binary().id: "BYTES",
         pyarrow.string().id: "STRING",  # also alias for pyarrow.utf8()
+        pyarrow.large_string().id: "STRING",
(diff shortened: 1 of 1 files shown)


tswast (approved):
Thanks!

PR #1794, opened 2024-01-19
fix: retry 'job exceeded rate limits' for DDL queries

kiraksi:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #1790 🦕

PR #1797, opened 2024-01-24
fix: avoid unnecessary call to jobs.get in case of retriable query failure

tswast:

Fixes issue identified here: https://github.com/googleapis/python-bigquery/pull/1794/files#r1465230014
 🦕

### DIFF
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -1554,6 +1554,8 @@ class QueryJob(_AsyncJob):
                     )
 
             first = True
+            if self.state is None:
+                self._begin(retry=retry, timeout=timeout)
 
             def do_get_result():
                 nonlocal first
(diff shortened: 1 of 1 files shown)


tswast:
I was just reminded on https://github.com/googleapis/python-bigquery/issues/1122 today. We need to be careful with this change. The `jobs.getQueryResults()` API can fail in ambiguous ways where the underlying job may or may not have failed. Might actually be best to revert this change.

tswast:
An example of where doing jobs.getQueryResults can lead to ambiguity is the rateLimitExceeded error. With jobs.getQueryResults, if we get such an error, we don't really know if it's because of the general API-level rate limit or if it's because we hit the max number of concurrent queries and need to retry the whole query with a new job ID.

By doing jobs.get, which doesn't raise on query job failures, we remove this ambiguity. A rateLimitExceeded on jobs.get will be that API-level error and never mean the query job failed for that reason.

Issue #1563, opened 2023-04-29
BigQuery seems to automatically convert STRING to BYTES if STRING > 186 bytes


cgeudens:
#### Environment details

  - OS type and version: macOS Ventura 13.3.1
  - Python version: 3.9.12
  - pip version: pip 23.1.1
  - `google-cloud-bigquery` version: 3.8.0, 3.9.0, 3.10.0

#### Steps to reproduce

See the code sample below. Loading a string > 186 bytes to a STRING column raises the following exception: `google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table. Field longstring has changed type from STRING to BYTES`

#### Code example

```python
import json
from google.cloud import bigquery
from google.cloud.bigquery.schema import SchemaField
from google.oauth2.service_account import Credentials
from pocs.poc_settings import settings

service_acc = json.loads(settings.GCP_BQ_SERVICE_ACC)
credentials = Credentials.from_service_account_info(info=service_acc)
bq_client = bigquery.Client(credentials=credentials)

def add_data(user_id):

    dataset_ref = bigquery.DatasetReference(project="myproject", dataset_id="mydataset")
    table_ref = bigquery.TableReference(dataset_ref=dataset_ref, table_id=user_id)

    json_rows = [
        {"longstring": "a" * 138}   # 187 bytes
    ]

    job_config = bigquery.LoadJobConfig(
        autodetect=False,
        create_disposition="CREATE_NEVER",
        max_bad_records=0,
        write_disposition="WRITE_APPEND"
    )

    job = bq_client.load_table_from_json(
        json_rows=json_rows,
        destination=table_ref,
        num_retries=2,
        job_config=job_config
    )

    job.result()

def create_table(user_id):

    dataset_ref = bigquery.DatasetReference(project="myproject", dataset_id="mydataset")
    table_ref = bigquery.TableReference(dataset_ref=dataset_ref, table_id=user_id)
    schema = [
        SchemaField(name="longstring", field_type="STRING")
    ]
    table = bigquery.Table(table_ref=table_ref, schema=schema)
    table = bq_client.create_table(table)

create_table("test")
add_data("test")
```

#### Stack trace
```
google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table. Field longstring has changed type from STRING to BYTES
```

[...]

PR #1801, opened 2024-01-25
fix: no longer set autodetect=True with empty schema in load_table_from_json()

### LINKED ISSUE #1228: load_table_from_json interpolates string as int
#### Issue

using `load_table_from_json()` to load data into a string column fails when using a value that can be interpolated as an int (eg. "123")

see code snippet, which shows inconsistency with the other data loading apis

#### Environment details

  - OS type and version: macos 12.2.1
  - Python version: 3.9.5 (also errors on 3.8)
  - pip version: `pip --version`
  - `google-cloud-bigquery` version: 2.31.0 & 3.0.1

#### Code example

```python
import io
import json

from google.api_core.exceptions import BadRequest
from google.cloud.bigquery import Client, Table, SchemaField

data = [{"col_1": "123"}]

project = 'my-project'
bq = Client(project)

table = Table(f'{project}.aaa.data_load_test', schema=[SchemaField("col_1", "STRING"), ])
bq.delete_table(table, not_found_ok=True)
bq.create_table(table)


# WORKS: bq load from byte stream
with io.StringIO() as fp:
    json.dump(data, fp)
    fp.seek(0)
    bq.load_table_from_file(fp, table).result()


# WORKS: streaming api
bq.insert_rows_json(table, data)


try:
    # FAILS: bq load from python dict.
    # 123 is interpolated as an int, even though i supply it as a string
    bq.load_table_from_json(data, table).result()
except BadRequest as e:
    print("ERROR")
    print(e)


# WORKS: bq load from python dict, but 123d cannot be interpolated as an int
bq.load_table_from_json([{"col_1": "123d"}], table).result()

```

#### Stack trace

```
  File "/Users/ben.marengo/code/scratch/bq_load.py", line 28, in <module>
    bq.load_table_from_json(data, table).result()
  File "/Users/ben.marengo/code/pyenvs/scratch/lib/python3.9/site-packages/google/cloud/bigquery/job/base.py", line 728, in result
    return super(_AsyncJob, self).result(timeout=timeout, **kwargs)
  File "/Users/ben.marengo/code/pyenvs/scratch/lib/python3.9/site-packages/google/api_core/future/polling.py", line 137, in result
    raise self._exception
google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table my-project:aaa.data_load_test. Field col_1 has changed type from STRING to INTEGER
```

Linchin:
In the past, if we don't set `autodetect=True` when schema is empty, the backend will throw an error. This has changed now, backend will autodetect even if we don't send autodetect value. This has caused some issues with users, when the detected schema doesn't match that of the existing table. (#1228, #1563) 

For more discussions, see #1228.

Fixes #1228 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2833,9 +2833,6 @@ class Client(ClientWithProject):
 
         new_job_config.source_format = job.SourceFormat.NEWLINE_DELIMITED_JSON
 
-        if new_job_config.schema is None:
-            new_job_config.autodetect = True
-
         if project is None:
             project = self.project
 
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -8986,7 +8986,7 @@ class TestClientUpload(object):
         sent_config = load_table_from_file.mock_calls[0][2]["job_config"]
         assert sent_config.source_format == job.SourceFormat.NEWLINE_DELIMITED_JSON
         assert sent_config.schema is None
-        assert sent_config.autodetect
+        assert sent_config.autodetect is None
 
     def test_load_table_from_json_non_default_args(self):
         from google.cloud.bigquery import job


google/cloud/bigquery/client.py
```diff
@@ -2833,9 +2833,6 @@ def load_table_from_json(
 
         new_job_config.source_format = job.SourceFormat.NEWLINE_DELIMITED_JSON
 
-        if new_job_config.schema is None:
```
tswast, line 2836: We are still getting test failures with this change:

```

>           raise exceptions.from_http_response(exc.response)
E           google.api_core.exceptions.BadRequest: 400 POST https://bigquery.googleapis.com/upload/bigquery/v2/projects/precise-truck-742/jobs?uploadType=multipart: No schema specified on job or table.
```

I think we can only omit these if the table already exists and we are appending rows.

Linchin: Thanks, I think I made a mistake in my test code and created the table with schema in it.

Linchin:
use #1804 instead

PR #1802, opened 2024-01-26
feat: make jobs awaitable with asyncio, v1 (Proof of Concept)

### LINKED ISSUE #18: BigQuery: make jobs awaitable

kiraksi:
This PR:
- Is a proof of concept test for the eventual AsyncQueryManager
- Should not be merged and will be closed after testing
- Successfully showed that async functionality works cohesively with our base code

Fixes #18 🦕

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -33,8 +33,6 @@ import pytest
 import sys
 import inspect
 
-from unittest.mock import patch
-
 try:
     import importlib.metadata as metadata
 except ImportError:
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -30,6 +30,7 @@ import mock
 import requests
 import packaging
 import pytest
+import sys
 
 try:
     import importlib.metadata as metadata
@@ -45,6 +46,10 @@ try:
     import opentelemetry
 except ImportError:
     opentelemetry = None
+
+if sys.version_info >= (3, 9):
+    import asyncio
+    from unittest.mock import patch, AsyncMock
 
 if opentelemetry is not None:
     try:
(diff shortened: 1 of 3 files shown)


tests/unit/test_client.py
```diff
@@ -9546,6 +9551,37 @@ def test_schema_to_json_with_file_object(self):
         client.schema_to_json(schema_list, fake_file)
         assert file_content == json.loads(fake_file.getvalue())
 
+    if sys.version_info >= (3, 9):
+
+        @patch("bigquery_client.async_query_and_wait", new_callable=AsyncMock)
+        async def test_async_execution(self, mock_query):
```
tswast, line 9557: Are we sure this test is detected and runnable by pytest? From https://stackoverflow.com/a/70016047/101923 it seems we may need to install another plugin for pytest and add a decorator.

kiraksi: I did run the test without the if statement, it understandably caused errors for 3.7/3.8 for not being supported however passed for other python versions. It seems like the proof of concept works, but I will revisit how to run this test according to this example in the next commit!

kiraksi: Because I am using asyncio.to_thread, should I test if it returns a coroutine before the await, instead of a future?

kiraksi:
The tests failing for this is for pytype, which is unimportant for the purpose of this PoC PR. Closing this as a successful demonstration of the viability of asynchrony in our library.

PR #1807, opened 2024-01-30
feat: support range sql

Linchin:
Part of #1724.

Supports User Defined Functions with RANGE type parameters.


google/cloud/bigquery/enums.py
```diff
@@ -292,6 +293,7 @@ class SqlTypeNames(str, enum.Enum):
     TIME = "TIME"
     DATETIME = "DATETIME"
     INTERVAL = "INTERVAL"  # NOTE: not available in legacy types
+    RANGE = "RANGE"
```
shollyman, line 296: is the comment for INTERVAL about legacy availability also true for RANGE?

shollyman (approved):

google/cloud/bigquery/enums.py
Linchin: I guess the legacy types here refer to the [types supported by legacy SQL](https://cloud.google.com/bigquery/docs/data-types#legacy_sql_data_types)? In that case, it is indeed not supported. If so, I will add the comment here.

PR #1809, opened 2024-02-06
fix: add google-auth as a direct dependency

parthea:
`google.auth` is used in this repository and should be added as a direct dependency so that the minimum version can be stated and tested
https://github.com/search?q=repo%3Agoogleapis%2Fpython-bigquery%20google.auth&type=code

https://github.com/googleapis/python-bigquery/blob/62490325f64e5d66303d9218992e28ac5f21cb3f/setup.py#L31-L44

BEGIN_COMMIT_OVERRIDE
fix(deps): Require google-api-core>=1.34.1, >=2.11.0
fix: add google-auth as a direct dependency
END_COMMIT_OVERRIDE


chalmerlowe:
@parthea 
Getting a dependency issue. From the traceback:

```
The conflict is caused by:
    google-cloud-bigquery 3.17.2 depends on google-auth<3.0.0dev and >=2.14.1
    google-api-core 1.31.5 depends on google-auth<2.0dev and >=1.25.0
    The user requested (constraint) google-auth==2.14.1
```

parthea:
I think the minimum version of google-api-core should be bumped to `1.34.1` to match the version that we use for the generated clients.

https://github.com/googleapis/google-cloud-python/blob/24f085185f0c90cbb71fb68a3bdf45fac1a5fe5a/packages/google-cloud-access-approval/setup.py#L44

chalmerlowe:
Most recent conflict errors that need resolution:
```
google-cloud-bigquery 3.17.2 depends on google-auth<3.0.0dev and >=2.14.1
google-api-core[grpc] 1.34.1 depends on google-auth<3.0dev and >=1.25.0
google-cloud-core 1.6.0 depends on google-auth<2.0dev and >=1.24.0
The user requested (constraint) google-auth==2.14.1
```

kiraksi:
@chalmerlowe This seems to be just a problem in python 3.7, which I believe we plan to deprecate for this library anyway no?

chalmerlowe (approved):
LGTM

PR #1823, opened 2024-02-21
chore: use mock from unittest

parthea:
Replace `mock` with `unittest.mock` to reduce the number of test dependencies


noxfile.py
```diff
@@ -89,6 +88,8 @@ def default(session, install_extras=True):
         install_target = "."
     session.install("-e", install_target, "-c", constraints_path)
 
+    session.run("python", "-m", "pip", "freeze")
```
kiraksi, line 91: Small comment, but would you want to continue to pip freeze in the default session going forward, or was this mostly for testing the changes in this PR?

kiraksi (commented):
LGTM with one comment

noxfile.py
chalmerlowe: Theoretically, I want this in every session, cause one of the questions I ask most frequently when looking at errors is: "what version caused this situation". But I did not wanna clutter up this PR with a bunch of tweaks to all the sessions unnecessarily.

kiraksi (approved):
LGTM

testing/constraints-3.7.txt
```diff
@@ -7,11 +7,16 @@
 # Then this file should have foo==1.14.0
 db-dtypes==0.3.0
 geopandas==0.9.0
-google-api-core==1.31.5
-google-cloud-bigquery-storage==2.6.0
-google-cloud-core==1.6.0
-google-resumable-media==0.6.0
+google-api-core==2.17.1
```
tswast, line 10: Looks like we updated these without updating the corresponding minimum versions in `setup.py`. I'll try to revert as much as I can. Do we know why we needed to upgrade these?

PR #1824, opened 2024-02-21
fix: throw exception for data type mismatch for load_table_from_dataf…

### LINKED ISSUE #1822: Raised `ArrowTypeError` should indicates which column is causing the error
Debugging typing errors when working with existing tables with a large number of columns is extremely frustrating, to say the least: the `pyarrow` error never indicates which column is causing the issue.

It would be an absolute life saver if more debugging information were raised with the exception.

E.g.:
```
pyarrow.lib.ArrowTypeError: object of type <class 'str'> cannot be converted to int
```

Gaurang033:
Fixes #1822

### DIFF
--- a/google/cloud/bigquery/_pandas_helpers.py
+++ b/google/cloud/bigquery/_pandas_helpers.py
@@ -302,11 +302,15 @@ def bq_to_arrow_array(series, bq_field):
 
     field_type_upper = bq_field.field_type.upper() if bq_field.field_type else ""
 
-    if bq_field.mode.upper() == "REPEATED":
-        return pyarrow.ListArray.from_pandas(series, type=arrow_type)
-    if field_type_upper in schema._STRUCT_TYPES:
-        return pyarrow.StructArray.from_pandas(series, type=arrow_type)
-    return pyarrow.Array.from_pandas(series, type=arrow_type)
+    try:
+        if bq_field.mode.upper() == "REPEATED":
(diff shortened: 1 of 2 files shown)


chalmerlowe:
Thanks for jumping on this so quickly.
I suspect that since we are adding a new flow path through the code, we will encounter a coverage issue because that path is not tested.

We will likely need at least one new test to test this path.

Gaurang033:
> Thanks for jumping on this so quickly. I suspect that since we are adding a new flow path through the code, we will encounter a coverage issue because that path is not tested.
> 
> We will likely need at least one new test to test this path.

I can add a unit test for this if you like

Gaurang033:
Unit test was failing as exception thrown in python 3.7 and 3.8 and higher version was different.  changed the assertion to use regex and tested locally with both the version. Tests should pass now.

chalmerlowe:
@Gaurang033 
I had multiple issues and PRs come together at the same time including this PR.

I took various parts of each of them and have merged them into a single PR.
This PR will be closed as soon as the tests and review are complete on PR #1836 

So please don't devote any additional time to this PR.
If you have comments, thoughts, ideas or suggestions for #1836, feel free to add them there. Thanks so much for getting this ball rolling and offering your time and help. It is much appreciated.

chalmerlowe:
Closed by the merger of https://github.com/googleapis/python-bigquery/pull/1836

PR #1827, opened 2024-02-23
feat: support RANGE query parameters

Linchin:
Part of #1724. 

In this PR, we add support for using a query parameter of RANGE type, either named or positional. Here is an example below:
```
from datetime import datetime, date

# Construct a BigQuery client object.
client = bigquery.Client()

query = """SELECT ? AS foo"""


# Set the name to None to use positional parameters.
# Note that you cannot mix named and positional parameters.
job_config = bigquery.QueryJobConfig(
    query_parameters=[
        bigquery.RangeQueryParameter("DATETIME", start=datetime(year=2020, month=10, day=30), end=None),
    ]
)

results = client.query_and_wait(query, job_config=job_config) 

for row in results:
    print(row)
``` 
More specifically:
- Added classes `RangeQueryParameter`, `RangeQueryParameterType` and their corresponding methods
- Added unit tests
- Added system tests

Towards #1724 🦕


chalmerlowe:
@Linchin please fill out the description of this PR so that reviewers and "future us" can quickly identify the purpose and scope of this change. Thanks.

Linchin:
Thanks @chalmerlowe! Definitely got lost in the work and forgot to add the description. I just edited it to add more details :)

tests/unit/test_query.py
```diff
@@ -663,6 +757,460 @@ def test___repr__(self):
...
+            "parameterValue": {
+                "rangeValue": {"start": {"value": None}, "end": {"value": None}}
+            },
+        }
+        param = self._make_one(range_element_type="DATE", name="foo")
+        self.assertEqual(param.to_api_repr(), EXPECTED)
+
+    def test_to_api_repr_unsupported_value_type(self):
+        with self.assertRaisesRegex(
+            ValueError, "Cannot covert range element value from type"
```
shollyman, line 1129: s/covert/convert/ here as well

tests/system/test_query.py
```diff
@@ -422,6 +423,38 @@ def test_query_statistics(bigquery_client, query_api_method):
                 )
             ],
         ),
+        (
+            "SELECT @range_date",
+            "[2016-12-05, UNBOUNDED)",
```
shollyman, line 428: Is 'UNBOUNDED' some kind of type alias, or does the service return that as the literal string?  If so, should we convert UNBOUNDED -> None when build query parameter values?

Linchin: The service returns 'UNBOUNDED' as part of the literal string. 'UNBOUNDED' isn't an alias of any type, so it's safe to convert it into `None`.

shollyman: Thanks!  Its fine to punt on this aspect, I just wanted to clarify.

shollyman (approved):

google/cloud/bigquery/query.py
```diff
@@ -811,6 +936,178 @@ def __repr__(self):
         return "StructQueryParameter{}".format(self._key())
 
 
+class RangeQueryParameter(_AbstractQueryParameter):
+    """Named / positional query parameters for range values.
+
+    Args:
+        range_element_type (Union[str, RangeQueryParameterType]):
+            The type of range elements. If given as a string, it must be one of
```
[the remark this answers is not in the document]
Linchin: Thank you, I just removed this line whenever `RangeQueryParameterType` may be either a string or `ScalarQueryParameterType`.

google/cloud/bigquery/query.py
```diff
@@ -811,6 +936,178 @@ def __repr__(self):
...
+    def _serialize_range_element_value(self, value, type_):
+        if value is None or isinstance(value, str):
+            return value
+        else:
+            converter = _SCALAR_VALUE_TO_JSON_PARAM.get(type_)
+            if converter is not None:
+                return converter(value)  # type: ignore
+            else:
+                raise ValueError(
+                    f"Cannot covert range element value from type {type_}, "
```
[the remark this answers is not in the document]
Linchin: done

tests/unit/test_query.py
Linchin: done

PR #1829, opened 2024-02-25
fix: keyword rendering and docstring improvements

### LINKED ISSUE #1662: keyword rendering in hosted docs
#### Steps to reproduce

  1. Open [this page](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_query): 
  2. Check the content of the `query` param, you can see that it not only includes the query param, but all other kwargs arguments too. I don't think this formatting is supposed to look like it. [Docs in the code](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/client.py#L3274) look fine

I couldn't really figure out which technology is used for the doc rendering, so I wasn't able to propose a change myself.
But looks like the used markdown doesn't work in this case. This might require an update of the docstrings.

kiraksi:
This PR:
- Removes "Keyword Arguments" section and adds arguments to "Arguments" to fix rendering on public docs
- Adds types to docstring arguments missing it for clarity

Fixes #1662 🦕


chalmerlowe (approved):
LGTM

PR #1837, opened 2024-02-29
fix: augment universe_domain handling

shollyman:
This PR revisits the universe resolution for the BQ client, and handles
new requirements like env-based specification and validation.


google/cloud/bigquery/_helpers.py
```diff
@@ -55,9 +57,63 @@
...
+        client_options: specified client options.
+    Returns:
+        str: resolved universe setting.
+
+    """
+    if isinstance(client_options, dict):
+        client_options = client_options_lib.from_dict(client_options)
+    universe = _DEFAULT_UNIVERSE
+    if hasattr(client_options, "universe_domain"):
+        options_universe = getattr(client_options, "universe_domain")
```
Linchin, line 85: Do we need to account for the case when `options_universe == ""`?

google/cloud/bigquery/client.py
```diff
@@ -253,16 +257,18 @@ def __init__(
...
+                if client_universe != _DEFAULT_UNIVERSE:
+                    kw_args["api_endpoint"] = _DEFAULT_HOST_TEMPLATE.replace(
+                        "{UNIVERSE_DOMAIN}", client_universe
+                    )
+        # Ensure credentials and universe are not in conflict.
+        if hasattr(self, "_credentials") and client_universe is not None:
+            _validate_universe(client_universe, self._credentials)
 
         self._connection = Connection(self, **kw_args)
+
```
Linchin, line 271: nit: should we keep the blank line here?

tests/unit/helpers.py
```diff
@@ -43,6 +43,20 @@ def make_client(project="PROJECT", **kw):
     return google.cloud.bigquery.client.Client(project, credentials, **kw)
 
 
+def make_creds(creds_universe: None):
+    from google.auth import credentials
+
+    class TestingCreds(credentials.Credentials):
+        def refresh(self, request):  # pragma: NO COVER
```
Linchin, line 50: For my education, what is the purpose of the `refresh()` method here?

Linchin:
Thank you @shollyman! Overall it looks good, despite a new nitpicks.

tests/unit/helpers.py
shollyman: This subclasses credentials.  Refresh is part of the mechanism to provide bearer credentials, but is marked abstract in the base class so we need to provide an impl even though we're not using it here.

google/cloud/bigquery/client.py
shollyman: removed

google/cloud/bigquery/_helpers.py
shollyman: good point, added conditionals and tests.

tests/unit/test__helpers.py
```diff
@@ -17,8 +17,74 @@
...
+
+        self.assertEqual("googleapis.com", _get_client_universe(None))
+
+    def test_with_dict(self):
+        from google.cloud.bigquery._helpers import _get_client_universe
+
+        options = {"universe_domain": "foo.com"}
+        self.assertEqual("foo.com", _get_client_universe(options))
+
+    def test_with_clientoptions(self):
```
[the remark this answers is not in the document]
shollyman: done

Linchin (approved):
LGTM. Thank you!

PR #1839, opened 2024-03-04
feat: AsyncClient for async query_and_wait for BQ jobs

### LINKED ISSUE #18: BigQuery: make jobs awaitable

kiraksi:
This PR:
- Instantiates the an asynchronous client specifically for an asynchronous AsyncClient.query_and_wait method, including asynchronous job helpers
- Makes a test suite for asynchronous method
- Benchmarks this method against Client.query_and_wait and Client.query

TO DO:
- [] Pass all tests
- [] Create samples for this method for user understanding of how to use AsyncClient
- [] Create separate PRs for other async RPC methods

Fixes #18  🦕


kiraksi:
#1809 needs to be merged as aiohttp is a extra of google-auth which needs to be added as a direct dependency. Also see: https://github.com/googleapis/google-auth-library-python/issues/613 for google-auth exposure of transport.aiohttp_requests

chalmerlowe:
**Migration Notice:** This library is moving to the [google-cloud-python](https://github.com/googleapis/google-cloud-python) monorepo soon.

We closed this PR due to inactivity to ensure a clean migration. Please re-open this work in the new monorepo once the migration is complete!

PR #1844, opened 2024-03-06
fix: supplementary fix to env-based universe resolution

shollyman:
There's a corner case where conversion from dict to a ClientOptions will return a universe_domain value as None that wasn't covered by initial testing.  This updates the resolution code and adds tests to exercise the new path.


Linchin:
LGTM, thank you!

Linchin (approved):

PR #1848, opened 2024-03-08
fix: correct type checking

### LINKED ISSUE #1838: Bug introduced in v3.12.0
Hi,

I think a bug was introduced in v3.12.0, it this commit: https://github.com/googleapis/python-bigquery/commit/3645e32aeebefe9d5a4bc71a6513942741f0f196#diff-80c3339e876a6651c796df9b6621e6c8fba7d8b168f65239753398a0441b06a9R1345

`isinstance(someinteger, object)` called like that is always True, because `int` is an instance of an `object`:
```
>>> isinstance(3, object)
True
```
As a result, the timeout is always changed to `None` at this point. This causes timeout to not be respected when for example `QueryJob.exception` method is called.

Linchin:
Correct the way we check whether `self._done_timeout` is an instance of `object` class or not.

Fixes #1838 🦕

### DIFF
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -1411,7 +1411,7 @@ class QueryJob(_AsyncJob):
         # Our system does not natively handle that and instead expects
         # either none or a numeric value. If passed a Python object, convert to
         # None.
-        if isinstance(self._done_timeout, object):  # pragma: NO COVER
+        if type(self._done_timeout) is object:  # pragma: NO COVER
             self._done_timeout = None
 
         if self._done_timeout is not None:  # pragma: NO COVER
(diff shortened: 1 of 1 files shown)


google/cloud/bigquery/job/query.py
```diff
@@ -1411,7 +1411,7 @@ def _reload_query_results(
         # Our system does not natively handle that and instead expects
         # either none or a numeric value. If passed a Python object, convert to
```
chalmerlowe, line 1412:
```suggestion
        # either None or a numeric value. If passed a Python object, convert to
```

chalmerlowe (changes requested):
LGTM

chalmerlowe (approved):
LGTM

PR #1851, opened 2024-03-11
testing: unhook prerelease-deps-3.12 from presubmit

shollyman:
Testing for prerelease-deps is done within continuous.


chalmerlowe (approved):
LGTM

PR #1864, opened 2024-03-20
fix: updates a number of optional dependencies

chalmerlowe:
This fix updates a number of optional dependencies.
We use a different module import process (pytest.importorskip versus unittest.skipif).

This first major commit gets the ball rolling, there are gonna be a few additional commits to cover other files.

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -593,7 +593,7 @@ class Client(ClientWithProject):
             )
             return None
 
-        if bqstorage_client is None: # pragma: NO COVER
+        if bqstorage_client is None:  # pragma: NO COVER
             bqstorage_client = bigquery_storage.BigQueryReadClient(
                 credentials=self._credentials,
                 client_options=client_options,
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -593,7 +593,7 @@ class Client(ClientWithProject):
             )
             return None
 
-        if bqstorage_client is None:
+        if bqstorage_client is None: # pragma: NO COVER
             bqstorage_client = bigquery_storage.BigQueryReadClient(
                 credentials=self._credentials,
                 client_options=client_options,
--- a/tests/system/test_client.py
+++ b/tests/system/test_client.py
@@ -1791,9 +1791,8 @@ class TestBigQuery(unittest.TestCase):
         self.assertEqual(row_tuples, [(5, "foo"), (6, "bar"), (7, "baz")])
 
     def test_dbapi_fetch_w_bqstorage_client_large_result_set(self):
-        pytest.importorskip("google.cloud.bigquery_storage")
+        bigquery_storage = pytest.importorskip("google.cloud.bigquery_storage")
         pytest.importorskip("pyarrow")
-        from google.cloud import bigquery_storage
 
         bqstorage_client = bigquery_storage.BigQueryReadClient(
             credentials=Config.CLIENT._credentials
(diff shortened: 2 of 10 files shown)


google/cloud/bigquery/client.py
```diff
@@ -593,7 +593,7 @@ def _ensure_bqstorage_client(
             )
             return None
 
-        if bqstorage_client is None:
+        if bqstorage_client is None:  # pragma: NO COVER
```
tswast, line 596: Looks like this branch isn't really needed anymore due to `raise_if_error=True` above.

tests/system/test_client.py
```diff
@@ -1800,11 +1790,11 @@ def test_dbapi_fetchall_from_script(self):
         row_tuples = [r.values() for r in rows]
         self.assertEqual(row_tuples, [(5, "foo"), (6, "bar"), (7, "baz")])
 
-    @unittest.skipIf(
-        bigquery_storage is None, "Requires `google-cloud-bigquery-storage`"
-    )
-    @unittest.skipIf(pyarrow is None, "Requires `pyarrow`")
     def test_dbapi_fetch_w_bqstorage_client_large_result_set(self):
+        pytest.importorskip("google-cloud-bigquery-storage")
```
tswast, line 1794: importorskip needs to be the import name not the package name

```suggestion
        bigquery_storage = pytest.importorskip("google.cloud.bigquery_storage")
```

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip

tests/unit/test_table.py
```diff
@@ -4282,7 +4281,7 @@ def test_to_dataframe_w_bqstorage_empty_streams(self):
         self.assertTrue(got.empty)
 
     def test_to_dataframe_w_bqstorage_nonempty(self):
-        pytest.importorskip("google-cloud-bigquery-storage")
+        pytest.importorskip("google.cloud.bigquery-storage")
```
tswast, line 4284:
```suggestion
        pytest.importorskip("google.cloud.bigquery_storage")
```

google/cloud/bigquery/client.py
chalmerlowe: @tswast 
Maybe I don't understand the logic here.
If `bigquery_storage` can't be imported, we raise one of two errors and we return from the `_ensure_bqstorage_client()` function with `None`.

But if we successfully import `bigquery_storage` AND if `bqstorage_client` is `None`, we then create a new `bqstorage_client`.

What am I missing, why would we remove that branch?

tswast: If we get to this line, we know bigquery_storage was imported

tswast: So we don't need the if statement

tswast: Ah yes, you're right. I misread.

tests/unit/conftest.py
```diff
@@ -39,6 +39,11 @@ def LOCATION():
     yield "us-central"
 
 
+@pytest.fixture
+def PYARROW_MINIMUM_VERSION():  # pragma: NO COVER
+    return "3.0.0"
+
+
```
tswast, line 46: Doesn't look like this fixture is used?

Also if it is used, could this use the constant from `_version_helpers` as the tests do below?

tests/unit/conftest.py
```diff
@@ -39,6 +39,11 @@ def LOCATION():
     yield "us-central"
 
 
+@pytest.fixture
+def PYARROW_MINIMUM_VERSION():  # pragma: NO COVER
+    return "3.0.0"
```
chalmerlowe, line 44:
```suggestion
```

tswast (approved):

PR #1868, opened 2024-03-23
feat: support RANGE in queries Part 2: Arrow

Linchin:
This PR supports getting RANGE values in queries as Arrow. This is the second part of the RANGE queries PR, and should not be merged until part 1 is merged. Part 1 #1884 supports RANGE as JSON.

Part of #1724🦕


tests/unit/test_table.py
```diff
@@ -3503,7 +3503,10 @@ def test_to_dataframe_no_tqdm_no_progress_bar(self):
         user_warnings = [
             warning for warning in warned if warning.category is UserWarning
         ]
-        self.assertEqual(len(user_warnings), 0)
+        # Note: number of warnings is inconsistent across python versions
```
Linchin, line 3506: I removed the assertion for number of warnings, because we get different number of them with different python versions. Same for line 3540

tests/unit/test_table.py
```diff
@@ -3534,7 +3537,10 @@ def test_to_dataframe_no_tqdm(self):
         user_warnings = [
             warning for warning in warned if warning.category is UserWarning
         ]
-        self.assertEqual(len(user_warnings), 1)
+        # Note: number of warnings is inconsistent across python versions
```
Linchin, line 3540: I removed the assertion for number of warnings, because we get different number of them with different python versions.

google/cloud/bigquery/_helpers.py
```diff
@@ -309,6 +309,46 @@ def _json_from_json(value, field):
         return None
 
 
+def _range_element_from_json(value, field):
+    """Coerce 'value' to a range element value, if set or not nullable."""
+    if value == "UNBOUNDED":
+        return None
+    elif field.element_type == "DATE":
```
shollyman, line 316: Do we have a list of the supported range element types somewhere?  If so, this could like be simplified to something like
```
if field.element_type IN _SUPPORTED_RANGE_ELEMENTS:
  return _CELLDATA[field.element_type](value, None)
else:
  raise ValueError...
```

google/cloud/bigquery/_pyarrow_helpers.py
```diff
@@ -68,6 +90,9 @@ def pyarrow_timestamp():
         "STRING": pyarrow.string,
         "TIME": pyarrow_time,
         "TIMESTAMP": pyarrow_timestamp,
+        "RANGE<TIMESTAMP>": pyarrow_range_timestamp,
```
shollyman, line 93: addmittedly I'm not super familiar with this code, but it seems like we're ending up declaring N scalar versions of range.  Do we get a simpler story from treating range more like a complex type and branching our struct processing logic?

google/cloud/bigquery/_helpers.py
Linchin: This way looks so much neater! I updated the code.

google/cloud/bigquery/_pyarrow_helpers.py
Linchin: Indeed. I changed the code and opted to add a new conversion function, parallel to scalar and struct. I chose this instead of using the existing function for struct, because RANGE has fixed field names, so the logic is simpler.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -142,6 +142,12 @@ def bq_to_arrow_struct_data_type(field):
     return pyarrow.struct(arrow_fields)
 
 
+def bq_to_arrow_range_data_type(field):
+    element_type = field.element_type.upper()
+    arrow_element_type = _pyarrow_helpers.bq_to_arrow_scalars(element_type)()
```
shollyman, line 147: do we need to do validation here?  None-check?

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -274,6 +286,22 @@ def types_mapper(arrow_data_type):
         elif time_dtype is not None and pyarrow.types.is_time(arrow_data_type):
             return time_dtype
 
+        elif pyarrow.types.is_struct(arrow_data_type):
```
shollyman, line 289: Do we need to handle structs more generally here, or is that logic elsewhere?

google/cloud/bigquery/table.py
```diff
@@ -2215,6 +2279,69 @@ def to_dataframe(
...
+                    pyarrow.struct(
+                        [("start", pyarrow.date32()), ("end", pyarrow.date32())]
+                    )
+                )
+            except AttributeError:
+                # pandas.ArrowDtype was introduced in pandas 1.5, but python 3.7
+                # only supports upto pandas 1.3. If pandas.ArrowDtype is not
+                # present, we raise a warning and set range_date_dtype to None.
+                msg = (
+                    "Unable ro find class ArrowDtype in pandas, setting "
```
shollyman, line 2294: s/ro/to/ here and in the other two msgs below

google/cloud/bigquery_v2/types/model.py
```diff
@@ -103,6 +103,7 @@ class Model(proto.Message):
 
     class ModelType(proto.Enum):
         r"""Indicates the type of the Model."""
+
```
shollyman, line 106: these formatting changes are benign, but it's not clear why they're in this PR.  Maybe pull out unrelated cleanups into a separate PR?

tests/unit/test_table.py
```diff
@@ -3497,13 +3497,9 @@ def test_to_dataframe_no_tqdm_no_progress_bar(self):
         api_request = mock.Mock(return_value={"rows": rows})
         row_iterator = self._make_one(_mock_client(), api_request, path, schema)
 
-        with warnings.catch_warnings(record=True) as warned:
+        with warnings.catch_warnings(record=True):
             df = row_iterator.to_dataframe(create_bqstorage_client=False)
 
-        user_warnings = [
```
shollyman, line 3503: Not familiar with these tests, but are we losing some existing signal here?  Or did test expectations change somehow?  Its not clear how this related to the PR intent.

shollyman (commented):
Thanks for putting this together. Mostly minor comments and questions.

google/cloud/bigquery_v2/types/model.py
Linchin: Indeed, I will revert these changes

google/cloud/bigquery/table.py
Linchin: Thanks for catching this!

google/cloud/bigquery/_pandas_helpers.py
Linchin: Great point, I will add a None-check here

tests/unit/test_table.py
Linchin: Indeed, we are losing some signals here - we no longer validate the number of warnings here. With this PR we will have different length of warnings [depending on the version of pandas](https://github.com/Linchin/python-bigquery/blob/get-query-results-range/google/cloud/bigquery/table.py#L2301-L2343). I deleted the warnings check, because I didn't want to have pandas version hard-coded in our tests (Should I be concerned about this?). Another alternative I can think of is to check `self.assertTrue(len(user_warnings) in {length_with_older_pandas, length_with_newer_pandas})`- this way we will lose some signal but less. What do you think?

google/cloud/bigquery/_pandas_helpers.py
Linchin: I added it, as well as the unit tests in `test__pandas_helpers.py`.

google/cloud/bigquery/_pandas_helpers.py
Linchin: Good question! Indeed, our types mapper function doesn't seem to do any conversion for STRUCT or ARRAY. This function is used as the parameter `types_mapper` to [pyarrow's `Table.to_pandas()`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas), allowing for customizable type mapping from pyarrow to pandas. I'm not entirely sure why we didn't provide struct/array mapping options, but I think it might be related to the fact that the fields of a struct can be of any type, so the conversion can become quite complicated.

google/cloud/bigquery/_helpers.py
```diff
@@ -310,17 +312,13 @@ def _json_from_json(value, field):
...
-        return _date_from_json(value, None)
-    elif field.element_type == "DATETIME":
-        return _datetime_from_json(value, None)
-    elif field.element_type == "TIMESTAMP":
-        return _timestamp_from_json(value, None)
+    if field.element_type in _SUPPORTED_RANGE_ELEMENTS:
+        return _CELLDATA_FROM_JSON[field.element_type](value, field.element_type)
     else:
-        raise ValueError(f"Unsupported range field type: {value}")
+        raise ValueError(f"Unsupported range field type: {field.element_type}")
```
shollyman, line 321: nit: should we change this to indicate the element type is unsupported, rather than "field type"?

tests/unit/test_table.py
shollyman: If all the supported versions emit at least one UserWarning we could still do an assert on the count being nonzero, but if we don't think this is a useful validation I'm fine removing it as well.

shollyman (approved):
Left some followup comments and a very minor nit, otherwise LGTM.  Thanks!

google/cloud/bigquery/_helpers.py
Linchin: Indeed, I'll change it to be consistent with the name of the field.

Linchin:
> If all the supported versions emit at least one UserWarning we could still do an assert on the count being nonzero, but if we don't think this is a useful validation I'm fine removing it as well.

Indeed, I added back the test, and it will pass when `len(warnings) == 0` or `len(warnings) == 3`, so we can cover both cases. I added a comment that explains why and when each value should be true. (Same for another warnings number check).

PR #1881, opened 2024-03-29
fix: creates linting-typing.cfg in presubmit

chalmerlowe:
This PR creates a new configuration file in the `.kokoro/presubmit` folder.
The intent is to eventually move multiple checks out of the baseline `Kokoro` check and put them into their own `Kokoro linting-typing`:

The checks we are moving out include:

#### Typing:
* mypy
* mypy_samples
* pytype
#### Linting/formatting:
* lint
* lint_setup_py
* blacken

### DIFF
--- a/.kokoro/presubmit/linting-typing.cfg
+++ b/.kokoro/presubmit/linting-typing.cfg
@@ -7,5 +7,5 @@ env_vars: {
 }
 env_vars: {
     key: "RUN_LINTING_TYPING_TESTS"
-    value: "true"
+    value: "false"
 }
--- a/.kokoro/presubmit/linting-typing.cfg
+++ b/.kokoro/presubmit/linting-typing.cfg
@@ -1,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run these nox sessions.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "lint lint_setup_py blacken mypy mypy_samples pytype"
+}
--- a/.kokoro/presubmit/linting-typing.cfg
+++ b/.kokoro/presubmit/linting-typing.cfg
@@ -5,7 +5,3 @@ env_vars: {
     key: "NOX_SESSION"
     value: "lint lint_setup_py blacken mypy mypy_samples pytype"
 }
-env_vars: {
-    key: "RUN_LINTING_TYPING_TESTS"
-    value: "false"
-}
--- a/.kokoro/presubmit/linting-typing.cfg
+++ b/.kokoro/presubmit/linting-typing.cfg
@@ -5,3 +5,7 @@ env_vars: {
     key: "NOX_SESSION"
     value: "lint lint_setup_py blacken mypy mypy_samples pytype"
 }
+env_vars: {
+    key: "RUN_LINTING_TYPING_TESTS"
+    value: "true"
+}
--- a/.kokoro/presubmit/presubmit.cfg
+++ b/.kokoro/presubmit/presubmit.cfg
@@ -9,3 +9,12 @@ env_vars: {
     key: "RUN_SNIPPETS_TESTS"
     value: "false"
 }
+env_vars: {
+    key: "RUN_TYPING_TESTS"
+    value: "false"
+}
+env_vars: {
+    key: "RUN_LINTING_TESTS"
+    value: "false"
+}
+
--- a/.kokoro/presubmit/presubmit.cfg
+++ b/.kokoro/presubmit/presubmit.cfg
@@ -11,5 +11,5 @@ env_vars: {
 }
 env_vars: {
     key: "RUN_LINTING_TYPING_TESTS"
-    value: "false"
+    value: "test_false"
 }
--- a/.kokoro/presubmit/presubmit.cfg
+++ b/.kokoro/presubmit/presubmit.cfg
@@ -10,10 +10,6 @@ env_vars: {
     value: "false"
 }
 env_vars: {
-    key: "RUN_TYPING_TESTS"
-    value: "false"
-}
-env_vars: {
-    key: "RUN_LINTING_TESTS"
+    key: "RUN_LINTING_TYPING_TESTS"
     value: "false"
 }
--- a/.kokoro/presubmit/presubmit.cfg
+++ b/.kokoro/presubmit/presubmit.cfg
@@ -11,5 +11,5 @@ env_vars: {
 }
 env_vars: {
     key: "RUN_LINTING_TYPING_TESTS"
-    value: "test_false"
+    value: "false"
 }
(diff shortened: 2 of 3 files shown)


chalmerlowe:
There is an internal Google CL to update the allowed env vars for this repository.
Once that gets approved, we should be able to confirm that this PR runs and determine what, if any time-savings are achieved by parallelizing these tests.

.kokoro/presubmit/presubmit.cfg
```diff
@@ -9,3 +9,11 @@ env_vars: {
     key: "RUN_SNIPPETS_TESTS"
     value: "false"
 }
+env_vars: {
+    key: "RUN_TYPING_TESTS"
```
shollyman, line 13: This PR proposes we effectively double our RUN_* boolean knobs from 2 to 4, which to me feels we have an opportunity to improve our approach further.  I like that NOX_SESSION allows us to be explicit with a list of sessions rather than a single session.

For the general case, should we consolidate the control knobs to something more expressive like some kind of session group filtering syntax?  I think we're mostly concerned about the cases where NOX_SESSION isn't set.

If we maintain a map of filtername -> list of sessions, we could potential skip sessions if a single filter variable is set.

chalmerlowe: Not gonna lie:

I have very little idea how to this style of config files (both here and on google3) are supposed to be formatted and how they interact.

I spent more time than I should in just trying to get to this point.

I agree whole heartedly that I want to have better knobs.

The RUN_* knobs are part of a bigger/different picture:
* As we discussed in a meeting, we create the virtual envs using a nox.session decorator and then we come back and check to see if we even need to run the session at all by checking these knobs, which wastes time making the virtual_envs needlessly

But in the interest of getting something that is functional, keeping this PR small and clean and can prove that we can add a new parallel kokoro job, I chose to aim for an approach that we already use and that works, knowing that per our CI/CD redesign, I was gonna come back and figure out whether we can improve the knob situation holistically.

I am not sure what you mean by these statements: I am not picturing your end state... can you elaborate?

> For the general case, should we consolidate the control knobs to something more expressive like some kind of session group filtering syntax? I think we're mostly concerned about the cases where NOX_SESSION isn't set.

> If we maintain a map of filtername -> list of sessions, we could potential skip sessions if a single filter variable is set.

shollyman (approved):
Thanks for simplifying this a bit!

noxfile.py
```diff
@@ -132,6 +132,10 @@ def unit_noextras(session):
 def mypy(session):
     """Run type checks with mypy."""
 
+    # Check the value of `RUN_TYPING_TESTS` env var. It defaults to true.
```
chalmerlowe, line 135:
```suggestion
    # Check the value of `RUN_LINTING_TYPING_TESTS` env var. It defaults to true.
```

noxfile.py
```diff
@@ -153,6 +157,10 @@ def pytype(session):
     # recent version avoids the error until a possibly better fix is found.
     # https://github.com/googleapis/python-bigquery/issues/655
 
+    # Check the value of `RUN_TYPING_TESTS` env var. It defaults to true.
```
chalmerlowe, line 160:
```suggestion
    # Check the value of `RUN_LINTING_TYPING_TESTS` env var. It defaults to true.
```

noxfile.py
```diff
@@ -213,6 +221,10 @@ def system(session):
 def mypy_samples(session):
     """Run type checks with mypy."""
 
+    # Check the value of `RUN_TYPING_TESTS` env var. It defaults to true.
```
chalmerlowe, line 224:
```suggestion
    # Check the value of `RUN_LINTING_TYPING_TESTS` env var. It defaults to true.
```

noxfile.py
```diff
@@ -394,6 +406,10 @@ def lint(session):
     serious code quality issues.
     """
 
+    # Check the value of `RUN_LINTING_TESTS` env var. It defaults to true.
```
chalmerlowe, line 409:
```suggestion
    # Check the value of `RUN_LINTING_TYPING_TESTS` env var. It defaults to true.
```

noxfile.py
```diff
@@ -408,6 +424,10 @@ def lint(session):
 def lint_setup_py(session):
     """Verify that setup.py is valid (including RST check)."""
 
+    # Check the value of `RUN_LINTING_TESTS` env var. It defaults to true.
```
chalmerlowe, line 427:
```suggestion
    # Check the value of `RUN_LINTING_TYPING_TESTS` env var. It defaults to true.
```

noxfile.py
```diff
@@ -418,6 +438,10 @@ def blacken(session):
     Format code to uniform standard.
     """
 
+    # Check the value of `RUN_LINTING_TESTS` env var. It defaults to true.
```
chalmerlowe, line 441:
```suggestion
    # Check the value of `RUN_LINTING_TYPING_TESTS` env var. It defaults to true.
```

chalmerlowe (commented):
updated the comments.

.kokoro/presubmit/linting-typing.cfg
```diff
@@ -0,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run these nox sessions.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "lint lint_setup_py blacken mypy mypy_samples pytype"
+}
```
chalmerlowe, line 7:
```suggestion
}
env_vars: {
    key: "RUN_LINTING_TYPING_TESTS"
    value: "true"
}
```

.kokoro/presubmit/linting-typing.cfg
```diff
@@ -0,0 +1,11 @@
...
+env_vars: {
+    key: "RUN_LINTING_TYPING_TESTS"
+    value: "true"
```
chalmerlowe, line 10:
```suggestion
    value: "false"
```

.kokoro/presubmit/linting-typing.cfg
```diff
@@ -0,0 +1,11 @@
...
+
+# Only run these nox sessions.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "lint lint_setup_py blacken mypy mypy_samples pytype"
+}
+env_vars: {
+    key: "RUN_LINTING_TYPING_TESTS"
+    value: "false"
+}
```
chalmerlowe, line 11:
```suggestion
```

.kokoro/presubmit/presubmit.cfg
```diff
@@ -9,3 +9,7 @@ env_vars: {
     key: "RUN_SNIPPETS_TESTS"
     value: "false"
 }
+env_vars: {
+    key: "RUN_LINTING_TYPING_TESTS"
+    value: "test_false"
```
chalmerlowe, line 14:
```suggestion
    value: "false"
```

PR #1884, opened 2024-04-03
feat: support RANGE in queries Part 1: JSON

Linchin:
Part 1 for supporting RANGE in tabledata.list and jobs.getQueryResults. This PR takes care of the json part when RowIterator is used as an iterator.

Part 2: #1868 

Part of #1724🦕


tests/unit/test__helpers.py
```diff
@@ -452,6 +452,99 @@ def test_w_bogus_string_value(self):
...
+    def test_w_none_nullable(self):
+        self.assertIsNone(self._call_fut(None, _Field("NULLABLE")))
+
+    def test_w_none_required(self):
+        with self.assertRaises(TypeError):
+            self._call_fut(None, _Field("REQUIRED"))
+
+    def test_w_wrong_format(self):
+        range_field = _Field(
+            "NULLIBLE",
```
shollyman, line 470: s/NULLIBLE/NULLABLE/ throughout this file.

shollyman (approved):
Note about misspelled values in the test classes,  but otherwise LGTM.  Thanks for putting this together.

tests/unit/test__helpers.py
Linchin: Thanks for catching this, I just replaced all of them

PR #1889, opened 2024-04-08
feat: adds billing to opentel

### LINKED ISSUE #1623: Extend the OTEL traces to track bytes billed and bytes processed

chalmerlowe:
This PR adds two features to OpenTelemetry logging in this module:

* `total_bytes_billed`
* `total_bytes_processed`

This is based on an original PR by MrwanBaghdad.

Fixes #1623  🦕

### DIFF
--- a/google/cloud/bigquery/opentelemetry_tracing.py
+++ b/google/cloud/bigquery/opentelemetry_tracing.py
@@ -153,4 +153,12 @@ def _set_job_attributes(job_ref):
     if job_ref.num_child_jobs is not None:
         job_attributes["num_child_jobs"] = job_ref.num_child_jobs
 
+    total_bytes_billed = getattr(job_ref, "total_bytes_billed", None)
+    if total_bytes_billed is not None:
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/opentelemetry_tracing.py
```diff
@@ -153,4 +153,12 @@ def _set_job_attributes(job_ref):
     if job_ref.num_child_jobs is not None:
         job_attributes["num_child_jobs"] = job_ref.num_child_jobs
 
+    total_bytes_billed = getattr(job_ref, "total_bytes_billed", None)
```
Linchin, line 156: I see that `job_ref` is of class `google.cloud.bigquery.job._AsyncJob`, which doesn't have the two attributes you added (nor does it inherit one). Under what condition will these attribute be present? Should we add them to `_AsyncJob`?

Linchin (approved):
lgtm

PR #967, opened 2021-09-13
feat: add `api_method` parameter to `Client.query` to select `INSERT`  or `QUERY` API

tswast:
This PR is the first step in enabling the "fast query path". At the moment, selecting `api_method="QUERY"` is actually likely to be slower because it fetches the first page of results and discards them. A PR will follow-up to cache this first page for use by the row iterator.

~This commit only refactors to allow jobs.insert to be selected.
Supporting jobs.query will require more transformations to QueryJobConfig,
QueryJob, and RowIterator.~

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Towards #589 🦕

PR #1900, opened 2024-04-12
fix: avoid unnecessary API call in QueryJob.result() when job is already finished

tswast:
BEGIN_COMMIT_OVERRIDE
perf: avoid unnecessary API call in `QueryJob.result()` when job is already finished (#1900)

fix: retry query jobs that fail even with ambiguous `jobs.getQueryResults` REST errors (#1903, googleapis/python-bigquery#1900)
END_COMMIT_OVERRIDE

* Maybe fixes internal issue 332850329
* Towards https://github.com/googleapis/google-cloud-python/issues/14476 and https://github.com/googleapis/google-cloud-python/issues/15513 since this loop for waiting for the query to finish will be easier to add a progress bar (if we decide that's needed)
* 🦕


tswast:
CC @chalmerlowe

tests/unit/job/test_query.py
```diff
@@ -1316,37 +1343,6 @@ def test_result_w_empty_schema(self):
         self.assertEqual(result.location, "asia-northeast1")
         self.assertEqual(result.query_id, "xyz-abc")
 
-    def test_result_invokes_begins(self):
```
tswast, line 1319: As of https://github.com/googleapis/python-bigquery/pull/967 released in google-cloud-bigquery 3.0.0, the _begin method is no longer used for query jobs.

tests/unit/job/test_query.py
```diff
@@ -1289,7 +1305,18 @@ def test_result_w_retry(self):
         )
 
         connection.api_request.assert_has_calls(
-            [query_results_call, query_results_call, reload_call]
+            [
+                reload_call,
```
tswast, line 1309: It seems we weren't actually retrying before. Oops!

google/cloud/bigquery/job/query.py
```diff
@@ -1430,40 +1430,6 @@ def _reload_query_results(
             timeout=transport_timeout,
         )
 
-    def _done_or_raise(self, retry=DEFAULT_RETRY, timeout=None):
```
tswast, line 1433: This was overridden because we wanted `result()` from the superclass to call `jobs.getQueryResults`, not just `jobs.get` (i.e. `job.reload()` in Python). Now that we aren't using the superclass for `result()`, this method is no longer necessary.

google/cloud/bigquery/job/query.py
```diff
@@ -1430,40 +1430,6 @@ def _reload_query_results(
...
-        # Only reload the job once we know the query is complete.
-        # This will ensure that fields such as the destination table are
-        # correctly populated.
-        if not self._query_results.complete:
-            raise polling_future._OperationNotComplete()
-        else:
-            try:
-                self.reload(retry=retry, timeout=transport_timeout)
-            except exceptions.GoogleAPIError as exc:
-                self.set_exception(exc)
```
tswast, line 1465: Thought: We probably should have been calling `set_exception` based on the job status. Need to look into this furteher.

tswast: OK. We are. 😅 

https://github.com/googleapis/python-bigquery/blob/bf8861c3473a1af978db7a06463ddc0bad86f326/google/cloud/bigquery/job/base.py#L909

Which we call from `_set_properties`

https://github.com/googleapis/python-bigquery/blob/bf8861c3473a1af978db7a06463ddc0bad86f326/google/cloud/bigquery/job/base.py#L677

Which we call from `reload`

https://github.com/googleapis/python-bigquery/blob/bf8861c3473a1af978db7a06463ddc0bad86f326/google/cloud/bigquery/job/base.py#L838

google/cloud/bigquery/job/query.py
```diff
@@ -1548,46 +1520,92 @@ def result(  # type: ignore  # (incompatible with supertype)
...
+                        # making any extra API calls if the previous loop
+                        # iteration fetched the finished job.
+                        self._reload_query_results(retry=retry, timeout=timeout)
+                        return True
+
+                # Call jobs.getQueryResults with max results set to 0 just to
+                # wait for the query to finish. Unlike most methods,
+                # jobs.getQueryResults hangs as long as it can to ensure we
+                # know when the query has finished as soon as possible.
+                self._reload_query_results(retry=retry, timeout=timeout)
```
tswast, line 1577: Uh oh, if `jobs.getQueryResults` fails because the job failed it can throw an exception but `restart_query_job` will still be `False`.

But we don't want `restart_query_job = True` because sometimes this can raise an ambiguous exception such as quota exceeded, where we don't know if it's the job quota and it's a failed job or at a higher level (Google Frontend - GFE) where the job might actually still be running and/or succeeded.

tswast: This isn't the worst way to fail, but it'd be nice to do the `jobs.get` call above in case of an exception to get a chance at retrying this job if the job failed.

tswast: Filed #1903 to track improvements to ambiguous errors. https://github.com/googleapis/python-bigquery/pull/1900/commits/12fa9fb5823a8d66ddd2fbfdd6dc252388bf3cda fixes an issue where we weren't actually retrying after an ambiguous failure even though we thought we were.

google/cloud/bigquery/retry.py
```diff
@@ -73,10 +73,32 @@ def _should_retry(exc):
...
+    "backendError",
+    "internalError",
+    "jobRateLimitExceeded",
+)
 
 
 def _job_should_retry(exc):
+    # Sometimes we have ambiguous errors, such as 'backendError' which could
+    # be due to an API problem or a job problem. For these, make sure we retry
+    # our is_job_done function.
```
chalmerlowe, line 87:
```suggestion
    # our is_job_done() function.
```

tests/unit/job/test_query.py
```diff
@@ -970,7 +877,12 @@ def test_result(self):
             "rows": [{"f": [{"v": "abc"}]}],
         }
         conn = make_connection(
-            query_resource, query_resource_done, job_resource_done, query_page_resource
+            job_resource,
```
chalmerlowe, line 880: Not sure I am tracking the relationship between the make connection inputs versus the assert_has_calls checks.

Can you explain how these tests are supposed to work?

tests/unit/job/test_query.py
```diff
@@ -1289,7 +1217,18 @@ def test_result_w_retry(self):
         )
 
         connection.api_request.assert_has_calls(
-            [query_results_call, query_results_call, reload_call]
+            [
+                reload_call,
```
chalmerlowe, line 1221: Same thing here. Can I get some clarity on what we are doing and looking for?

chalmerlowe (changes requested):
I am halfway through my review. 
Releasing these comments for now.
Will come back to this to finish out my review as soon as possible.

tests/unit/test_job_retry.py
```diff
@@ -126,6 +126,168 @@ def api_request(method, path, query_params=None, data=None, **kw):
...
+    )
+    responses = [
+        # jobs.insert
+        {"jobReference": job_reference_1, "status": {"state": "PENDING"}},
+        # jobs.get
+        {"jobReference": job_reference_1, "status": {"state": "RUNNING"}},
+        # jobs.getQueryResults x2
+        #
+        # Note: internalError is ambiguous in jobs.getQueryResults. The
+        # problem could be at the Google Frontend leve or it could be because
```
chalmerlowe, line 168:
```suggestion
        # problem could be at the Google Frontend level or it could be because
```

chalmerlowe (changes requested):
LGTM

tests/unit/job/test_query.py
tswast: Added some explanation here as well as above in the `make_connection()` call.

tests/unit/job/test_query.py
```diff
@@ -1316,41 +1255,54 @@ def test_result_w_empty_schema(self):
         self.assertEqual(result.location, "asia-northeast1")
         self.assertEqual(result.query_id, "xyz-abc")
 
-    def test_result_invokes_begins(self):
+    def test_result_w_timeout_doesnt_raise(self):
+        import google.cloud.bigquery.client
+
         begun_resource = self._make_resource()
```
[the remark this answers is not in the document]
tswast: You're correct. Removed. I don't recall why I added this.

tests/unit/job/test_query.py
```diff
@@ -1316,41 +1255,54 @@ def test_result_w_empty_schema(self):
...
+            method="GET",
+            path=f"/projects/{self.PROJECT}/jobs/{self.JOB_ID}",
+            query_params={"location": "US"},
+            timeout=1.125,
+        )
+        get_query_results_call = mock.call(
+            method="GET",
+            path=f"/projects/{self.PROJECT}/queries/{self.JOB_ID}",
+            query_params={
+                "maxResults": 0,
```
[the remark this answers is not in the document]
tswast: We actually want 0 rows. If we omit this or ask for non-zero number of rows, the `jobs.getQueryResults` API can hang when the query has wide rows (many columns).

google/cloud/bigquery/job/query.py
```diff
@@ -1548,46 +1520,124 @@ def result(  # type: ignore  # (incompatible with supertype)
...
+            # sentinel object indicating a default timeout if we choose to add
+            # one some day. This value can come from our PollingFuture
+            # superclass and was introduced in
+            # https://github.com/googleapis/python-api-core/pull/462.
+            if isinstance(timeout, (float, int)):
+                remaining_timeout = timeout
+            else:
+                # Note: we may need to handle _DEFAULT_VALUE as a separate
+                # case someday, but even then the best we can do for queries
+                # is 72+ hours for hyperpareter tuning jobs:
```
chalmerlowe, line 1616:
```suggestion
                # is 72+ hours for hyperparameter tuning jobs:
```

chalmerlowe (approved):
Thanks for all the comments, etc.
Future me thanks you as well.

LGTM, APPROVED.

PR #1914, opened 2024-05-09
fix: add pyarrow version check for range support

### LINKED ISSUE #1912: Dependency version mismatch on pyarrow
version 3.22.0 uses ArrowDtype which is only available in pyarrow>=10.0.1, but dependency requirement specifies pyarrow >= 3.0.0.

#### Environment details

  - OS type and version: Darwin mac-mini.lan 20.6.0 Darwin Kernel Version 20.6.0 (also on GCP AppEngine)
  - Python version: Python 3.10.11
  - pip version: pip 23.3.1
  - `google-cloud-bigquery` version: 3.22.0

#### Steps to reproduce
  note: google-cloud-aiplatform requires pyarrow 7.0.0 for python 3.10
  1. install google-cloud-bigquery and google-cloud-aiplatform db-dtypes
  2. call to_dataframe on RowIterator

#### Code example

```python
# example
client = bigquery.Client()
df = bqclient.query(sql_statement).result().to_dataframe()
```

#### Stack trace
```
File "/layers/google.python.pip/pip/lib/python3.10/site-packages/google/cloud/bigquery/table.py", line 2283, in to_dataframe
    range_date_dtype = pandas.ArrowDtype(
  File "/layers/google.python.pip/pip/lib/python3.10/site-packages/pandas/core/dtypes/dtypes.py", line 2096, in __init__
    raise ImportError("pyarrow>=10.0.1 is required for ArrowDtype")
ImportError: pyarrow>=10.0.1 is required for ArrowDtype"
```

Linchin:
* adds a version check for pyarrow
* consolidates warning messages
* use a constant in _versions_helper.py for checking range support in both pandas and pyarrow

TODO: add nox test session for the case when a newer pandas is installed, but pyarrow is very old.

Fixes #1912 🦕

### DIFF
--- a/google/cloud/bigquery/_versions_helpers.py
+++ b/google/cloud/bigquery/_versions_helpers.py
@@ -238,6 +238,10 @@ class PandasVersions:
 
 PANDAS_VERSIONS = PandasVersions()
 
+# Since RANGE support in pandas requires specific versions
+# of both pyarrow and pandas, we make this a separate
+# constant instead of as a property of PANDAS_VERSIONS
+# or PYARROW_VERSIONS.
 SUPPORTS_RANGE_PYARROW = (
     PANDAS_VERSIONS.try_import() is not None
     and PANDAS_VERSIONS.installed_version >= _MIN_PANDAS_VERSION_RANGE
--- a/google/cloud/bigquery/_versions_helpers.py
+++ b/google/cloud/bigquery/_versions_helpers.py
@@ -26,6 +26,9 @@ _MIN_BQ_STORAGE_VERSION = packaging.version.Version("2.0.0")
 _BQ_STORAGE_OPTIONAL_READ_SESSION_VERSION = packaging.version.Version("2.6.0")
 _MIN_PANDAS_VERSION = packaging.version.Version("1.1.0")
 
+_MIN_PANDAS_VERSION_RANGE = packaging.version.Version("1.5.0")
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/_versions_helpers.py
```diff
@@ -234,3 +237,10 @@ def try_import(self, raise_if_error: bool = False) -> Any:
 
 
 PANDAS_VERSIONS = PandasVersions()
+
+SUPPORTS_RANGE_PYARROW = (
```
tswast, line 241: I wonder if we should make this a property of `PANDAS_VERSIONS` or `PYARROW_VERSIONS` 🤔 

I suppose since it requires both, it makes sense to pull it separate. Let's add a comment explaining that.

```suggestion
# Since RANGE support in pandas requires specific versions
# of both pyarrow and pandas, we make this a separate
# constant instead of as a property of PANDAS_VERSIONS
# or PYARROW_VERSIONS.
SUPPORTS_RANGE_PYARROW = (
```

tswast (approved):
Thanks!

tswast:
> TODO: add nox test session for the case when a newer pandas is installed, but pyarrow is very old.

Since we unit test on multiple versions of Python, we could use one of the middle `constraints` files for this. That said, I think we usually just mock out the version mismatches in unit tests.

google/cloud/bigquery/_versions_helpers.py
Linchin: Thanks for the suggestion!

PR #1915, opened 2024-05-09
fix: edit presubmit for to simplify configuration

chalmerlowe:
This creates a new cfg that could be used in place of the existing kokoro presubmit test.
The intent here is to avoid using variables to skip certain nox sessions and simply call the nox sessions that we want to execute. The benefit is that...because nox sessions create the virtual env before they execute the session steps, we end up needlessly creating virtual envs for sessions that we end up skipping when we reach the skip/not skip logic.


.kokoro/presubmit/presubmit-2.cfg
```diff
@@ -0,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run these nox sessions.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "unit_noextras-3.7 unit_noextras-3.12 unit-3.7 unit-3.8 unit-3.12 cover docs"
```
chalmerlowe, line 6:
```suggestion
    value: "unit_noextras-3.7" # unit_noextras-3.12 unit-3.7 unit-3.8 unit-3.12 cover docs"
```

.kokoro/presubmit/presubmit-2.cfg
```diff
@@ -0,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run these nox sessions.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "unit_noextras-3.7" # unit_noextras-3.12 unit-3.7 unit-3.8 unit-3.12 cover docs"
```
chalmerlowe, line 6:
```suggestion
    value: "unit_noextras-3.7 unit_noextras-3.12" # unit-3.7 unit-3.8 unit-3.12 cover docs"
```

.kokoro/presubmit/presubmit-2.cfg
```diff
@@ -0,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run these nox sessions.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "unit_noextras-3.7 unit_noextras-3.12" # unit-3.7 unit-3.8 unit-3.12 cover docs"
```
chalmerlowe, line 6:
```suggestion
    value: "unit_noextras-3.7 unit_noextras-3.12 unit-3.7 unit-3.8 unit-3.12 cover docs"
```

chalmerlowe:
Using base names in the config file works fine.
`unit` in the config will lead to `unit-3.7` and `unit-3.12` (whatever versions of Python are fed into the `nox.session`) in the results.

noxfile.py
```diff
@@ -56,6 +56,9 @@
     "docs",
 ]
 
+import time
+from functools import wraps
+
```
chalmerlowe, line 61:
```suggestion
```

shollyman (approved):
the title here seems out of sync with the changes, but I may be missing some context. It appears to simply be an update to the existing config, not a new one.

Linchin (approved):

PR #1917, opened 2024-05-15
feat: adds timer decorator to facilitate debugging

chalmerlowe:
This PR adds a calculation_duration function to figure out how long each of the nox sessions takes to complete.

This allows us to troubleshoot issues if a session seems to be taking longer than expected.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -14,11 +14,13 @@
 
 from __future__ import absolute_import
 
+from functools import wraps
 import pathlib
 import os
 import re
 import shutil
 import nox
+import time
 
 
 MYPY_VERSION = "mypy==1.6.1"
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -40,6 +42,27 @@
...
+
+def _calculate_duration(func):
+    """This decorator prints the execution time for the decorated function."""
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        start = time.monotonic()
+        result = func(*args, **kwargs)
+        end = time.monotonic()
+        total_seconds = round(end - start)
```
shollyman, line 54: nit: I think seconds is sufficient for our needs, and we can probably elide the human readable aspects.

shollyman (approved):

tswast:
FYI: This showed up in the public release notes because of the use of `feat: ` in the commit. `chore: ` may have been more appropriate since users of `google-cloud-bigquery` shouldn't care about this change.

chalmerlowe:
@tswast 
I was unaware of the downstream effects of choosing `feat:` vs `chore:`. Thanks for this heads up.
Oops.
Will keep that in mind for the future.

CC: @Linchin

PR #1919, opened 2024-05-15
feat: add support for map target type in Parquet options

sclmn:
The map target type creates a schema without the added key_value repeated field.

### DIFF
--- a/google/cloud/bigquery/format_options.py
+++ b/google/cloud/bigquery/format_options.py
@@ -105,6 +105,21 @@ class ParquetOptions:
     def enable_list_inference(self, value: bool) -> None:
         self._properties["enableListInference"] = value
 
+    @property
+    def map_target_type(self) -> str:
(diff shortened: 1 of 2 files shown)


Linchin:
Thank you @sclmn for adding the new fields for us! We usually require 100% test coverage. Could you please also add unit tests for the new properties, similar to unit tests [here](https://github.com/googleapis/python-bigquery/blob/a429e8fd997a8850d15d434089869cf24e53c9e6/tests/unit/test_format_options.py#L39)? We also require the first commit and PR title to follow [conventional commits](https://github.com/googleapis/python-bigquery/pull/1919/checks?check_run_id=25017312294). Could you follow the instruction in the link to update your PR? If you need any help with these, feel free to let us know :)

sclmn:
I updated the PR. PTAL.

Linchin (approved):

PR #1925, opened 2024-05-21
perf: decrease the threshold in which we use the BQ Storage Read API

tswast:


google/cloud/bigquery/table.py
```diff
@@ -108,7 +108,10 @@
 
 # How many of the total rows need to be downloaded already for us to skip
 # calling the BQ Storage API?
-ALMOST_COMPLETELY_CACHED_RATIO = 0.333
+# In microbenchmarks, the first full page will be about 10 MB. If a page is
+# about 2 MB, then it's the same speed or faster to download all the results
+# using the BigQuery Storage Read API.
+ALMOST_COMPLETELY_CACHED_RATIO = 10 / (10 + 2)
```
shollyman, line 114: I'm not sure I'm following the comment. In usage this appears to be a row ratio, but the comment is about response wire sizes?  Also, should the value just remain a float literal?  There's no parameter plugging in here.

tswast: Updated the comment to explain more my methodology. Unfortunately, this heuristic isn't perfect, as it breaks down with large rows.

shollyman (approved):
Thanks for the clarifications

PR #1935, opened 2024-05-29
feat: add default timeout for Client.get_job()

### LINKED ISSUE #1922: `QueryJob.result()` not identifying and returning after submitted query completes
(this issue was first taken to a Google employee; they recommended this bug report be submitted here too)

### Expected Behavior
`QueryJob.result()` always returns when a submitted query completes

### Issue

At random and unexplainable frequencies, the `QueryJob.result()` runs indefinitely even after the submitted `job_id` is shown as completed in BigQuery's Job History.

Anecdotical observation which motivated this outreach:
* On 2024-04-29 17:16:20 UTC, job_id "A" was submitted to BigQuery.
* On 2024-05-06, we identified the hung python process, observed through its logs that it was waiting for `job.result()` to complete.
* In the BigQuery console, the Job History for"A" shows that the query was submitted on 2024-04-29 at 17:16:20 UTC and completed ~2 minutes later on 2024-04-29 at 17:18:26 UTC.
* The python process executing `job.result()` required manual termination over 6 days later on 2024-05-06.

We are implementing process-level timeouts to prevent this specific issue from running indefinitely, but this is a bandaid solution to a bug in the `google-cloud-bigquery` python package.

#### Environment details

  - OS type and version: Debian Bookworm
  - Python version: 3.11.4
  - pip version: 24.0
  - `google-cloud-bigquery` version: 3.21.0

#### Steps to reproduce

We are unable to deterministically reproduce this issue.

#### Code example

Simplified code example depicting the objects and methods used to 
```python
import uuid
from google.cloud import bigquery as bq

project_id = "dummy-project"
client = bq.Client(project=project_id)

job_id = str(uuid.uuid4())
sql = "select 1;"  # fake query

job_config = bq.QueryJobConfig(priority=bq.enums.QueryPriority.BATCH)
job = client.query(
    sql,
    job_config=job_config,
    api_method=bq.enums.QueryApiMethod.INSERT,  # this necessary to specify the job_id
    job_id=job_id
)
print(f"Submitting query: {job_id}")
result = job.result()  # run job and wait

# do other things with the result once completed
# sometimes this code is never reached, nor any error raised
```

Linchin:
When calling `QueryJob.result()`, `QueryJob.reload()` is called in the background with an empty timeout value. This leads `QueryJob` to waiting indefinitely if `QueryJob.reload()` loses connection.

This PR:
- Reroutes `QueryJob.reload()` to call `Client.get_job()` instead of calling API directly
- Adds a 128s default timeout for `Client.get_job()`

Fixes #1922 🦕

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2163,7 +2163,7 @@ class Client(ClientWithProject):
                 object.
             retry (Optional[google.api_core.retry.Retry]):
                 How to retry the RPC.
-            timeout (Optinal[float]):
+            timeout (Optional[float]):
                 The number of seconds to wait for the underlying HTTP transport
                 before using ``retry``.
 
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2140,7 +2140,7 @@ class Client(ClientWithProject):
         project: Optional[str] = None,
         location: Optional[str] = None,
         retry: retries.Retry = DEFAULT_RETRY,
-        timeout: Optional[Union[TimeoutType, object]] = DEFAULT_GET_JOB_TIMEOUT,
+        timeout: Optional[TimeoutType] = DEFAULT_GET_JOB_TIMEOUT,
     ) -> Union[job.LoadJob, job.CopyJob, job.ExtractJob, job.QueryJob, job.UnknownJob]:
         """Fetch a job for the project associated with this client.
 
@@ -2163,9 +2163,7 @@ class Client(ClientWithProject):
                 object.
             retry (Optional[google.api_core.retry.Retry]):
                 How to retry the RPC.
-            timeout (Optinal[Union[float, \
-                google.api_core.future.polling.PollingFuture._DEFAULT_VALUE, \
-            ]]):
+            timeout (Optinal[TimeoutType]):
                 The number of seconds to wait for the underlying HTTP transport
                 before using ``retry``.
 
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -82,6 +82,7 @@ from google.cloud.bigquery._helpers import _DEFAULT_HOST_TEMPLATE
 from google.cloud.bigquery._helpers import _DEFAULT_UNIVERSE
 from google.cloud.bigquery._helpers import _validate_universe
 from google.cloud.bigquery._helpers import _get_client_universe
+from google.cloud.bigquery._helpers import TimeoutType
 from google.cloud.bigquery._job_helpers import make_job_id as _make_job_id
 from google.cloud.bigquery.dataset import Dataset
 from google.cloud.bigquery.dataset import DatasetListItem
@@ -124,7 +125,6 @@ pandas = (
     _versions_helpers.PANDAS_VERSIONS.try_import()
 )  # mypy check fails because pandas import is outside module, there are type: ignore comments related to this
 
-TimeoutType = Union[float, None]
 ResumableTimeoutType = Union[
     None, float, Tuple[float, float]
 ]  # for resumable media methods
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -53,7 +53,6 @@ import google.api_core.exceptions as core_exceptions
 from google.api_core.iam import Policy
 from google.api_core import page_iterator
 from google.api_core import retry as retries
-from google.api_core.future.polling import PollingFuture
 import google.cloud._helpers  # type: ignore
 from google.cloud import exceptions  # pytype: disable=import-error
 from google.cloud.client import ClientWithProject  # type: ignore  # pytype: disable=import-error
@@ -2141,7 +2140,7 @@ class Client(ClientWithProject):
         project: Optional[str] = None,
         location: Optional[str] = None,
         retry: retries.Retry = DEFAULT_RETRY,
-        timeout: Union[TimeoutType, object] = DEFAULT_GET_JOB_TIMEOUT,
+        timeout: Optional[Union[TimeoutType, object]] = DEFAULT_GET_JOB_TIMEOUT,
     ) -> Union[job.LoadJob, job.CopyJob, job.ExtractJob, job.QueryJob, job.UnknownJob]:
         """Fetch a job for the project associated with this client.
 
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -2140,7 +2140,7 @@ class Client(ClientWithProject):
         project: Optional[str] = None,
         location: Optional[str] = None,
         retry: retries.Retry = DEFAULT_RETRY,
-        timeout: Optional[TimeoutType] = DEFAULT_GET_JOB_TIMEOUT,
+        timeout: TimeoutType = DEFAULT_GET_JOB_TIMEOUT,
     ) -> Union[job.LoadJob, job.CopyJob, job.ExtractJob, job.QueryJob, job.UnknownJob]:
         """Fetch a job for the project associated with this client.
 
@@ -2163,7 +2163,7 @@ class Client(ClientWithProject):
                 object.
             retry (Optional[google.api_core.retry.Retry]):
                 How to retry the RPC.
-            timeout (Optinal[TimeoutType]):
+            timeout (Optinal[float]):
                 The number of seconds to wait for the underlying HTTP transport
                 before using ``retry``.
 
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -53,6 +53,7 @@ import google.api_core.exceptions as core_exceptions
 from google.api_core.iam import Policy
 from google.api_core import page_iterator
 from google.api_core import retry as retries
+from google.api_core.future.polling import PollingFuture
 import google.cloud._helpers  # type: ignore
 from google.cloud import exceptions  # pytype: disable=import-error
 from google.cloud.client import ClientWithProject  # type: ignore  # pytype: disable=import-error
@@ -107,6 +108,7 @@ from google.cloud.bigquery.retry import (
     DEFAULT_JOB_RETRY,
     DEFAULT_RETRY,
     DEFAULT_TIMEOUT,
+    DEFAULT_GET_JOB_TIMEOUT,
 )
 from google.cloud.bigquery.routine import Routine
 from google.cloud.bigquery.routine import RoutineReference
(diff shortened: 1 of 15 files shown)


google/cloud/bigquery/client.py
```diff
@@ -2139,7 +2140,7 @@ def get_job(
         project: Optional[str] = None,
         location: Optional[str] = None,
         retry: retries.Retry = DEFAULT_RETRY,
-        timeout: TimeoutType = DEFAULT_TIMEOUT,
+        timeout: Optional[Union[TimeoutType, object]] = DEFAULT_GET_JOB_TIMEOUT,
```
tswast, line 2143: I think we discussed that `object` isn't necessary here, only in the job classes, right?

google/cloud/bigquery/client.py
```diff
@@ -2162,7 +2163,9 @@ def get_job(
                 object.
             retry (Optional[google.api_core.retry.Retry]):
                 How to retry the RPC.
-            timeout (Optional[float]):
+            timeout (Optinal[Union[float, \
+                google.api_core.future.polling.PollingFuture._DEFAULT_VALUE, \
```
tswast, line 2167: Same here. `Client` shouldn't have to know about the `PollingFuture` object.

google/cloud/bigquery/job/base.py
```diff
@@ -19,11 +19,12 @@
 import http
 import threading
 import typing
-from typing import ClassVar, Dict, Optional, Sequence
+from typing import Any, ClassVar, Dict, Optional, Sequence, Union
 
 from google.api_core import retry as retries
 from google.api_core import exceptions
 import google.api_core.future.polling
+from google.api_core.future.polling import PollingFuture
```
tswast, line 27: Import modules not classes. https://google.github.io/styleguide/pyguide.html#s2.2-imports

google/cloud/bigquery/job/base.py
```diff
@@ -814,28 +815,36 @@ def reload(
...
                 before using ``retry``.
         """
         client = self._require_client(client)
 
-        extra_params = {}
-        if self.location:
-            extra_params["location"] = self.location
-        span_attributes = {"path": self.path}
+        kwargs: Dict[str, Any] = {}
+        if type(timeout) is object:
```
tswast, line 827: Let's add some comments explaining why we need to omit the timeout if we get `_DEFAULT_VALUE`. (Because we want to use the default API-level timeouts but wait indefinitely for the job to finish)

google/cloud/bigquery/job/base.py
```diff
@@ -913,7 +922,7 @@ def _set_future_result(self):
     def done(
         self,
         retry: "retries.Retry" = DEFAULT_RETRY,
-        timeout: Optional[float] = None,
+        timeout: Optional[Union[float, object]] = PollingFuture._DEFAULT_VALUE,
```
tswast, line 925: Done isn't waiting for the job to complete, so I wonder if we actually need `PollingFuture._DEFAULT_VALUE` here? 

It seems more analogous to `Client.get_job` so maybe we can use the same `DEFAULT_GET_JOB_TIMEOUT,` here? 🤔  We need to double-check that we aren't calling any other API requests like getQueryResults, if so.

google/cloud/bigquery/job/query.py
```diff
@@ -1507,6 +1512,15 @@ def result(  # type: ignore  # (incompatible with supertype)
...
                 # actually correspond to a finished query job.
             )
+
+        # When timeout has default sentinel value, only pass the sentinel
+        # timeout to QueryJob.done(), and use None for the other timeouts.
+        if type(timeout) is object:
+            timeout = None
+            get_job_timeout = PollingFuture._DEFAULT_VALUE
+        else:
+            get_job_timeout = timeout
```
tswast, line 1522: Why not do the kwargs trick here too?

tests/unit/test__job_helpers.py
```diff
@@ -1063,20 +1046,15 @@ def test_query_and_wait_incomplete_query():
...
-        None,  # retry
-        span_name="BigQuery.job.reload",
-        span_attributes={"path": jobs_get_path},
-        job_ref=mock.ANY,
+    conn.api_request.assert_any_call(
         method="GET",
         path=jobs_get_path,
-        query_params={"location": "response-location"},
+        query_params={"projection": "full", "location": "response-location"},
         timeout=None,
```
tswast, line 1053: Why do these tests still have a timeout set to `None` for `jobs.get`?

tswast (changes requested):
Please add some tests where `*Job.reload(timeout=None)`, `*Job.done(timeout=None)`, and `*Job.result(timeout=None)` disable the timeouts for `jobs.get`.

google/cloud/bigquery/job/query.py
Linchin: I wonder if it's necessary - the purpose here is slightly different from when calling `client.get_job()`. I'm only preserving the sentinel value for the `job.done()` call, and use None as default for any other calls.

google/cloud/bigquery/job/base.py
Linchin: `done()` calls `reload()`, which in turn calls `Client.get_job()`. So I added the sentinel values along the path `result()` -> `done()` -> `reload()`. Indeed I realize we don't have to add sentinel for `done()` or `reload()`, because it will be passed from `result()` anyway. As to getQueryResults, I don't think it's called from `done()` on (it's used in `result()`). I think either way (using `PollingFuture._DEFAULT_VALUE` or `DEFAULT_GET_JOB_TIMEOUT`) it's effectively the same, so I think either way it's fine. Are there any other factors I'm not aware of?

google/cloud/bigquery/job/base.py
Linchin: I added explanation why `kwargs` is used here, but I feel like `Queryjob.result()` is a better place to explain the default timeout behavior, so I added it in docstring there instead.

google/cloud/bigquery/client.py
Linchin: Thanks! Cleaned up imports.

google/cloud/bigquery/client.py
Linchin: done

google/cloud/bigquery/job/base.py
Linchin: Thanks, cleaned up the imports.

tests/unit/test__job_helpers.py
Linchin: Thanks for catching this! The None timeout came from the default value of `query_and_wait()`. I will add sentinel value for this function too.

google/cloud/bigquery/client.py
Linchin: Indeed, I have reverted this.

google/cloud/bigquery/client.py
Linchin: Reverted too.

tests/unit/test__job_helpers.py
Linchin: done

google/cloud/bigquery/job/base.py
```diff
@@ -801,7 +801,7 @@ def reload(
         self,
         client=None,
         retry: "retries.Retry" = DEFAULT_RETRY,
-        timeout: Optional[float] = None,
+        timeout: Optional[Union[float, object]] = POLLING_DEFAULT_VALUE,
```
tswast, line 804: I'd much rather this defaulted to the same timeout as `get_job()` and anywhere that calls `reload()`, we make sure we do the check for `POLLING_DEFAULT_VALUE` before calling `reload()` to make sure we only pass in `float` or `None` or omit the value.

google/cloud/bigquery/job/base.py
```diff
@@ -913,7 +924,7 @@ def _set_future_result(self):
     def done(
         self,
         retry: "retries.Retry" = DEFAULT_RETRY,
-        timeout: Optional[float] = None,
+        timeout: Optional[Union[float, object]] = POLLING_DEFAULT_VALUE,
```
tswast, line 927: Same here. I don't think this needs to handle `POLLING_DEFAULT_VALUE` since it's not waiting for the job to finish, just making at most 1 API call.

google/cloud/bigquery/job/query.py
```diff
@@ -1457,11 +1461,14 @@ def result(  # type: ignore  # (incompatible with supertype)
...
-            timeout (Optional[float]):
+            timeout (Optinal[Union[float, \
+                google.api_core.future.polling.PollingFuture._DEFAULT_VALUE, \
+            ]]):
                 The number of seconds to wait for the underlying HTTP transport
-                before using ``retry``.
-                If multiple requests are made under the hood, ``timeout``
-                applies to each individual request.
+                before using ``retry``. If ``None``, wait indefinitely
+                unless a retriable error is returned. If unset, only the
```
tswast, line 1469: If it's not retriable we'd still exit early.

```suggestion
                unless an error is returned. If unset, only the
```

google/cloud/bigquery/job/query.py
```diff
@@ -1457,11 +1461,14 @@ def result(  # type: ignore  # (incompatible with supertype)
...
+                underlying Client.get_job() API call has timeout, but we still
```
tswast, line 1470: I'm not sure if we need to be this specific, as we do make other API calls and may want to set a default timeout there too (and do sometimes already in the case of jobs.getQueryResults)

```suggestion
                underlying API calls have their default timeouts, but we still
```

google/cloud/bigquery/job/query.py
```diff
@@ -1507,6 +1514,15 @@ def result(  # type: ignore  # (incompatible with supertype)
                 # Intentionally omit job_id and query_id since this doesn't
                 # actually correspond to a finished query job.
             )
+
+        # When timeout has default sentinel value, only pass the sentinel
+        # timeout to QueryJob.done(), and use None for the other timeouts.
+        if type(timeout) is object:
+            timeout = None
+            get_job_timeout = POLLING_DEFAULT_VALUE
```
tswast, line 1522: Why not set to `DEFAULT_GET_JOB_TIMEOUT` or do the kwargs trick to omit it?

google/cloud/bigquery/job/base.py
tswast: I'd like to keep our type annotations as narrow as we can. My worry with `PollingFuture._DEFAULT_VALUE` is it opens it up to `object`, which technically is anything (string, bytes, anything) in Python.

tests/unit/test__job_helpers.py
tswast: Thanks!

google/cloud/bigquery/job/base.py
Linchin: Indeed, I ended up only using sentinel for `result()`, and used `DEFAULT_GET_JOB_TIMEOUT` for subsequent calls.

google/cloud/bigquery/job/query.py
Linchin: done, following later comments

google/cloud/bigquery/job/base.py
Linchin: done

google/cloud/bigquery/job/base.py
Linchin: done

google/cloud/bigquery/job/query.py
Linchin: done

tswast (approved):
Love it! Just a couple of typos to fix please before merging

PR #1941, opened 2024-06-03
chore: updates continuous CI/CD tests against specific versions of Python

chalmerlowe:
Updates the regular continuous CI/CD checks to test against specific versions of Python (versions that aren't our most recent supported version and aren't our oldest supported version).

Also removes a CI/CD check that is superceded by a more recent version of check (prerelease-deps >>> replaced by prerelease-deps-3.12).

Modifies owlbot to avoid it adding prerelease-deps back into the mix since that file is a default in synthtool.

### DIFF
--- a/.kokoro/continuous/prerelease-deps.cfg
+++ b/.kokoro/continuous/prerelease-deps.cfg
@@ -1,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run this nox session.
+env_vars: {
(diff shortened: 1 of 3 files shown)


parthea:
@chalmerlowe, To delete a file which is part of synthtool templates, add file to the list of excludes here: https://github.com/googleapis/python-bigquery/blob/9fbad767cc228e02040436742d0cb6743d370b90/owlbot.py#L64-L75

shollyman (approved):

PR #1942, opened 2024-06-03
perf: if `page_size` or `max_results` is set on `QueryJob.result()`, use to download first page of results

tswast:

Fixes internal issue 344008814
 🦕


google/cloud/bigquery/job/query.py
```diff
@@ -1515,11 +1529,25 @@ def result(  # type: ignore  # (incompatible with supertype)
                 # actually correspond to a finished query job.
             )
 
+        # Setting max_results should be equivalent to setting page_size with
+        # regards to allowing the user to tune how many results to download
+        # while we wait for the query to finish. See internal issue:
+        # 344008814.
+        if page_size is None and max_results is not None:
```
Linchin, line 1536: Just for my education, why are setting max_results and page_size equivalent here?

google/cloud/bigquery/client.py
```diff
@@ -1981,20 +1983,26 @@ def _get_query_results(
...
-        extra_params: Dict[str, Any] = {"maxResults": 0}
+        extra_params: Dict[str, Any] = {"maxResults": page_size}
 
         if timeout is not None:
             if not isinstance(timeout, (int, float)):
                 timeout = _MIN_GET_QUERY_RESULTS_TIMEOUT
             else:
                 timeout = max(timeout, _MIN_GET_QUERY_RESULTS_TIMEOUT)
 
+        if page_size > 0:
```
Linchin, line 2003: Also for my education, why is the timestamp option related to maxResults?

Linchin (approved):
LGTM and thank you :)

If I understood correctly, in the future if a user wants to make `query_and_wait()` more efficient, they pass a `page_size` value so we can use it to reduce an additional API call. Should we add some documentation for this use case?

google/cloud/bigquery/client.py
tswast: If we accept any results in the response, we want to make sure the data is encoded correctly. Without this option, BigQuery encodes timestamps as floating point numbers, which results in a loss of microsecond-level precision.

The reason it matters is that previously we always asked for 0 rows, in which case we didn't include this option.

I suppose we could always included it, but it's not necessary if we don't download any data.

google/cloud/bigquery/job/query.py
tswast: Imagine the user has selected a value for `max_results`. That means we should not download more than that number of rows, in which case it's important to set the page size.

Also, it means the user is expecting to download the results and do something with them, which means it wouldn't be a waste to ask for the first page of results while we wait for the query to finish.

tests/unit/job/test_query.py
```diff
@@ -1139,114 +1156,92 @@ def test_result_with_none_timeout(self):
             timeout=None,
         )
 
-    def test_result_with_done_jobs_query_response_and_page_size_invalidates_cache(self):
```
tswast, line 1142: >         """We don't call jobs.query with a page size, so if the user explicitly
>        requests a certain size, invalidate the cache.
>        """

This is no longer true. We do set the page size in the initial getQueryResults call.

PR #1944, opened 2024-06-04
fix: create query job in job.result() if doesn't exist

### LINKED ISSUE #1940: QueryJob class break flow after upgrade to 3.21.0
Affect version: >=3.21.0, <=3.23.1

#### Steps to reproduce

``` python
import uuid

client = bigquery.Client()
from google.cloud.bigquery import QueryJob
QueryJob(job_id=str(uuid.uuid4()), query="select 1", client=client).result() # return 404 error, work with version 3.20.1
```

Linchin:
Support a specific user path `QueryJob(job_id="id", query="select 1", client=client).result()` that was broken in #1900.

Fixes #1940 🦕

### DIFF
--- a/tests/unit/job/test_query.py
+++ b/tests/unit/job/test_query.py
@@ -1106,9 +1106,10 @@ class TestQueryJob(_Base):
             timeout=None,
         )
 
-        # assert connection.api_request.call_args == []
         connection.api_request.assert_has_calls(
             [
+                # Make sure we start a job that hasn't started yet. See:
+                # https://github.com/googleapis/python-bigquery/issues/1940
                 create_job_call,
                 reload_call,
                 get_query_results_call,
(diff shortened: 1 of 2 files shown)


tests/unit/job/test_query.py
```diff
@@ -1037,6 +1037,85 @@ def test_result_dry_run(self):
...
+            method="GET",
+            path=f"/projects/{self.PROJECT}/queries/{self.JOB_ID}",
+            query_params={
+                "maxResults": 0,
+                "location": "US",
+            },
+            timeout=None,
+        )
+
+        # assert connection.api_request.call_args == []
```
tswast, line 1109: Nit: leftover test code?

```suggestion
```

tests/unit/job/test_query.py
```diff
@@ -1037,6 +1037,85 @@ def test_result_dry_run(self):
...
+        connection.api_request.assert_has_calls(
+            [
+                create_job_call,
```
tswast, line 1112: Since this is the most important assertion, let's add a comment so folks don't remove it by accident.

```suggestion
                # Make sure we start a job that hasn't started yet. See:
                # https://github.com/googleapis/python-bigquery/issues/1940
                create_job_call,
```

tswast (approved):

PR #1945, opened 2024-06-04
feat: add prefer_bqstorage_client option for Connection

### LINKED ISSUE #1920: Allow enabling/disabling the storage API on a per-connection basis
In #16, support was added for the BigQuery storage API in the form of a `bqstorage_client` which is passed to or created by [`connect()`](https://github.com/googleapis/python-bigquery/blob/b739596f37b8c00b375cc811c316b618097d761a/google/cloud/bigquery/dbapi/connection.py#L98). Yet there is no way, other than installing or uninstalling the storage API itself, to toggle whether the storage API is enabled/disabled.

Since use of the storage API does affect [pricing](https://cloud.google.com/bigquery/pricing), it would be preferable to enable or disable on a per-connection basis, perhaps as a `use_storage_api: bool` parameter on `connect()`. This way, callers could choose at will whether to use the storage API without having to change the Python environment the code is running in.

As a hack to workaround this issue, I attempted calling `connect` passing a normal `Client` object rather than a `BigQueryReadClient` in the `bqstorage_client` parameter, but this does not create a functioning usable connection.

Linchin:
This gives users the option to not use the Storage client if they prefer the REST client.

Fixes #1920 🦕

### DIFF
--- a/google/cloud/bigquery/dbapi/connection.py
+++ b/google/cloud/bigquery/dbapi/connection.py
@@ -35,12 +35,18 @@ class Connection(object):
             A client that uses the faster BigQuery Storage API to fetch rows from
             BigQuery. If not passed, it is created using the same credentials
             as ``client`` (provided that BigQuery Storage dependencies are installed).
-
-            If both clients are available, ``bqstorage_client`` is used for
-            fetching query results.
+        prefer_bqstorage_client (Optional[bool]):
+            Prefer the BigQuery Storage client over the REST client. If Storage
+            client isn't available, fall back to the REST client. Defaults to
+            ``True``.
     """
 
-    def __init__(self, client=None, bqstorage_client=None):
+    def __init__(
+        self,
+        client=None,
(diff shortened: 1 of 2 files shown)


chalmerlowe (approved):
QUESTION: Would we consider this a breaking change since we are moving to having a default setting of `True` instead of `False` which is what I feel like the current setting is.

I am approving this, pending an answer to the above question.
Would be happy to get in a chat OR a meeting to discuss, as need. be.

Linchin:
> QUESTION: Would we consider this a breaking change since we are moving to having a default setting of `True` instead of `False` which is what I feel like the current setting is.
> 
> I am approving this, pending an answer to the above question. Would be happy to get in a chat OR a meeting to discuss, as need. be.

Thanks for the question! The default behavior in the past is `True`. As in https://github.com/googleapis/python-bigquery/issues/1920, the customer couldn't disable the storage client unless they uninstall the package altogether.

PR #1949, opened 2024-06-12
test: update the results of test based on change to hacker news data

chalmerlowe:
Several tests were failing. presubmit testing. This PR updates the expected results to match the fetched results.

The reason for the problem is because we updated the online version of the hacker-news data dump and the data changed slightly. This now matches the data dump.

### DIFF
--- a/tests/system/test_client.py
+++ b/tests/system/test_client.py
@@ -1820,8 +1820,6 @@ class TestBigQuery(unittest.TestCase):
             ],
         ]
 
-        print(f"DINOSAUR fetched:\n{fetched_data}")
-        print(f"DINOSAUR expected:\n{expected_data}")
         self.assertEqual(fetched_data, expected_data)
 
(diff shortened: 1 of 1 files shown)


tests/system/test_client.py
```diff
@@ -1788,23 +1788,40 @@ def test_dbapi_fetch_w_bqstorage_client_large_result_set(self):
...
+                    "timestamp",
+                    datetime.datetime(
+                        2006, 10, 9, 18, 40, 33, tzinfo=datetime.timezone.utc
+                    ),
+                ),
             ],
         ]
 
+        print(f"DINOSAUR fetched:\n{fetched_data}")
+        print(f"DINOSAUR expected:\n{expected_data}")
```
Linchin, line 1824:
```suggestion
```

Linchin (approved):
Thanks for fixing the sys test!

PR #1952, opened 2024-06-12
feat: support load job option ColumnNameCharacterMap

### LINKED ISSUE #1951: support load job option ColumnNameCharacterMap
Support new field ColumnNameCharacterMap: https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#columnnamecharactermap

Linchin:
Fixes #1951  🦕

### DIFF
--- a/tests/unit/job/test_load_config.py
+++ b/tests/unit/job/test_load_config.py
@@ -872,3 +872,13 @@ class TestLoadJobConfig(_Base):
             config._properties["load"]["columnNameCharacterMap"],
             ColumnNameCharacterMap.V1,
         )
+
+    def test_column_name_character_map_none(self):
+        from google.cloud.bigquery.job.load import ColumnNameCharacterMap
+
+        config = self._get_target_class()()
+        config.column_name_character_map = None
+        self.assertEqual(
+            config._properties["load"]["columnNameCharacterMap"],
+            ColumnNameCharacterMap.COLUMN_NAME_CHARACTER_MAP_UNSPECIFIED,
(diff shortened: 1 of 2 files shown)


tests/unit/job/test_load_config.py
```diff
@@ -843,3 +843,42 @@ def test_parquet_options_setter_clearing(self):
 
         config.parquet_options = None
         self.assertNotIn("parquetOptions", config._properties["load"])
+
+    def test_column_name_character_map_missing(self):
```
chalmerlowe, line 847: PREFERENCE:

These tests are nearly identical. The only real change is the choice of setting and whether the outcome matches.

This feels like an ideal situation for a parameterized test.

I would recommend we shorten this code by about 27 lines by using a pytest parameterization.

chalmerlowe (approved):
LGTM.
I have one PREFERENCE to make the test code simpler and potentially easier to parse and maintain in the long run.

tests/unit/job/test_load_config.py
Linchin: Thanks for the suggestion! It would be really nice if we could make the tests more compact. For the purpose of this PR, I just followed the same pattern as the rest of this test file. I think we can consider revamping the test as a whole so the style of the tests can be more consistent. I do wonder though, because we are not calling the exact same methods in each of the tests (sometimes we assign value in initialization, sometimes by calling the method, or directly accessing the `_properties` dict), is there a nice way to represent these in  test parameterization?

PR #1953, opened 2024-06-13
feat: support julian day timestamp value conversion to datetime.datetime #1948

### LINKED ISSUE #1948: Support julian day timestamp value conversion to datetime.datetime
I'm using an image called ghcr.io/goccy/bigquery-emulator:0.6.1 as a test container in my unit tests. 
However this image is using sqlite for database. I also use bigquery python client library 3.24.0. 
The problem is the following: 
sqlite is using floats to store timestamps (julian day form) whereas bigquery stores them as integers. 
When converting the int into datetime in _timstamp_from_json function there is a usage of int() which throws a ValueError when a string of a float is provided. 
The error: 
ValueError: invalid literal for int() with base 10

My suggestion is to change the line in _helpers.py  to int(float(value)).

erenkeskus:
I have added float casting as discussed in the issue to be able to support from julian day format. 
Fixes #1948  🦕

### DIFF
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -206,7 +206,7 @@ def _timestamp_from_json(value, field):
     """Coerce 'value' to a datetime, if set or not nullable."""
     if _not_null(value, field):
         # value will be a integer in seconds, to microsecond precision, in UTC.
-        return _datetime_from_microseconds(int(value))
+        return _datetime_from_microseconds(int(float(value)))
 
 
 def _timestamp_query_param_from_json(value, field):


Linchin:
Cheers! Could you
- sign the CLA that's linked in [this comment](https://github.com/googleapis/python-bigquery/pull/1953#issuecomment-2165126855), so we can legally merge your code
- Add a unit test [here](https://github.com/googleapis/python-bigquery/blob/7d757b84da119f45647041196e4921556793404b/tests/unit/test__helpers.py#L303), you can use adjacent methods as a reference

Thanks for your help :)

chalmerlowe:
@erenkeskus @Linchin 

Before we move forward with this, I have a number of questions. See Issue #1948 associated with this PR.

I added the Do Not Merge tag to prevent merging this until we ensure that we are on the same page. 

Please discuss in the Issue. Thanks.

chalmerlowe:
I closed Issue #1948 due to inactivity.
Closing this PR, as well.

PR #1989, opened 2024-08-05
fix: add warning when encountering unknown field types 

### LINKED ISSUE #1988: Add a warning in the type mapping if no conversion function is found
Unsupported types are silently handled in order to allow for tables with column types the client library cannot handle yet. Once these types are supported, the returned type for the column is likely to change to reflect the correct handling of the type. However, this may require different code to handle from the user. 

See #1587 for an example when support for json values was added.

Add a warning to the type mapping when the type does not have a conversion function, since the handling for this type may change:
https://github.com/googleapis/python-bigquery/blob/4383cfe7c7571ffaa3efd5e45ca33c6ff978f274/google/cloud/bigquery/_helpers.py#L385

suzmue:
Fixes #1988

### DIFF
--- a/google/cloud/bigquery/_helpers.py
+++ b/google/cloud/bigquery/_helpers.py
@@ -380,7 +380,9 @@ def _field_to_index_mapping(schema):
 def _field_from_json(resource, field):
     def default_converter(value, field):
         warnings.warn(
-            "Unknown type '{}' for field '{}'.".format(field.field_type, field.name),
+            "Unknown type '{}' for field '{}'. Behavior reading this type is not officially supported and may change in the future.".format(
+                field.field_type, field.name
+            ),
             FutureWarning,
         )
         return value
@@ -485,6 +487,11 @@ def _json_to_json(value):
     if value is None:
         return None
     return json.dumps(value)
+
+
+def _string_to_json(value):
+    """NOOP string -> string coercion"""
+    return value
 
 
 def _timestamp_to_json_parameter(value):
@@ -599,6 +606,7 @@ _SCALAR_VALUE_TO_JSON_ROW = {
     "DATE": _date_to_json,
     "TIME": _time_to_json,
     "JSON": _json_to_json,
+    "STRING": _string_to_json,
     # Make sure DECIMAL and BIGDECIMAL are handled, even though
     # requests for them should be converted to NUMERIC.  Better safe
     # than sorry.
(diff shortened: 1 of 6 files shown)


google/cloud/bigquery/_helpers.py
```diff
@@ -382,7 +378,14 @@ def _field_to_index_mapping(schema):
 
 
 def _field_from_json(resource, field):
-    converter = _CELLDATA_FROM_JSON.get(field.field_type, lambda value, _: value)
+    def default_converter(value, field):
+        warnings.warn(
+            "Unknown type '{}' for field '{}'.".format(field.field_type, field.name),
```
tswast, line 383: Let's be explicit and say "Behavior reading this type is not officially supported and may change in future." or something scary along those lines.

tswast: Before we mark this as "fixed", I think the inverse direction is needed too (where we write JSON from a record), which is used in methods like `client.insert_rows()` (though maybe that already fails when encountering unknown types?)

suzmue: Got it. Yeah looks like https://github.com/googleapis/python-bigquery/blob/4383cfe7c7571ffaa3efd5e45ca33c6ff978f274/google/cloud/bigquery/_helpers.py#L690 and https://github.com/googleapis/python-bigquery/blob/4383cfe7c7571ffaa3efd5e45ca33c6ff978f274/google/cloud/bigquery/_helpers.py#L624 are two points. The first there is no type info so probably just _scalar_field_to_json needs to be updated. Updated PR to include a warning for the write.

google/cloud/bigquery/_helpers.py
```diff
@@ -382,7 +378,16 @@ def _field_to_index_mapping(schema):
 
 
 def _field_from_json(resource, field):
-    converter = _CELLDATA_FROM_JSON.get(field.field_type, lambda value, _: value)
+    def default_converter(value, field):
```
Linchin, line 381: Maybe we can move this function to outside, so we can reuse it for `_scalar_field_to_json()` as well?

suzmue: I ended up just moving the warning itself to a new function, since the converter in _scalar_field_to_json has a different function signature.

Linchin:
I think the docs test isn't failing due to changes in this PR, it appears to relate to the docker image used. I will approve this PR and force merge.

LGTM.

Linchin (approved):

PR #1994, opened 2024-08-14
Bug: revises Exception type

### LINKED ISSUE #1985: Should a rateLimitExceeded have a "429 Too Many Requests" instead of "403 Forbidden"?
In many cases, users want to be able to retry `rateLimitExceeded` errors. Many retry libraries and codebases support this by automatically retrying HTTP exceptions with a `429 Too Many Requests` HTTP error, and any `5XX` error codes. For example, [BigQuery itself](https://github.com/googleapis/python-bigquery/blob/4383cfe7c7571ffaa3efd5e45ca33c6ff978f274/google/cloud/bigquery/retry.py#L26-L36). However BigQuery will raise a `403 Forbidden` for rate limit problems - see: https://github.com/googleapis/python-bigquery/blob/f55864ec3d6381f2b31598428a64822fdc73cb56/google/cloud/bigquery/job/base.py#L49

A `403 Forbidden` implies that the API user does not have the permissions or authentication credentials to do the thing. That seems inappropriate for a `rateLimitError`, right?

This mis-coding has two consequences for retry logic:

* In codebases where you'd expect to get retries for free, you don't, because libraries won't retry a `403 Forbidden`, and rightly so - why would you retry an access error.
* Users who do want to implement retries for BQ `rateLimitExceeded` errors now have to retry on `403 Forbidden`, possibly causing them to retry genuine permissions/auth errors that will fail no matter how many attempts.

chalmerlowe:
Revises an exception type to a type that is more descriptive of the true problem.

Specifically, migrates from an http.client.FORBIDDEN exception to an TooManyRequests Error if a user experiences a rateLimitExceeded condition. This enables automated systems to more appropriately respond in situations where the rate limit is exceeded.

Fixes #1985 🦕

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -49,7 +49,7 @@ _ERROR_REASON_TO_EXCEPTION = {
     "notImplemented": http.client.NOT_IMPLEMENTED,
     "policyViolation": http.client.FORBIDDEN,
     "quotaExceeded": http.client.FORBIDDEN,
-    "rateLimitExceeded": exceptions.TooManyRequests,
+    "rateLimitExceeded": http.client.TOO_MANY_REQUESTS,
     "resourceInUse": http.client.BAD_REQUEST,
     "resourcesExceeded": http.client.BAD_REQUEST,
     "responseTooLarge": http.client.FORBIDDEN,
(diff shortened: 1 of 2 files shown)


suzmue (approved):

cozos:
Thank you so much @chalmerlowe

PR #2012, opened 2024-09-04
Feat: Add classes and objects to enable PANGEA support 

chalmerlowe:
WIP: do not merge.

Internal reference: `b/358215039`

### DIFF
--- a/google/cloud/bigquery/dataset.py
+++ b/google/cloud/bigquery/dataset.py
@@ -530,6 +530,7 @@ class Dataset(object):
         "storage_billing_model": "storageBillingModel",
         "max_time_travel_hours": "maxTimeTravelHours",
         "default_rounding_mode": "defaultRoundingMode",
+        "external_catalog_dataset_options": "externalCatalogDatasetOptions",
     }
 
     def __init__(self, dataset_ref) -> None:
@@ -937,6 +938,24 @@ class Dataset(object):
         """Generate a resource for ``update``."""
         return _helpers._build_resource_from_properties(self, filter_fields)
 
+    @property
+    def external_catalog_dataset_options(self):
+        """Options defining open source compatible datasets living in the
+        BigQuery catalog. Contains metadata of open source database, schema
+        or namespace represented by the current dataset."""
+       
+        return self._properties.get("externalCatalogDatasetOptions")
+
+    @external_catalog_dataset_options.setter
+    def external_catalog_dataset_options(self, value):
+        if not isinstance(value, ExternalCatalogDatasetOptions) and value is not None:
+            raise ValueError(
+                "external_catalog_dataset_options must be an "
+                "ExternalCatalogDatasetOptions object or None. "
+                f"Got {repr(value)}."
+            )
+        self._properties["externalCatalogDatasetOptions"] = value
+
     table = _get_table_reference
 
     model = _get_model_reference
@@ -1026,3 +1045,36 @@ class DatasetListItem(object):
     model = _get_model_reference
 
     routine = _get_routine_reference
+
+
+class ExternalCatalogDatasetOptions(object):
+    """Options defining open source compatible datasets living in the BigQuery
+    catalog. Contains metadata of open source database, schema or namespace
+    represented by the current dataset.
+    
+    Args:
+        defaultStorageLocationUri:
+            Optional. The storage location URI for all tables in the dataset.
+            Equivalent to hive metastore's database locationUri. Maximum length
+            of 1024 characters. (str)
+
+        parameters:
+            Optional. A map of key value pairs defining the parameters and
+            properties of the open source schema. Maximum size of 2Mib.
+
+    Raises:
+        ValueError: If either argument is not of type ``str``.
+    """
+
+    def __init__(
+        self,
+        default_storage_location_uri: Optional[str] = None,
+        parameters: Optional[dict] = None
+    ):
+        self._properties = {}
+        if not isinstance(default_storage_location_uri, str):
+            raise ValueError("Pass a string as default_storage_location_uri")
+        if not isinstance(parameters, dict):
+            raise ValueError("Pass a dict as parameters to define the schema.")
+        self._properties["default_storage_location_uri"] = default_storage_location_uri
+        self._properties["parameters"] = parameters


chalmerlowe:
@Linchin 

Can you take a preliminary pass through this PR?
I will send you links in internal chat with some additional details.

Primarily hoping for:
* does this approach seem to make sense?
* do you have any edits that I might need to make on all the additional classes, setters, `@property` attributes, etc that I will be generating.

google/cloud/bigquery/external_config.py
```diff
@@ -1003,3 +1003,95 @@ def from_api_repr(cls, resource: dict) -> "ExternalConfig":
...
+
+    Args:
+        defaultStorageLocationUri: Optional. The storage location URI for all
+            tables in the dataset. Equivalent to hive metastore's database
+            locationUri. Maximum length of 1024 characters. (str)
+        parameters: Optional. A map of key value pairs defining the parameters
+            and properties of the open source schema. Maximum size of 2Mib.
+    """
+
+    def __init__(self, defaultStorageLocationUri: Optional[str] = None, parameters: Optional[dict] = None):
```
Linchin, line 1021: I think the parameter names need to be snake case, see [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names).

google/cloud/bigquery/external_config.py
```diff
@@ -1003,3 +1003,95 @@ def from_api_repr(cls, resource: dict) -> "ExternalConfig":
...
+        self._properties = {}
+        if not isinstance(defaultStorageLocationUri, (str, None)):
+            raise ValueError(
+                "Pass defaultStorageLocationUri as a 'str'."
+                f"Got {repr(dtype)}."
```
Linchin, line 1026: I'm not sure I get it, but where does `dtype` come from?

google/cloud/bigquery/external_config.py
```diff
@@ -1003,3 +1003,95 @@ def from_api_repr(cls, resource: dict) -> "ExternalConfig":
...
+        defaultStorageLocationUri: Optional. The storage location URI for all
+            tables in the dataset. Equivalent to hive metastore's database
+            locationUri. Maximum length of 1024 characters. (str)
+        parameters: Optional. A map of key value pairs defining the parameters
+            and properties of the open source schema. Maximum size of 2Mib.
+    """
+
+    def __init__(self, defaultStorageLocationUri: Optional[str] = None, parameters: Optional[dict] = None):
+        self._properties = {}
+        if not isinstance(defaultStorageLocationUri, (str, None)):
```
Linchin, line 1023: I guess you want to verify that `defaultStorageLocationUri` is of type `str` or is `None`, but trying it in Python it doesn't seem to work:
```
>>> isinstance(None, (str, None))
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

google/cloud/bigquery/table.py
```diff
@@ -999,6 +1000,24 @@ def table_constraints(self) -> Optional["TableConstraints"]:
...
+
+    @external_catalog_table_options.setter
+    def external_catalog_table_options(self, value):
+        if not isinstance(value, ExternalCatalogTableOptions) and value is not None:
+            raise ValueError(
+                "external_catalog_table_options must be an "
+                "ExternalCatalogTableOptions object or None. "
+                f"Got {repr(value)}."
+            )
+        self._properties["externalCatalogTableOptions"] = value
```
Linchin, line 1019: When we are calling `self.to_api_repr()`, we are basically returning the dictionary in `self._properties` directly. So we probably need to do `self._properties["externalCatalogTableOptions"] = value.to_api_repr()` here to convert it to a dict.

Linchin:
Thank you for working to add PANGEA support! It's great that I can do a quick review early to make sure we are on the same page. I think the approach looks good, but I left a few comments on implementation details.

google/cloud/bigquery/external_config.py
chalmerlowe: Nice. Learned something new today.

Generally `None` should be tested using the `is` or `is not` comparisons. 

Per PEP 008:
```
Comparisons to singletons like None should always be done with
is or is not, never the equality operators.
```

I believe we prolly wanna go with something akin to this:

### OPTION 1.
```
if item is None or isinstance(item, str)
```


### OPTION 2.
Behind the scenes, the type of a `None` object is `NoneType`. Knowing this, it is possible to use a workaround:

```
> mylist = ['alpha', 1, 42.0, None]
> for item in mylist:
    print(f"{item}:", isinstance(item, (str, type(None))))
alpha: True
1: False
42.0: False
None: True
```

But I think OPTION 1 is cleaner and more in line with the PEP.

Thoughts? @Linchin

Linchin: I like option 1 as well! I think it's more readable.

google/cloud/bigquery/external_config.py
chalmerlowe: done.

google/cloud/bigquery/external_config.py
chalmerlowe: carryover from an earlier iteration. revised.

google/cloud/bigquery/table.py
chalmerlowe: Done. Upgraded several instances throughout the code.

chalmerlowe:
> Thank you for working to add PANGEA support! It's great that I can do a quick review early to make sure we are on the same page. I think the approach looks good, but I left a few comments on implementation details.
> 
> Also, before generating the rest, I think it'd be helpful to add tests to ensure the current code is working as intended.

Tests are next on my list for this first batch of classes.
With some proof of concept classes and tests in hand, we should be able to move on and add the remaining classes.

google/cloud/bigquery/dataset.py
```diff
@@ -937,6 +940,24 @@ def _build_resource(self, filter_fields):
...
+
+        return self._properties.get("externalCatalogDatasetOptions")
+
+    @external_catalog_dataset_options.setter
+    def external_catalog_dataset_options(self, value):
+        if not isinstance(value, ExternalCatalogDatasetOptions) and value is not None:
+            raise ValueError(
+                "external_catalog_dataset_options must be an "
+                "ExternalCatalogDatasetOptions object or None. "
+                f"Got {repr(value)}."
```
Linchin, line 957: NIT

(a really small one) Use `type(value)` to make the error message more explicit. 
```suggestion
                f"Got {type(value)}."
```

google/cloud/bigquery/external_config.py
```diff
@@ -1003,3 +1003,104 @@ def from_api_repr(cls, resource: dict) -> "ExternalConfig":
...
         return config
+
+
+class ExternalCatalogDatasetOptions(object):
+    """Options defining open source compatible datasets living in the BigQuery catalog.
+    Contains metadata of open source database, schema or namespace represented
+    by the current dataset.
+
+    Args:
+        default_storage_location_uri: Optional. The storage location URI for all
```
Linchin, line 1014: NIT:

add type annotations for init arguments too.

google/cloud/bigquery/external_config.py
```diff
@@ -1003,3 +1003,104 @@ def from_api_repr(cls, resource: dict) -> "ExternalConfig":
...
+            tables in the dataset. Equivalent to hive metastore's database
+            locationUri. Maximum length of 1024 characters. (str)
+        parameters: Optional. A map of key value pairs defining the parameters
+            and properties of the open source schema. Maximum size of 2Mib.
+    """
+
+    def __init__(
+        self,
+        default_storage_location_uri: Optional[str] = None,
```
Linchin, line 1023: I think it would be good if we add properties for these values too. I wonder if we should also provide a property setter, or make it read only.

google/cloud/bigquery/table.py
```diff
@@ -999,6 +1000,24 @@ def table_constraints(self) -> Optional["TableConstraints"]:
             table_constraints = TableConstraints.from_api_repr(table_constraints)
         return table_constraints
 
+    @property
+    def external_catalog_table_options(self):
```
Linchin, line 1004: Here we might want to return an `ExternalCatalogTableOptions` object, so the user get the same thing as their input, instead of a dictionary representation of it. [Here is a similar example](https://github.com/googleapis/python-bigquery/blob/255472359f3ed6b6cee06039ebe9059607fd9894/google/cloud/bigquery/job/query.py#L339-L352).

google/cloud/bigquery/dataset.py
chalmerlowe: Cool. I will make these changes from `repr` to `type`.

google/cloud/bigquery/external_config.py
chalmerlowe: I am not sure what we mean by this.
The init arguments on lines 1023 and 1024 have type hinting as do init args on lines 1072-1074.
(I don't know if that is the result of my continuing to muck around with changes while this comment was being generated OR if you are requesting something else that I am just not understanding).

Please advise. Thanks.

google/cloud/bigquery/table.py
chalmerlowe: This makes sense. I can see where something like this could be better represented by an object not a dict but where other items might be fine as just a dict.

Good call.

Will do.

google/cloud/bigquery/external_config.py
chalmerlowe: @Linchin 

In thinking about this...

I think we want/need a different approach.

Rather than having all the input validation in the `__init__()` method, I think I wanna have `__init__()` call the setter and let the setter do the input validation so that I only need that code in one place.

Thoughts?

This `Table` class in the `table.py` file does this with the schema argument:

```
if schema is not None:
    self.schema = schema
```

and relies on input validation here:

```
@schema.setter
def schema(self, value):
    api_field = self._PROPERTY_TO_API_FIELD["schema"]

    if value is None:
        self._properties[api_field] = None
    else:
        value = _to_schema_fields(value)
        self._properties[api_field] = {"fields": _build_schema_resource(value)}
```

https://github.com/googleapis/python-bigquery/blob/255472359f3ed6b6cee06039ebe9059607fd9894/google/cloud/bigquery/table.py#L355C1-L417C1

Linchin: I think this is a great idea! This way we can reduce redundancy significantly.

google/cloud/bigquery/external_config.py
chalmerlowe: Gotcha. will see what we can do.

tests/unit/test_schema.py
```diff
@@ -1108,3 +1114,198 @@ def test_to_api_repr_parameterized(field, api):
...
+    @pytest.mark.parametrize(
+        "serde_info,expected",
+        [
+            ({"key": "value"}, {"key": "value"}),
+            ([1, 2, 3], [1, 2, 3]),
+            (123, 123),
+            ("string", "string"),
+        ],
+    )
+    '''
```
chalmerlowe, line 1245:
```suggestion
```

google/cloud/bigquery/_helpers.py
```diff
@@ -1004,3 +1005,61 @@ def _verify_job_config_type(job_config, expected_type, param_name="job_config"):
...
+
+    Raises:
+        TypeError: If the input value's type does not match the expected data type(s).
+    """
+    
+    # Simplest case
+    if dtype is None and value is None:
+        return value
+    
+    elif isinstance(dtype, tuple):
```
Linchin, line 1052: I think the built-in `isinstance()` supports checking multiple types: https://stackoverflow.com/questions/33311258/python-check-if-variable-isinstance-of-any-type-in-list

google/cloud/bigquery/_helpers.py
```diff
@@ -1004,3 +1005,61 @@ def _verify_job_config_type(job_config, expected_type, param_name="job_config"):
...
+
+    Returns:
+        Any: Returns the input value if the type check is successful.
+
+    Raises:
+        TypeError: If the input value's type does not match the expected data type(s).
+    """
+    
+    # Simplest case
+    if dtype is None and value is None:
```
Linchin, line 1049: I suppose this case is trying to verify `value is None`, but in this case shouldn't we have `dtype == NoneType`, and `isinstance(value, dtype)` still applies here? Also, when `dtype is None`, it's a bit ambiguous whether we are expecting `None`, or `dtype` is by accident empty.

google/cloud/bigquery/_helpers.py
```diff
@@ -1004,3 +1005,61 @@ def _verify_job_config_type(job_config, expected_type, param_name="job_config"):
                 job_config=job_config,
             )
         )
+
+
+class ResourceBase:
```
Linchin, line 1010: I saw the discussion between you and Tim, but on a second thought, I feel like it might be more prudent to keep the boilerplate code instead of using mixin. It indeed is some duplicate code, but I think
- they are very readable duplicate code
- we avoid any unintended consequence of multiple inheritance, for example, if the child class by accident has a method or property with the same name as the parent class, it may cause mysterious behavior and become very hard to debug
- we ensure each resource class is strictly self-contained
- it's easier to maintain, for example, future programmers don't have to dig into the parent class to find where `self._properties` is
- having `def from_api_repr():` present in each resource class reminds people to write unit tests for it, and coverage test will report an error if we didn't

Overall I feel like the benefit is much outweighed by the cost. This is just a preference of mine, and I would be glad to hear your opinion on this and discuss more!

google/cloud/bigquery/_helpers.py
chalmerlowe: Both `dtype` and `value` are optional and thus the default argument is set to `None`. So we are checking for that condition.

My understanding is that `None` and `NoneType` don't behave the way we might anticipate:
`NoneType` is not defined in Python as a value that can be assigned.
This means we cannot use `isinstance()` to compare `None` to `NoneType` nor can we use `== NoneType` for comparisons.

Relatedly, [PEP 8 speaks to how to compare `None`](https://peps.python.org/pep-0008/#programming-recommendations
). Due to `None`'s unique nature in Python and the fact that `__eq__()` methods can be created somewhat arbitrarily, one should not use `value == None` for comparison sake, because that is checking to see if they are equal and that is subject possible issues. Instead, the recommended approach is `value is None`.

> Comparisons to singletons like `None` should always be done with `is` or `is not`, never the equality operators.

Here are some examples of how Python reacts to various uses of `None` and `NoneType`:

```
In [2]: dtype = NoneType
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 dtype = NoneType

NameError: name 'NoneType' is not defined

In [3]: dtype = None

In [4]: isinstance(dtype, NoneType)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 isinstance(dtype, NoneType)

NameError: name 'NoneType' is not defined

In [5]: isinstance(dtype, None)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 isinstance(dtype, None)

TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union

In [6]: dtype == None
Out[6]: True

In [7]: dtype == NoneType
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 dtype == NoneType

NameError: name 'NoneType' is not defined
```

google/cloud/bigquery/_helpers.py
chalmerlowe: You are absolutely right, `isinstance()` will check multiple types if you supply a `tuple`.

But it does not correctly handle the object `None`. 

*That is one of the main reasons this function has so many checks, because we are adding to the capability of the built-in `isinstance()` functionality.*

```
In [5]: isinstance(dtype, None)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 isinstance(dtype, None)

TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

Interestingly, `isinstance()` **shortcircuits** (stops all checks as soon as one passes), so it is very easy to accidentally believe that it is behaving as one expects.

Here it does not balk unless `None` is first, cause it succeeds with `int` in the other cases and never checks against `None`.

```
In [8]: test_value = 42

In [9]: isinstance(test_value, (int, float, None))
Out[9]: True

In [11]: isinstance(test_value, (int, None, float))
Out[11]: True
```

But here it freaks out:

```
In [10]: isinstance(test_value, (None, int, float))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 isinstance(test_value, (None, int, float))
```

chalmerlowe: This is also why my test cases include tuples and in the tuples include None in the first and last position to ensure that our function behaves as expected in all cases.

```
(None, None, None),
('hello world.uri', str, 'hello world.uri'),
(None, (str, None), None),
(None, (None, str), None),
(None, (str, None), None),
('hello world.uri', (None, str), 'hello world.uri'),
('hello world.uri', (str, None), 'hello world.uri'),
```

google/cloud/bigquery/_helpers.py
chalmerlowe: I hear you on the desire to avoid inheritance where we can. Good point.

I have several desires/hopes for our work in BQ over time.
* Improve the functions as a whole with docstrings, etc.
* Reduce duplicate code where appropriate

If we pursue the idea of having a docstring for a function like this (which arguably may OR may not be necessary), we are looking at something akin to 10+ lines of code in a function that shows up ~60 times for about 600+ lines of code if we universally update all the docstrings now that we have one that is viable.

Another alternative is to create a helper function and then reference the helper function in our classes (NOTE: THIS IS PSEUDOCODE ... I have not actually tested all this. Just posting something to facilitate discussion. The actual code will likely be slightly different):

```
def _from_api_repr(obj, resource: dict):
    # Add docstring here ...
    config = obj
    return config

from google.cloud.bigquery._helpers import _from_api_repr

class <ClassName>:
    def from_api_repr(self, resource):
        return = _from_api_repr(self, resource)  # Pass self to function as an arg.
```

The benefit of a helper function:
* No class inheritance
* Trivial to override for specific use cases
* Reduces some duplication in the Docstring area (I am aware it doesn't really save much in terms of reducing the amount of code duplication in each of the classes)

### Function in question:
```
@classmethod
def from_api_repr(cls, resource: dict):
    """Factory: constructs an instance of the class (cls)
    given its API representation.
    Args:
        resource (Dict[str, Any]):
            API representation of the object to be instantiated.
    Returns:
        ResourceBase: An instance of the class initialized with data
            from 'resource'.
    """
    config = cls()
    return config     
```

suzmue: I think the idea of a helper function is a good balance. It's clear that the code is duplicated, and adds the clarity of having the function be present in the class.

google/cloud/bigquery/_helpers.py
```diff
@@ -1004,3 +1005,61 @@ def _verify_job_config_type(job_config, expected_type, param_name="job_config"):
...
+        # Iterate through the tuple and check if value is an instance of any type
+        if not any(isinstance(value, t) for t in dtype if t is not None):
+            valid_types_str = ", ".join(str(t) for t in dtype if t is not None)
+            msg = f"Pass {value} as one of '{valid_types_str}' or None. Got {type(value)}."
+            raise TypeError(msg)
+    else:
+        if not isinstance(value, dtype):
+            msg = f"Pass {value} as a '{dtype}'. Got {type(value)}."
+            raise TypeError(msg)
+    return value
```
suzmue, line 1065: I think its a bit strange mixing None and types, but also the type hints not really being able to indicate that a type is expected. Since None is such a special case (as shown in the comment thread with Lingqing), I think it could make sense to split out the None part into its own argument, to effectively duplicate the isinstance() function signature, but add a boolean for if none is allowed.

```suggestion
def _isinstance_or_raise(value: Any, dtype: type, none_allowed: Optional[bool]):
    """Determine whether a value type matches a given datatype or None.

    Args:
        value (Any): Value to be checked.
        dtype (type): Expected data type(s).
        none_allowed Optional(bool): whether value is allowed to be None.

    Returns:
        Any: Returns the input value if the type check is successful.

    Raises:
        TypeError: If the input value's type does not match the expected data type(s).
    """
    if none_allowed and value is None:
        return value

    if isinstance(value, dtype):
        return value
    
    msg = f"Pass {value} as a '{dtype}' (or None). Got {type(value)}." # Add the 'or None' conditionally
    raise TypeError(msg)
```

WDYT?

google/cloud/bigquery/_helpers.py
```diff
@@ -1007,59 +1007,50 @@ def _verify_job_config_type(job_config, expected_type, param_name="job_config"):
...
-        """
-        config = cls()
-        config._properties = copy.deepcopy(resource)
-        return config
-
-
-def _isinstance_or_raise(value: Any, dtype: Optional[Union[Any, Tuple[Any, ...]]]):
+def _isinstance_or_raise(
+    value: Any,
+    dtype: Union[Type, Tuple[Type, ...], None],
```
suzmue, line 1012: Where do we use this with dtype=None? If nowhere, can we remove it?

```suggestion
    dtype: Union[Type, Tuple[Type, ...]],
```

suzmue (commented):
Thanks for making the changes! I think this update looks really good.

google/cloud/bigquery/dataset.py
suzmue: It looks like this didn't get updated.

This looks like another case where the _isinstance_or_raise helper can be used as well.

I'm also a little confused with why this case raises a ValueError and the helper raises a TypeError, is there something different about the case here?

chalmerlowe: To answer your question: 

This is a moving target. Thus after this bit of code was made, we eventually migrated to using a function to handle the same/similar capability. But before going back and changing all the things, I felt I needed to get the `isinstance_or_raise()` function nailed down. Somewhere in that process we looked at what the appropriate error ought to be and I assert that it should be a TypeError (but mistakenly thought is was a ValueError when I started out).

I think `isinstance_or_raise()` is mostly nailed down now and can be used in future classes and can also replace older methods I created way back when this task first started.

sigh.

= /

tests/unit/test_dataset.py
```diff
@@ -1014,6 +1014,61 @@ def test_from_string_legacy_string(self):
...
             cls.from_string("string-project:string_dataset")
 
+    API_REPR = {
+        "datasetReference": {"projectId": "project", "datasetId": "dataset-id"},
+        "labels": {},
+        "externalCatalogDatasetOptions": {
+            "defaultStorageLocationUri": "gs://test-bucket/test-path",
+            "parameters": {"key": "value"},
+        },
+    }
```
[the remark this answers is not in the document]
suzmue: Changes LGTM

PR #2016, opened 2024-09-10
docs: improve QueryJobConfig.destination docstring

Linchin:
Improve the docstring to note that only table ID is sent to the backend, thus no configs in `Table` will be preserved.

See b/365471371 for more info.


chalmerlowe (approved):
LGTM

PR #2034, opened 2024-09-28
fix: avoid blocking in download thread when using BQ Storage API

### LINKED ISSUE #2032: Deadlock due to race condition in download thread when using Bigquery Storage API
There is a race condition in the download thread and the main thread when using Bigquery Storage API to fetch data.

https://github.com/googleapis/python-bigquery/blob/a76af359525ef3c49c958663f81fd24c9d35e1e7/google/cloud/bigquery/_pandas_helpers.py#L797-L801

https://github.com/googleapis/python-bigquery/blob/a76af359525ef3c49c958663f81fd24c9d35e1e7/google/cloud/bigquery/_pandas_helpers.py#L934-L942

When the download thread is blocked on `worker_queue.put(item)`, if the main thread exit, causing the pool to shut down, the download thread will be stuck. This behavior prevents the program from exiting.

kien-truong:
This prevents a deadlock between the main thead and download threads when the threadpool is shutdown prematurely.

Fixes #2032 🦕


kien-truong:
Just dropping by to see if there's any concerns about this PR.

chalmerlowe:
I took a quick look.
Gonna run our CI/CD checks on this.
My first thought is that because we are adding a couple code paths `coverage` is likely to fail.
Also curious why the pragma for `queue.full`: we recently got bitten by having a code branch that was not tested and I am hesitant about adding more such paths to the code base without understanding for myself why it is necessary/desireable.

kien-truong:
The `queue.Full` code path is hard to test because it's non-deterministic: the queue can be full or not depending on the timing of consumers and producers. 

So, I just put a `pragma` there to ignore the coverage, similar to the existing code a few lines below.

https://github.com/googleapis/python-bigquery/blob/aaec3ceb0746e80ce617ffb7c4baa844ed702abf/google/cloud/bigquery/_pandas_helpers.py#L968

tswast (approved):

tswast:
Looks like cover is passing.

tswast:
After sleeping on this, I do think it's important to try and cover the currently uncovered branch. I'll see what I can do in a follow-up PR today.

PR #2039, opened 2024-10-04
feat: updates to allow users to set max_stream_count

### LINKED ISSUE #2030: Make `max_stream_count` configurable when using Bigquery Storage API
Currently, for API that can use BQ Storage Client to fetch data like `to_dataframe_iterable` or `to_arrow_iterable`, the client library always uses the maximum number of read streams recommended by BQ server. 

https://github.com/googleapis/python-bigquery/blob/ef8e92787941ed23b9b2b5ce7c956bcb3754b995/google/cloud/bigquery/_pandas_helpers.py#L840

https://github.com/googleapis/python-bigquery/blob/ef8e92787941ed23b9b2b5ce7c956bcb3754b995/google/cloud/bigquery/_pandas_helpers.py#L854-L858

This behavior has the advantage of maximizing throughput but can lead to out-of-memory issue when there are too many streams being opened and result are not read fast enough: we've encountered queries that open hundreds of streams and consuming GBs of memory.

BQ Storage Client API also suggests capping `max_stream_count` when resource is constrained

https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#createreadsessionrequest

> Typically, clients should either leave this unset to let the system to determine an upper bound OR set this a size for the maximum "units of work" it can gracefully handle.

This problem has been encountered by others before and can be worked-around by monkey-patching the `create_read_session` on the BQ Client object: https://github.com/googleapis/python-bigquery/issues/1292

However, it should really be fixed by allowing the `max_stream_count` parameter to be set through public API.

chalmerlowe:
Adds a function `determine_requested_streams()` to compare `preserve_order` and the new argument `max_stream_count` to determine how many streams to request.

```
preserve_order (bool): Whether to preserve the order of streams. If True,
    this limits the number of streams to one (more than one cannot guarantee order).
max_stream_count (Union[int, None]]): The maximum number of streams
    allowed. Must be a non-negative number or None, where None indicates
    the value is unset. If `max_stream_count` is set, it overrides
    `preserve_order`.
```

Fixes #2030 🦕


kien-truong:
Shouldn't `preserve_order` take priority over `max_stream_count` for correctness ?

chalmerlowe:
@kien-truong 

In the question of which argument overrides which, I could probably be convinced to go either way.

I purposely chose for `max_streams_count` to `override preserve_order` and wrote the docstring to state that, because:
* `max_streams_count` defaults to `None` and any user whose in-house code currently makes the call to `_download_table_bqstorage()` but does NOT include `max_streams_count` but does have a value for `preserve_order` will see the same behavior. 
* If however, they go to the trouble of including a value for `max_streams_count` I would presume that they are doing so with an understanding of the new parameter and how it will affect any previous settings for `preserve_order` as defined in the docstring.

suzmue:
@chalmerlowe @kien-truong 

I am also in favor of having preserve_order override the max_streams_count setting. I think this would make it easier to use and understand.

If `max_streams_count >= 1` this condition would still be satisfied by setting requested_streams = 1 since `max_streams_count` isn't a guarantee of any number of streams even in the bq storage api and from the issue, it seems that an import use case is to limit the number of streams, not increase them (though I am sure there are use cases for that as well).

Perhaps an error for when `max_streams_count == 0 and preserve_order` would make sense to have if we swap to favoring preserve_order, since then they are explicitly requesting both unlimited streams and a single stream.

chalmerlowe:
let me cogitate on this. thanks for the input, both of you.

chalmerlowe:
@suzmue @kien-truong 

I revised the logic. Once the CI/CD checks pass, can you approve it, Suzy?

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -854,11 +853,11 @@ def _download_table_bqstorage(
...
-            concurrent streams to use for downloading data. If None, the
-            number of streams is determined automatically based on the
-            `preserve_order` parameter. If `max_stream_count` is set to a
-            non-negative value it will override values for `preserve_order`.
-            Defaults to None.
+            concurrent streams to use for downloading data. If `preserve_order`
+            is True, `max_stream_count` is ignored. If `preserve_order` is False
+            and `max_stream_count` is set to a non-negative integer, that number
+            of streams will be requested. If `max_stream_count` is None, then the
+            number of requested streams will be unbounded. Defaults to None.
```
suzmue, line 860: nit: I think this could be simplified. Maybe something like below. 
```suggestion
        max_stream_count (Optional[int]): The maximum number of
            concurrent streams to use for downloading data. If `preserve_order`
            is True, the requested streams are limited to 1 regardless of the 
            `max_stream_count` value. If 0 or None, then the number of
            requested streams will be unbounded. Defaults to None.
```

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -1087,26 +1086,31 @@ def determine_requested_streams(
...
         if max_stream_count <= -1:
             raise ValueError("max_stream_count must be non-negative OR None")
         return max_stream_count
-    elif preserve_order:
-        # If max_stream_count is unset but preserve_order is set,
-        # use 1 (to limit the max_stream_count to 1, to ensure that order
-        # is preserved)
-        return 1
+
     else:
```
suzmue, line 1113: preference: drop else

Since all branches return, prefer to drop the extra else and just end with the default return value of 0.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -1087,26 +1086,31 @@ def determine_requested_streams(
...
-        # If both max_stream_count and preserve_order are unset, use 0 (unbounded)
+        # When preserve_order is False and max_stream_count is None, # of
+        # requested streams is zero (unbounded)
```
suzmue, line 1115: nit: probably not necessary to explain above if statements. 

```suggestion
        # Default to zero requested streams (unbounded).
```

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -837,18 +880,22 @@ def _download_table_bqstorage(
     if "@" in table.table_id:
         raise ValueError("Reading from a specific snapshot is not currently supported.")
 
-    requested_streams = 1 if preserve_order else 0
+    # Compares preserve_order vs max_stream_count to determine how many
+    # streams to use.
```
suzmue, line 884: nit: this comment doesn't add all that much to the readability of the code, so I would just drop it (preserve_order and max_stream_count are being used as arguments to a well named function, which implies to the reader they are used to determine the number of requested streams).

```suggestion
```

suzmue (approved):

PR #2044, opened 2024-10-21
feat: add type hints to Client

rinarakaki:


google/cloud/bigquery/client.py
```diff
@@ -231,15 +230,23 @@ class Client(ClientWithProject):
...
+        client_options: Optional[
+            Union[google.api_core.client_options.ClientOptions, Dict[str, Any]]
+        ] = None,
     ) -> None:
+        if client_options is None:
+            client_options = {}
+        if isinstance(client_options, dict):
+            client_options = google.api_core.client_options.from_dict(client_options)
+        assert isinstance(client_options, google.api_core.client_options.ClientOptions)
+
```
chalmerlowe, line 249: I am curious, why did we choose to:

* move this content from after the call to `super()` to before the call to `super()`? What is the reason for reordering this content?
* add an assert statement when one was not used previously?

rinarakaki: - Moving this part before the call to `super()` is optional, but I did it because actually the exact same type check & coercion happens inside the `super()` (`Client.__init__`) as well, so the I let it early break in the second time.
- The reason for adding the assert statement is to let the type checker know the type of `client_options`, otherwise it wouldn't know that the `if client_options.api_endpoint:` expression at the line 261 is valid .

chalmerlowe:
@rinarakaki I ran the kokoro CI/CD checks.
There are a number of things that are going wrong both in the standard tests and in mypy.
I spent as much time as I could today trying to trouble shoot the issues, to no avail.
I have other tasks that require my attention. If you are able to figure out why these tests fail I am willing to come back and take another look at this.

#### mypy errors:
```
google/cloud/bigquery/client.py:265: error: Incompatible types in assignment
   (expression has type "str", target has type "Optional[ClientInfo]")  [assignment]
google/cloud/bigquery/client.py:269: error: Incompatible types in assignment
   (expression has type "str", target has type "Optional[ClientInfo]")  [assignment]
```

#### unit test errors:
```ImportError while loading conftest '/tmpfs/src/github/python-bigquery/tests/unit/conftest.py'.
tests/unit/conftest.py:19: in 
    from .helpers import make_client
tests/unit/helpers.py:19: in 
    import google.cloud.bigquery.client
google/cloud/bigquery/__init__.py:36: in 
    from google.cloud.bigquery.client import Client
google/cloud/bigquery/client.py:185: in 
    class Client(ClientWithProject):
google/cloud/bigquery/client.py:235: in Client
    _http: Optional[requests.Session] = None,
E   NameError: name 'requests' is not defined
```

google/cloud/bigquery/client.py
```diff
@@ -133,8 +134,7 @@
 if typing.TYPE_CHECKING:  # pragma: NO COVER
     # os.PathLike is only subscriptable in Python 3.9+, thus shielding with a condition.
     PathType = Union[str, bytes, os.PathLike[str], os.PathLike[bytes]]
-    import requests  # required by api-core
-
+
```
chalmerlowe, line 137:
```suggestion
```

google/cloud/bigquery/client.py
```diff
@@ -122,9 +125,7 @@
 from google.cloud.bigquery.table import RowIterator
 
 pyarrow = _versions_helpers.PYARROW_VERSIONS.try_import()
-pandas = (
-    _versions_helpers.PANDAS_VERSIONS.try_import()
-)  # mypy check fails because pandas import is outside module, there are type: ignore comments related to this
+pandas = _versions_helpers.PANDAS_VERSIONS.try_import()  # mypy check fails because pandas import is outside module, there are type: ignore comments related to this
```
chalmerlowe, line 128:
```suggestion
pandas = (
    _versions_helpers.PANDAS_VERSIONS.try_import()
)  # mypy check fails because pandas import is outside module, there are type: ignore comments related to this

```

chalmerlowe (approved):
LGTM

PR #2051, opened 2024-11-03
feat: support setting max_stream_count when fetching query result

### LINKED ISSUE #2030: Make `max_stream_count` configurable when using Bigquery Storage API

kien-truong:
Allow user to set max_stream_count when fetching result using BigQuery Storage API with incremental methods:

* to_arrow_iterable
* to_dataframe_iterable

Fixes #2030 🦕

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1842,6 +1842,10 @@ class RowIterator(HTTPIterator):
                 using BigQuery Storage API. Ignored if
                 BigQuery Storage API is not used.
 
+                This setting also has no effect if the query result
+                is deterministically ordered with ORDER BY,
+                in which case, the number of download stream is always 1.
+
                 By default, this value is unnset, and the number of download
                 streams is determined by BigQuery the server. However, this behaviour
                 can require a lot of memory to store temporary download result,
@@ -2026,6 +2030,10 @@ class RowIterator(HTTPIterator):
                 The maximum number of parallel download streams when
                 using BigQuery Storage API. Ignored if
                 BigQuery Storage API is not used.
+
+                This setting also has no effect if the query result
+                is deterministically ordered with ORDER BY,
+                in which case, the number of download stream is always 1.
 
                 By default, this value is unnset, and the number of download
                 streams is determined by BigQuery the server. However, this behaviour
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1846,11 +1846,12 @@ class RowIterator(HTTPIterator):
                 is deterministically ordered with ORDER BY,
                 in which case, the number of download stream is always 1.
 
-                By default, this value is unnset, and the number of download
+                If set to 0 or None (the default), the number of download
                 streams is determined by BigQuery the server. However, this behaviour
                 can require a lot of memory to store temporary download result,
-                especially with very large queries. If that's the case,
-                this parameter can be used to reduce memory consumption.
+                especially with very large queries. In that case,
+                setting this parameter value to a value > 0 can help
+                reduce system resource consumption.
 
         Returns:
             pyarrow.RecordBatch:
@@ -2035,11 +2036,12 @@ class RowIterator(HTTPIterator):
                 is deterministically ordered with ORDER BY,
                 in which case, the number of download stream is always 1.
 
-                By default, this value is unnset, and the number of download
+                If set to 0 or None (the default), the number of download
                 streams is determined by BigQuery the server. However, this behaviour
                 can require a lot of memory to store temporary download result,
-                especially with very large queries. If that's the case,
-                this parameter can be used to reduce memory consumption.
+                especially with very large queries. In that case,
+                setting this parameter value to a value > 0 can help
+                reduce system resource consumption.
 
         Returns:
             pandas.DataFrame:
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/table.py
```diff
@@ -1836,6 +1837,17 @@ def to_arrow_iterable(
                 created by the server. If ``max_queue_size`` is :data:`None`, the queue
                 size is infinite.
 
+            max_stream_count (Optional[int]):
```
Linchin, line 1840: I think it would be more consistent if we use the same docstring as [here](https://github.com/googleapis/python-bigquery/blob/a4d9534a900f13ae7355904cda05097d781f27e3/google/cloud/bigquery/_pandas_helpers.py#L1083-L1087). It also mentions the effect of `preserve_order` (in this case `self._preserve_order`), which I think we should make clear here.

google/cloud/bigquery/table.py
```diff
@@ -2008,6 +2022,17 @@ def to_dataframe_iterable(
 
                 .. versionadded:: 2.14.0
 
+            max_stream_count (Optional[int]):
```
Linchin, line 2025: same as https://github.com/googleapis/python-bigquery/pull/2051/files#r1836568949

Linchin:
Thank you @kien-truong for adding further support for `max_stream_count`! I left some comments, and I wonder if you could add some unit tests too (we typically require 100% unit test coverage). You could find similar tests in PR #2039. If you need any help with it, feel free to let me know. I'd be more than glad to help!

kien-truong:
Hi, the default code path with the default arguments is already covered by the current tests. 
I'll see if I can add tests for when the user manually set the argument when I have time at weekend.

Linchin:
@kien-truong sounds good. The mypy test is also failing, could you fix it too? You can run it by running `nox -s mypy` in the repo. (You can install nox by `pip install nox`)

kien-truong:
I have added tests to cover user-provided `max_stream_count` flow and fixed the `mypy` test as well.

google/cloud/bigquery/table.py
kien-truong: Updated

tests/unit/test_table.py
```diff
@@ -5822,3 +5822,75 @@ def test_table_reference_to_bqstorage_v1_stable(table_path):
...
+        ],
+        table=mut.TableReference.from_string("proj.dset.tbl"),
+    )
+    row_iterator._preserve_order = preserve_order
+
+    max_stream_count = 132
+    result_iterable = row_iterator.to_arrow_iterable(
+        bqstorage_client=bqstorage_client, max_stream_count=max_stream_count
+    )
+    for _ in result_iterable:  # pragma: NO COVER
```
Linchin, line 5854: Is there a specific reason why we are ignoring test coverage here?

tests/unit/test_table.py
```diff
@@ -5822,3 +5822,75 @@ def test_table_reference_to_bqstorage_v1_stable(table_path):
...
+        ],
+        table=mut.TableReference.from_string("proj.dset.tbl"),
+    )
+    row_iterator._preserve_order = preserve_order
+
+    max_stream_count = 132
+    result_iterable = row_iterator.to_dataframe_iterable(
+        bqstorage_client=bqstorage_client, max_stream_count=max_stream_count
+    )
+    for _ in result_iterable:  # pragma: NO COVER
```
Linchin, line 5890: Same as https://github.com/googleapis/python-bigquery/pull/2051/files#r1849076123

Linchin:
Thanks @kien-truong, the PR mostly looks good. I just have some small question regarding ignored coverage for the tests.

tests/unit/test_table.py
kien-truong: Because we mocked `create_read_session` to return an empty ReadSession, `result_iterable` is always empty and the loop is never entered.

Linchin: I see, I will use `list(result_iterable)` so we don't have to skip coverage, just in case.

Linchin (approved):
LGTM. Thank you!

PR #2055, opened 2024-11-07
docs: render fields correctly for update calls

leahecole:
Fixes internal bug 376692000

Code blocks don't render properly in tables on the docs site, so I moved the examples up in the description, where they do! 😄


chalmerlowe (approved):
LGTM

PR #2074, opened 2024-11-19
feat: adds two from_api_tests: StorageDesc & SerDeInfo

chalmerlowe:
Adds a test to each of the test sets for:

StorageDescriptor: test_from_api_repr
SerDeInfo: test_from_api_repr

### DIFF
--- a/tests/unit/test_schema.py
+++ b/tests/unit/test_schema.py
@@ -1161,6 +1161,23 @@ class TestForeignTypeInfo:
         assert result.to_api_repr() == expected
 
 
+@pytest.fixture
+def _make_storage_descriptor():
+    serdeinfo = SerDeInfo(
+        serialization_library="testpath.to.LazySimpleSerDe",
+        name="serde_lib_name",
+        parameters={"key": "value"},
+    )
+
+    obj = StorageDescriptor(
+        input_format="testpath.to.OrcInputFormat",
+        location_uri="gs://test/path/",
+        output_format="testpath.to.OrcOutputFormat",
+        serde_info=serdeinfo,
+    )
+    return obj
+
+
 class TestStorageDescriptor:
     """Tests for the StorageDescriptor class."""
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_schema.py
```diff
@@ -1315,4 +1359,26 @@ def test_to_api_repr(self):
...
+        }
+        
+        expected = self._make_one(
+            serialization_library="testpath.to.LazySimpleSerDe",
+            name="serde_name",
+            parameters={"key": "value"},
+        )
+                
+        klass = self._get_target_class()
+        result = klass.from_api_repr(resource)
```
chalmerlowe, line 1382:
```suggestion
        resource = {
            "serializationLibrary": "testpath.to.LazySimpleSerDe",
            "name": "serde_name",
            "parameters": {"key": "value"},
        }

        expected = self._make_one(
            serialization_library="testpath.to.LazySimpleSerDe",
            name="serde_name",
            parameters={"key": "value"},
        )

        klass = self._get_target_class()
        result = klass.from_api_repr(resource)
```

tests/unit/test_schema.py
```diff
@@ -1251,7 +1268,34 @@ def test_to_api_repr(self):
...
+        directly (via the fixture) and displayed as a dict.
+        """
+        # generate via fixture
+        expected = _make_storage_descriptor
+        resource = self.API_REPR
+        klass = self._get_target_class()
+        # generate via API_REPR
+        result = klass.from_api_repr(resource)
+
+        assert result.to_api_repr() == expected.to_api_repr()
```
suzmue, line 1298: It feels a little strange to have this test, which is intending to test `from_api_repr`, instead comparing the results of to_api_repr.

Could you write a comparison that checks that the result and expected match more directly?

suzmue (approved):

tests/unit/test_schema.py
suzmue: On second thought, as I noted in github.com/googleapis/python-bigquery/pull/2076 since to_api_repr is tested I think this is actually ok.

PR #2076, opened 2024-11-22
feat: adds ForeignTypeInfo test

chalmerlowe:
Adds a test to the test sets for ForeignTypeInfo.
* from_api_repr

Also removed `class TableSchema` as unneeded.

### DIFF
--- a/tests/unit/test_schema.py
+++ b/tests/unit/test_schema.py
@@ -19,7 +19,7 @@ from google.cloud.bigquery.schema import (
     ForeignTypeInfo,
     StorageDescriptor,
     SerDeInfo,
-    TableSchema,
+    # TODO: delete this line...TableSchema,
 )
 
 import unittest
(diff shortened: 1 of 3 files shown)


tests/unit/test_schema.py
```diff
@@ -1160,6 +1158,22 @@ def test_to_api_repr(self, type_system, expected):
...
+        directly (via _make_one()) and displayed as a dict.
+        """
+        resource = {"typeSystem": "TYPE_SYSTEM_UNSPECIFIED"}
+
+        expected = self._make_one(type_system="TYPE_SYSTEM_UNSPECIFIED")
+
+        klass = self._get_target_class()
+        result = klass.from_api_repr(resource)
+
+        assert result.to_api_repr() == expected.to_api_repr()
```
suzmue, line 1175: Similar feedback to the other CL, though maybe though now I'm thinking maybe we need to test that from_api_repr has the same response, though since to_api_repr is tested I think this is actually ok. 

Your doc comment does describe that the results we are comparing are displayed as a dict, maybe its just the test naming that is throwing it off for me (maybe test_convert_api_repr or something?). Either way do think its fine to leave as is.

suzmue (approved):

PR #2081, opened 2024-11-27
feat: adds Schema class and modifies schema handling

chalmerlowe:
This PR adds a `Schema` class to `python-bigquery` as part of the Google internal PANGEA work.

Historically, `table.schema` has been a `list` of `SchemaFields`.

`Schema` objects will now include not only a `list` of fields, but will have an attribute called `foreign_type_info`.

The aim of the work here is create, as much as possible, a drop-in replacement for `schema`.

The end user should not need to know whether they are looking at a list-based `schema` OR a class Schema-based object unless they need access to the `foreign_type_info` attr.

They should still be able to:
* iterate over a `Schema` just like you would a `list`
* index and slice the `Schema` just like a list

For users who already have a `list` in their code, nothing should break.

For future users who will need a `Schema` object, they should be able to instantiate one and it should just work.

TODO:
 - [ ] add docstrings
 - [ ] add type hints
 - [ ] remove or modify comment strings in the code to help future maintainers


google/cloud/bigquery/schema.py
```diff
@@ -921,3 +954,57 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+        del self._fields[index]
+
+    def __iter__(self):
+        return iter(self._fields)
+
+    def __str__(self):
+        return str(self._fields)
+
+    def __repr__(self):
+        return f"Schema({self.foreign_type_info!r}, {self._properties['_fields']!r})"
```
chalmerlowe, line 1004:
```suggestion
        return f"Schema({self.foreign_type_info!r}, {self._fields!r})"
```

google/cloud/bigquery/client.py
```diff
@@ -3680,7 +3680,7 @@ def insert_rows(
         if selected_fields is not None:
             schema = selected_fields
 
-        if len(schema) == 0:
+        if not schema:
```
tswast, line 3683: I'm curious, could you make `len(schema_object)` work? I guess this is to allow `None` as `schema` now, too? If `table.schema` could indeed return `None` now, that might be considered a breaking change, depending on how strict we want to be about types.

Either way, folks might have to change their code to deal with missing schema, so something to consider. Maybe we return an object of `len(...) == 0` for now and have a warning that it will be changed to `None` in future?

google/cloud/bigquery/client.py
```diff
@@ -4029,7 +4029,7 @@ def list_rows(
 
         # No schema, but no selected_fields. Assume the developer wants all
         # columns, so get the table resource for them rather than failing.
-        elif len(schema) == 0:
+        elif not schema:
```
tswast, line 4032: Same here. Let's be careful about breaking changes regarding `table.schema` allowing `None`.

google/cloud/bigquery/schema.py
```diff
@@ -549,33 +549,29 @@ def _build_schema_resource(fields):
...
-            :meth:`~google.cloud.bigquery.schema.SchemaField.from_api_repr`.
-
-    Returns:
-        Sequence[:class:`~google.cloud.bigquery.schema.SchemaField`]
-
-    Raises:
-        Exception: If ``schema`` is not a sequence, or if any item in the
-        sequence is not a :class:`~google.cloud.bigquery.schema.SchemaField`
-        instance or a compatible mapping representation of the field.
+    """TODO docstring
```
tswast, line 552: Flagging with a comment just so we don't lose track of this TODO.

google/cloud/bigquery/schema.py
```diff
@@ -796,8 +792,6 @@ def serde_info(self) -> Any:
         prop = _get_sub_prop(self._properties, ["serDeInfo"])
         if prop is not None:
             prop = StorageDescriptor().from_api_repr(prop)
-            print(f"DINOSAUR prop: {prop}")
```
tswast, line 799: Lol, oops! 😆 Could be worth a separate cleanup PR to have something small and easy to approve.

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+        """TODO: docstring"""
+        return self._properties.get("foreignTypeInfo")
+
+    @foreign_type_info.setter
+    def foreign_type_info(self, value: str) -> None:
+        value = _isinstance_or_raise(value, str, none_allowed=True)
+        self._properties["foreignTypeInfo"] = value
+
+    @property
+    def _fields(self) -> Any:
```
tswast, line 938: Why private? I guess because we're trying to emulate a list and want to support mutation that way?

Would be worth an explanation.

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
         config = cls("")
         config._properties = copy.deepcopy(resource)
         return config
+
+
+class Schema:
+    # TODO docstrings and type hints
+    def __init__(self, fields=None, foreign_type_info=None):
+        self._properties = {}
+        self._fields = [] if fields is None else list(fields)  # Internal List
```
tswast, line 924: Do we need to convert the items of `fields` from `SchemaField` into dictionaries here? Ideally `_properties` would directly reflect the REST representation.

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
         config = cls("")
         config._properties = copy.deepcopy(resource)
         return config
+
+
+class Schema:
+    # TODO docstrings and type hints
+    def __init__(self, fields=None, foreign_type_info=None):
```
tswast, line 922: Presumably we have these `fields` and `foreign_type_info` arguments because we expect users to be constructing `Schema` objects directly? If so, accepting an Iterable of `SchemaField` or `dict` would be the most user-friendly IMO.

google/cloud/bigquery/table.py
```diff
@@ -452,17 +453,20 @@ def schema(self):
                 instance or a compatible mapping representation of the field.
         """
         prop = self._properties.get(self._PROPERTY_TO_API_FIELD["schema"])
-        if not prop:
+        if not prop:  # if empty Schema, empty list, None
             return []
-        else:
-            return _parse_schema_resource(prop)
+        elif isinstance(prop, Schema):
```
tswast, line 458: This would be concerning if this were ever true. `_properties` should always be equivalent to the `REST` representation.

google/cloud/bigquery/table.py
```diff
@@ -452,17 +453,20 @@ def schema(self):
...
+            return prop
+        return _parse_schema_resource(prop)
```
tswast, line 460: Presumably we want to return a `Schema` object now. 

```suggestion
        return Schema.from_api_repr(prop)
```

Aside: Is `_parse_schema_resource` dead code? I assume it still returns a list of `SchemaField`?

google/cloud/bigquery/table.py
```diff
@@ -452,17 +453,20 @@ def schema(self):
...
 
     @schema.setter
     def schema(self, value):
         api_field = self._PROPERTY_TO_API_FIELD["schema"]
 
         if value is None:
             self._properties[api_field] = None
+        elif isinstance(value, Schema):
+            self._properties[api_field] = value
```
tswast, line 469:
```suggestion
            self._properties[api_field] = value.to_api_repr()
```

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+        self._fields.extend(iterable)
+
+    def to_api_repr(self) -> dict:
+        """Build an API representation of this object.
+
+        Returns:
+            Dict[str, Any]:
+                A dictionary in the format used by the BigQuery API.
+        """
+        return copy.deepcopy(self._properties)
```
tswast, line 981: It might be a bit wasteful to make deepcopy here and in `from_api_repr`. Indeed it's safer, but could add a lot of overhead. IIRC we actually removed some `deepcopy` calls from `SchemaField` because it was slowing down customers who build dynamic schemas in their code.

See: https://github.com/googleapis/python-bigquery/issues/6 and https://github.com/googleapis/python-bigquery/pull/26

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+
+        Args:
+            resource (Dict[str, Any]):
+                API representation of the object to be instantiated.
+
+        Returns:
+            An instance of the class initialized with data from 'resource'.
+        """
+        config = cls("")
+        config._properties = copy.deepcopy(resource)
```
tswast, line 996: Same here. May be better to omit the deepcopy.

google/cloud/bigquery/table.py
```diff
@@ -452,17 +453,20 @@ def schema(self):
                 instance or a compatible mapping representation of the field.
         """
         prop = self._properties.get(self._PROPERTY_TO_API_FIELD["schema"])
-        if not prop:
+        if not prop:  # if empty Schema, empty list, None
             return []
```
tswast, line 457: Could we return an "empty" Schema here for consistency?

```suggestion
            return Schema()
```

tswast (changes requested):

tswast:
Thanks! Overall goals outlined in the description look good to me.

> The aim of the work here is create, as much as possible, a drop-in replacement for schema.
> 
> The end user should not need to know whether they are looking at a list-based schema OR a class Schema-based object unless they need access to the foreign_type_info attr.
> 
> They should still be able to:
> 
> * iterate over a Schema just like you would a list
> * index and slice the Schema just like a list
>
> For users who already have a list in their code, nothing should break.
> 
> For future users who will need a Schema object, they should be able to instantiate one and it should just work.

google/cloud/bigquery/schema.py
chalmerlowe: You are correct, the idea behind having it be private was to try and emulate a list.
I can add a note to that effect when I add docstrings.

google/cloud/bigquery/table.py
chalmerlowe: `_parse_schema_resource` still returns a list of `SchemaFields` for backwards compatibility.

Where possible I *-tried-* to stay consist about what the user provides versus gets back: if someone handed the library a `list` of `SchemaField`s (i.e. cause that is what they are used to) then we maintained that state. I tried to avoid randomly converting `lists` into `Schema` objects to avoid user surprise.

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+    def _fields(self) -> Any:
+        """TODO: docstring"""
+        return self._properties.get("_fields")
+
+    @_fields.setter
+    def _fields(self, value: list) -> None:
+        value = _isinstance_or_raise(value, list, none_allowed=True)
+        self._properties["_fields"] = value
+
+    def __len__(self):
```
Linchin, line 947: I feel slightly concerned about our way of emulating list using class `Schema` here. There are many other methods that the [built-in list supports](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists), such as `insert()`, `remove()`, `pop()` and so on, and if any of our customer has been using these, this PR could end up a breaking change. Using a new class could also imply we will need to maintain a full-blown python list type in the future. I think inheriting `list` might make it easier for us to maintain?

google/cloud/bigquery/client.py
chalmerlowe: I was under the impression that `Table.schema` can currently be set to `None`. It appears that we allow for that in the existing code:

https://github.com/googleapis/python-bigquery/blob/d4d39acb8574f0d06d4e490b859e5fe6b57d0d9e/google/cloud/bigquery/table.py#L458C1-L466C84

```python
@schema.setter
def schema(self, value):
    api_field = self._PROPERTY_TO_API_FIELD["schema"]
    if value is None:
        self._properties[api_field] = None
    else:
        value = _to_schema_fields(value)
        self._properties[api_field] = {"fields": _build_schema_resource(value)}
```

chalmerlowe: I switched it back to the old check.

google/cloud/bigquery/client.py
chalmerlowe: I also switched this back to the old check.

google/cloud/bigquery/schema.py
chalmerlowe: Done.

google/cloud/bigquery/schema.py
chalmerlowe: Removed.

google/cloud/bigquery/schema.py
tswast: Resolving. I missed that `_fields` was a property with a setter. I think your code does actually convert Iterable of SchemaField or dict to the correct API representation in _properties

google/cloud/bigquery/schema.py
tswast: Resolving. I missed that _fields was a property with a setter. I think your code does actually convert Iterable of SchemaField or dict to the correct API representation in _properties

google/cloud/bigquery/schema.py
```diff
@@ -921,3 +915,83 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+
+    @property
+    def _fields(self) -> Any:
+        """TODO: docstring"""
+        return self._properties.get("_fields")
+
+    @_fields.setter
+    def _fields(self, value: list) -> None:
+        value = _isinstance_or_raise(value, list, none_allowed=True)
+        self._properties["_fields"] = value
```
tswast, line 945: Why `_fields`. It's `fields` in the REST API. Also, this should convert to the list to API representation. https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableSchema

chalmerlowe: Changed `_fields` to `fields`. Changed tests to match.

google/cloud/bigquery/table.py
chalmerlowe: Along the lines of least surprise... if the user passes in an empty `list` or `None`, they will get back an empty `list`. If they pass in an empty `Schema()` they will get back an empty `Schema()`

google/cloud/bigquery/table.py
chalmerlowe: @tswast Question for you.

This response relates to setting the `schema` on the table as a whole and thus applies to several comments in this review:

In general, I see the reasoning behind maintaining the `_properties` attribute as a dict that is pretty close to the API resource representation.

If the user provides a `Schema` object based on the `Schema` class, does it make sense to maintain this as a `Schema` object until we need to explicitly convert it back into an API resource using `.to_api_repr()`?

tswast: If so, I wouldn't put it in `_properties`.

Ideally, the objects we provide would just be a "view" on top of the REST resource (`_properties`). I wouldn't put anything that requires conversion before sending it to the BigQuery API into such objects.

In some cases, we have to store state outside of `_properties`, in which case we populate the necessary dictionaries and such in `to_api_repr()`. The `UserList` approach may be such a case. If so, then don't store the backing list of `SchemaField` in `_properties`. Store it in some other instance variable.

google/cloud/bigquery/table.py
chalmerlowe: @tswast 

Does the following influence your thoughts on this comment?

Historically, we have returned an empty `list`.
An existing user would still expect a `list`.
A new user should probably not see an empty `Schema` object, since they are only needed if the user is gonna try to access an external data source (where they have to identify the `foreign_type_info`).

My general principal, focused on backwards compatibility is:
* avoid as many changes to what the code did before, unless necessary
* as much as possible, force the choice to move to Schema to be an explicit decision by any user (especially since external data sources are likely to be the exception, not the rule).

tswast: I could go either way on this one. One one hand, using `Schema` everywhere would make the new Schema class more discoverable and fewer cases to consider when type checking. On the other hand, providing a real list in all cases where `Schema` is not needed would provide the most backwards compatible option.

On further consideration, I agree with your assessment that avoiding breaking existing users is preferable. +1 on empty list in this case.

google/cloud/bigquery/schema.py
```diff
@@ -886,3 +903,131 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+        Returns:
+            Dict[str, Any]:
+                A dictionary in the format used by the BigQuery API.
+                If the schema contains SchemaField objects, the fields are
+                also converted to their API representations.
+        """
+        # If this is a RECORD type, then sub-fields are also included,
+        # add this to the serialized representation.
+        answer = self._properties.copy()
+        if self._fields == []:
```
tswast, line 1012: This would have to be updated if we remove `"fields"` from `_properties`, as I suggest above.

Also, per go/pystyle#truefalse-evaluations-decision it is preferred to use the falsey nature of empty list if possible.

```suggestion
        if not self.data:
```

google/cloud/bigquery/schema.py
```diff
@@ -886,3 +903,131 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+            return answer
+        schemafields = any([isinstance(f, SchemaField) for f in self._fields])
+        if schemafields:
+            answer["fields"] = [f.to_api_repr() for f in self._fields]
```
tswast, line 1016: This seems wrong. We should allow people to insert `Mapping` / `dict` into their Schema / UserList objects. Even if no remaining field is a `SchemaField`, we should still return something.

I would expect something more like this:

```suggestion
        answer["fields"] = [
            # Users may inject the REST representation of schema fields into the list.
            # If it's not a `SchemaField` object, then we can assume that's what they've
            # done.
            field.to_api_repr() if hasattr(field, "to_api_repr") else field
            for field in self.data
        ]
```

I'd also like to see some tests where the user replaces all fields with the REST representation of different fields. Something like:

```python
def test_schema_to_api_repr_with_rest_entries():
    schema = Schema(fields=[SchemaField(), ....])
    del schema[:]
    rest_fields = [{"name": "...", "type": "..."}, ...]
    schema.extend(rest_fields)
    got = schema.to_api_repr()
    assert got = {"fields": rest_fields}
```

google/cloud/bigquery/schema.py
```diff
@@ -886,3 +903,131 @@ def from_api_repr(cls, resource: dict) -> SerDeInfo:
...
+
+        Args:
+            resource (Dict[str, Any]):
+                API representation of the object to be instantiated.
+
+        Returns:
+            An instance of the class initialized with data from 'resource'.
+        """
+        config = cls([])
+        config._properties = copy.copy(resource)
```
tswast, line 1032: With using `UserList`, you will need to pull out `fields` and pass them into the constructor.

```suggestion
        resource = copy.copy(resource)
        fields = resource.pop("fields", [])
        config = cls(fields)
        config._properties = resource
```

This makes sure `.data` has all `SchemaField` objects while preserving the future-proof ability to add `_properties` (other than `fields`).

google/cloud/bigquery/table.py
tswast: This comment still stands. We shouldn't have `Schema` objects in `_properties`. Instead, check for the existence of any keys besides `"fields"` to determine if `Schema` or list should be used.

tswast (changes requested):

google/cloud/bigquery/table.py
chalmerlowe: @tswast 

this is a related concept.
Right now, if someone gives me a Schema object, I have been storing that in _properties and I see your point about avoiding that.

That complicates getting a schema object back out, if/when you need it.

What are your thoughts on this (I seem to recall you mention keeping objects in a separate place and ensuring that _properties is a view into the object)

```

    @schema.setter
    def schema(self, value):
        api_field = self._PROPERTY_TO_API_FIELD["schema"]

        if value is None:
            self._properties[api_field] = None
        elif isinstance(value, Schema):
            # store original as selt.<original_schema_object> # TODO find a better name
            self.<original_schema_object> = value
            self.properties[api_field] = self.<original_schema_object>.to_api_repr()
        else:
            value = _to_schema_fields(value)
            self._properties[api_field] = {"fields": _build_schema_resource(value)}


    @property
    def schema(self):
        # if user stored a Schema object, short circuit and return early.
        <original_schema_object> = self.get(<original_schema_object>, None)
        if isinstance(original_schema_object, Schema):
            return self.<original_schema_object>

        # if user did not store a Schema object, behave as normal
        prop = self._properties.get(self._PROPERTY_TO_API_FIELD["schema"])
        if not prop: # either None OR empty list  
            return []
        
        return _parse_schema_resource(prop)
```

PR #2093, opened 2024-12-21
feat: support resource_tags for table

### LINKED ISSUE #2089: Support resource_tags for Table

Currently, Python Client for Google BigQuery does not support `resource_tags` in `Table` class. This limitation makes it challenging to programmatically manage resource tags.

Add support for `resource_tags` in `Table` class, enabling users to set, update, and retrieve tags using the client library, similar to how other metadata fields are managed.

hrkh:

Fixes #2089 🦕

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1030,7 +1030,7 @@ class Table(_TableBase):
 
         See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.resource_tags
         """
-        return self._properties.get(
+        return self._properties.setdefault(
             self._PROPERTY_TO_API_FIELD["resource_tags"], {}
         )
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1033,7 +1033,7 @@ class Table(_TableBase):
         Raises:
             ValueError: For invalid value types.
         """
-        return self._properties.setdefault(
+        return self._properties.get(
             self._PROPERTY_TO_API_FIELD["resource_tags"], {}
         )
 
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -408,6 +408,7 @@ class Table(_TableBase):
         "require_partition_filter": "requirePartitionFilter",
         "table_constraints": "tableConstraints",
         "max_staleness": "maxStaleness",
+        "resource_tags": "resourceTags",
     }
 
     def __init__(self, table_ref, schema=None) -> None:
@@ -1022,6 +1023,25 @@ class Table(_TableBase):
         if table_constraints is not None:
             table_constraints = TableConstraints.from_api_repr(table_constraints)
         return table_constraints
+
+    @property
+    def resource_tags(self):
+        """Dict[str, str]: Resource tags for the table.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.resource_tags
+
+        Raises:
+            ValueError: For invalid value types.
+        """
+        return self._properties.setdefault(
+            self._PROPERTY_TO_API_FIELD["resource_tags"], {}
+        )
+
+    @resource_tags.setter
+    def resource_tags(self, value):
+        if not isinstance(value, dict):
+            raise ValueError("resource_tags must be a dict")
+        self._properties[self._PROPERTY_TO_API_FIELD["resource_tags"]] = value
 
     @classmethod
     def from_string(cls, full_table_id: str) -> "Table":
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1029,9 +1029,6 @@ class Table(_TableBase):
         """Dict[str, str]: Resource tags for the table.
 
         See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.resource_tags
-
-        Raises:
-            ValueError: For invalid value types.
         """
         return self._properties.get(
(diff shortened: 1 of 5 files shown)


Linchin:
Sorry, this seems to be a duplicate of #2090.

hrkh:
Hi @Linchin ,
Thank you for your feedback.

This PR was closed as a duplicate of #2090, but I’d like to clarify that #2090 addresses the resource tags for datasets, while this PR focuses on the resource tags for tables.

Could you please reopen this PR for further review? Let me know if additional clarification is needed.

Linchin:
Thank you @hrkh for clarifying! I have reopened and will review this PR.

google/cloud/bigquery/table.py
```diff
@@ -1023,6 +1024,25 @@ def table_constraints(self) -> Optional["TableConstraints"]:
...
+    @property
+    def resource_tags(self):
+        """Dict[str, str]: Resource tags for the table.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.resource_tags
+
+        Raises:
+            ValueError: For invalid value types.
+        """
+        return self._properties.setdefault(
```
Linchin, line 1036:
```suggestion
        return self._properties.get(
```

Use `get()` just for consistency.

google/cloud/bigquery/table.py
```diff
@@ -1023,6 +1024,25 @@ def table_constraints(self) -> Optional["TableConstraints"]:
...
         return table_constraints
 
+    @property
+    def resource_tags(self):
+        """Dict[str, str]: Resource tags for the table.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.resource_tags
+
+        Raises:
+            ValueError: For invalid value types.
```
Linchin, line 1034:
```suggestion
```
I don't think the exception applies here.

tests/unit/test_client.py
```diff
@@ -2369,14 +2371,16 @@ def test_update_table(self):
             "description": description,
             "friendlyName": title,
             "labels": {"x": "y"},
+            "resourceTags": {"123456789012/key": "value"},
         }
         conn.api_request.assert_called_once_with(
-            method="PATCH", data=sent, path="/" + path, timeout=7.5
+            method="PATCH", path="/" + path, data=sent, timeout=7.5
```
Linchin, line 2377:
```suggestion
            method="PATCH", data=sent, path="/" + path, timeout=7.5
```

Linchin:
Hi @hrkh, I think you need to fix the unit test, and need to do something to fix the system test (similar to [this](https://github.com/googleapis/python-bigquery/pull/2090/commits/7e75599b387e0de323c6c205543613d4f8b66872)). If you need any help, feel free to let me know, I'd be glad to help!

google/cloud/bigquery/table.py
hrkh: Hi @Linchin, thank you for your review.
I have a question regarding `resource_tags` that I'd like to discuss further.

Given that `resource_tags` is a dictionary, similar to `labels`, should we consider a similar implementation approach?

https://github.com/googleapis/python-bigquery/blob/9c504186f03ffb8b86836c683912b310db2435e9/google/cloud/bigquery/table.py#L1220-L1228

Using `get()` with `{}` as the default works well for:

```python
table.resource_tags = {"key": "value"}
```

But I'm not sure if it would properly handle:

```python
table.resource_tags["key"] = "value"
```

What are your thoughts?

Linchin: Thanks for pointing this out, indeed `get()` with default value will not handle the second case you provided. Let me undo this change.

google/cloud/bigquery/table.py
```diff
@@ -1023,6 +1024,22 @@ def table_constraints(self) -> Optional["TableConstraints"]:
             table_constraints = TableConstraints.from_api_repr(table_constraints)
         return table_constraints
 
+    @property
+    def resource_tags(self):
+        """Dict[str, str]: Resource tags for the table.
+
+        See: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table.FIELDS.resource_tags
+        """
+        return self._properties.get(
```
Linchin, line 1033:
```suggestion
        return self._properties.setdefault(
```

Linchin:
Thank you @hrkh for your contribution! It seems we just have the system test with python 3.12 failing, and it was caused by the same resource tag having been created by the python 3.8 system test. We encountered the exact same issue in #2090, and it was resolved by prefixing a randomly generated string to the tags (as in [this commit](https://github.com/googleapis/python-bigquery/pull/2090/commits/233f276c590e00f322ed530e3fcd13a5bd5447d2)). Could you see if you could implement this, so we can get this PR across the line? Cheers.

hrkh:
Thank you @Linchin for your detailed feedback and guidance!

I've updated the PR to append randomly generated strings as suffixes to the tag keys, following the approach used in the referenced commit. This should resolve the test failures with Python 3.12.

Please let me know if any further adjustments are needed.

google/cloud/bigquery/table.py
```diff
@@ -1026,6 +1027,21 @@ def table_constraints(self) -> Optional["TableConstraints"]:
...
+        return self._properties.setdefault(
+            self._PROPERTY_TO_API_FIELD["resource_tags"], {}
+        )
+
+    @resource_tags.setter
+    def resource_tags(self, value):
+        if not isinstance(value, dict) and value is not None:
+            raise ValueError("resource_tags must be a dict or None")
+        self._properties[self._PROPERTY_TO_API_FIELD["resource_tags"]] = value
+
```
Linchin, line 1044:
```suggestion

    @property
```

Linchin (approved):
LGTM. Thank you!

PR #2095, opened 2024-12-26
deps: use pandas-gbq to determine schema in `load_table_from_dataframe`

tswast:

Towards go/pandas-gbq-and-bigframes-redundancy
 🦕


tswast:
`tests/unit/test_client.py::TestClientUpload::test_load_table_from_dataframe_w_higher_scale_decimal128_datatype` looks like a real failure and might need a change in pandas-gbq.

tswast:
I've tested locally with https://github.com/googleapis/python-bigquery-pandas/pull/844 and that PR should fix the remaining system test.

```
(venv) ➜  python-bigquery git:(b323176126-pandas-gbq) pytest tests/unit/test_client.py::TestClientUpload::test_load_table_from_dataframe_w_higher_scale_decimal128_datatype
==================================================== test session starts ====================================================
platform linux -- Python 3.12.6, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/swast/src/github.com/googleapis/python-bigquery
configfile: pyproject.toml
collected 1 item                                                                                                            

tests/unit/test_client.py .                                                                                           [100%]

===================================================== 1 passed in 0.08s =====================================================
```

pyproject.toml
```diff
@@ -74,6 +74,7 @@ bqstorage = [
 ]
 pandas = [
   "pandas >= 1.1.0",
+  "pandas-gbq >= 0.26.0",
```
tswast, line 77: Update to 0.27.0 once https://github.com/googleapis/python-bigquery-pandas/pull/844 is merged and released

tswast: Updated!

tswast:
> test_load_table_from_dataframe_w_partial_schema_extra_types

Looks like a real failure (the warning message changed slightly in pandas-gbq). I'll try to update the test to be a bit more robust.

tswast:
@Linchin @chalmerlowe I believe this is finally ready for review. The failing checks are for infrastructure / Kokoro problems for a new setup, possibly?

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -444,6 +463,20 @@ def dataframe_to_bq_schema(dataframe, bq_schema):
             The automatically determined schema. Returns None if the type of
             any column cannot be determined.
     """
+    if pandas_gbq is None:
+        warnings.warn(
+            "Loading pandas DataFrame into BigQuery will require pandas-gbq "
+            "package version 0.26.0 or greater in future. "
```
tswast, line 469:
```suggestion
            "package version 0.26.1 or greater in future. "
```

tests/system/test_pandas.py
```diff
@@ -1259,7 +1259,7 @@ def test_upload_time_and_datetime_56(bigquery_client, dataset_id):
     df = pandas.DataFrame(
         dict(
             dt=[
-                datetime.datetime(2020, 1, 8, 8, 0, 0),
+                datetime.datetime(2020, 1, 8, 8, 0, 0, tzinfo=datetime.timezone.utc),
```
tswast, line 1262: Not sure is we should have defined what happens when you mix and match naive datetime (DATETIME in BQ) with datetime + timezone (TIMESTAMP in BQ). Since the test expects TIMESTAMP, I update this to be a datetime + timezone.

We still mix/match timezones, so I think that part is important to make sure we end up returning UTC values.

google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -444,6 +463,20 @@ def dataframe_to_bq_schema(dataframe, bq_schema):
             The automatically determined schema. Returns None if the type of
             any column cannot be determined.
     """
+    if pandas_gbq is None:
+        warnings.warn(
+            "Loading pandas DataFrame into BigQuery will require pandas-gbq "
+            "package version 0.26.1 or greater in future. "
```
chalmerlowe, line 469:
```suggestion
            "package version 0.26.1 or greater in the future. "
```

chalmerlowe: Question about this Deprecation.

#QUESTION/POTENTIAL BLOCKER

We state that we will require `0.26.1` in the future, but it appears that the `pyproject.toml` requires `0.26.1` **now**.

Just so I am tracking, what approach are we using for this deprecation?

* If we intend for it to be at some time in the future, what is the intermediate state that we will lean toward so that folks can migrate?
* If there is no intermediate state, could we create a situation where we break folk's code without reasonable warning?

Depending on the answer to this, it might affect other portions of this PR (docstring messaging, conditional checks, etc, TBD).

chalmerlowe (commented):
By and large, this looks good.
I have a question that might be a potential blocker.

google/cloud/bigquery/_pandas_helpers.py
tswast: > If we intend for it to be at some time in the future, what is the intermediate state that we will lean toward so that folks can migrate?

Yes, it would be at some distant future.

> If there is no intermediate state, could we create a situation where we break folk's code without reasonable warning?

I kept the existing code intact and we are testing without pandas-gbq installed in at least one session (Python 3.7). When we drop Python 3.7 support, we will need a different approach with regards to testing. Perhaps we add some custom logic to uninstall pandas-gbq before running some system/unit tests?

An alternative to consider: we could deprecate the `load_dataframe_from_table` and `to_dataframe` methods (maybe `to_arrow`, too?) in favor of the equivalent methods on pandas-gbq, in which case testing would be easier, as we could keep the existing tests intact and just add some tests checking for the warning pointing folks towards pandas-gbq.

chalmerlowe: I have an open PR that I am actively working to remove 3.7, 3.8, and add 3.13.

I would suggest the option that requires the least rework once PR #2104 is applied.

tswast: Good call. I'll take a look at how we can test with missing pandas-gbq on some test sessions.

tswast:
@chalmerlowe In https://github.com/googleapis/python-bigquery/pull/2095/commits/9a5656ba69007be17a8fedaa831252bbffda1432 I have explicitly uninstalled `pandas-gbq` in at least one unit and system test version. This should be robust to when we drop Python 3.7 and 3.8. I have verified the `pip freeze` output locally to confirm that pandas-gbq is not present in the sessions where we uninstall it.

google/cloud/bigquery/_pandas_helpers.py
tswast: Added a commit that explicitly uninstalls pandas-gbq sometimes in the noxfile.

chalmerlowe (approved):
LGTM.
Approved.

PR #2107, opened 2025-01-09
feat: adds new input validation function similar to isinstance.

chalmerlowe:
This PR adds a new helper function `isinstance_or_raise()` that raises a `TypeError` if a user inputs a value that is the wrong datatype.


Linchin (approved):
LGTM. Thank you!

PR #2108, opened 2025-01-09
feat: adds the SerDeInfo class and tests

chalmerlowe:
This PR adds the SerDeInfo (serializer/deserializer info) class and the associated tests, plus minor tweaks to support both of those changes.

### DIFF
--- a/google/cloud/bigquery/schema.py
+++ b/google/cloud/bigquery/schema.py
@@ -17,13 +17,11 @@
 from __future__ import annotations
 import collections
 import enum
+import typing
 from typing import Any, cast, Dict, Iterable, Optional, Union
 
 from google.cloud.bigquery import _helpers
 from google.cloud.bigquery import standard_sql
-from google.cloud.bigquery._helpers import (
-    _isinstance_or_raise,
-)
 from google.cloud.bigquery.enums import StandardSqlTypeNames
 
 
@@ -564,8 +562,9 @@ class PolicyTagList(object):
 
 class SerDeInfo:
     """Serializer and deserializer information.
+
     Args:
-        serializationLibrary (str): Required. Specifies a fully-qualified class
+        serialization_library (str): Required. Specifies a fully-qualified class
             name of the serialization library that is responsible for the
             translation of data between table representation and the underlying
             low-level input and output format structures. The maximum length is
@@ -588,32 +587,32 @@ class SerDeInfo:
         self.parameters = parameters
 
     @property
-    def serialization_library(self) -> Any:
+    def serialization_library(self) -> str:
         """Required. Specifies a fully-qualified class name of the serialization
         library that is responsible for the translation of data between table
         representation and the underlying low-level input and output format
         structures. The maximum length is 256 characters."""
 
-        return self._properties.get("serializationLibrary")
+        return typing.cast(str, self._properties.get("serializationLibrary"))
 
     @serialization_library.setter
     def serialization_library(self, value: str):
-        value = _isinstance_or_raise(value, str, none_allowed=False)
+        value = _helpers._isinstance_or_raise(value, str, none_allowed=False)
         self._properties["serializationLibrary"] = value
 
     @property
-    def name(self) -> Any:
+    def name(self) -> Optional[str]:
         """Optional. Name of the SerDe. The maximum length is 256 characters."""
 
         return self._properties.get("name")
 
     @name.setter
     def name(self, value: Optional[str] = None):
-        value = _isinstance_or_raise(value, str, none_allowed=True)
+        value = _helpers._isinstance_or_raise(value, str, none_allowed=True)
         self._properties["name"] = value
 
     @property
-    def parameters(self) -> Any:
+    def parameters(self) -> Optional[dict[str, str]]:
         """Optional. Key-value pairs that define the initialization parameters
         for the serialization library. Maximum size 10 Kib."""
 
@@ -621,7 +620,7 @@ class SerDeInfo:
 
     @parameters.setter
     def parameters(self, value: Optional[dict[str, str]] = None):
-        value = _isinstance_or_raise(value, dict, none_allowed=True)
+        value = _helpers._isinstance_or_raise(value, dict, none_allowed=True)
         self._properties["parameters"] = value
 
     def to_api_repr(self) -> dict:
--- a/google/cloud/bigquery/schema.py
+++ b/google/cloud/bigquery/schema.py
@@ -14,12 +14,16 @@
 
 """Schemas for BigQuery tables / queries."""
 
+from __future__ import annotations
 import collections
 import enum
 from typing import Any, cast, Dict, Iterable, Optional, Union
 
 from google.cloud.bigquery import _helpers
 from google.cloud.bigquery import standard_sql
+from google.cloud.bigquery._helpers import (
+    _isinstance_or_raise,
+)
 from google.cloud.bigquery.enums import StandardSqlTypeNames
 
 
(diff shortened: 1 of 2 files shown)


google/cloud/bigquery/schema.py
```diff
@@ -14,12 +14,16 @@
...
 
+from __future__ import annotations
 import collections
 import enum
 from typing import Any, cast, Dict, Iterable, Optional, Union
 
 from google.cloud.bigquery import _helpers
 from google.cloud.bigquery import standard_sql
+from google.cloud.bigquery._helpers import (
+    _isinstance_or_raise,
```
tswast, line 25: Nit: Use `_helpers._isinstance_or_raise` in the code instead.

> Use import statements for packages and modules only, not for individual types, classes, or functions.

https://google.github.io/styleguide/pyguide.html#22-imports

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
         """
         answer = {"names": list(self.names)}
         return answer
+
+
+class SerDeInfo:
+    """Serializer and deserializer information.
+    Args:
```
tswast, line 567: Might need a blank line here for the docs renderer to work?

```suggestion
    """Serializer and deserializer information.

    Args:
```

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+        serializationLibrary (str): Required. Specifies a fully-qualified class
```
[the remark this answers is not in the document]
tswast: +1 to transforming to PEP-8 compliant `snake_case`

tests/unit/test_schema.py
```diff
@@ -20,7 +20,10 @@
 
 from google.cloud import bigquery
 from google.cloud.bigquery.standard_sql import StandardSqlStructType
-from google.cloud.bigquery.schema import PolicyTagList
+from google.cloud.bigquery.schema import (
+    PolicyTagList,
+    SerDeInfo,
+)
```
tswast, line 26: Same nit here, regarding https://google.github.io/styleguide/pyguide.html#22-imports

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
         """
         answer = {"names": list(self.names)}
         return answer
+
+
+class SerDeInfo:
```
tswast, line 565: Is this the name in the REST API? I'd prefer the full names rather than abbreviations, but I'm OK with this if it's to be consistent with the proto/REST API.

tswast (approved):
Overall approach looks great. A couple nits.

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+        parameters: Optional[dict[str, str]] = None,
+    ):
+        self._properties: Dict[str, Any] = {}
+        self.serialization_library = serialization_library
+        self.name = name
+        self.parameters = parameters
+
+    @property
+    def serialization_library(self) -> Any:
+        """Required. Specifies a fully-qualified class name of the serialization
```
Linchin, line 592: nit: I think `Required.` should only apply to the init.

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+        return self._properties.get("serializationLibrary")
+
+    @serialization_library.setter
+    def serialization_library(self, value: str):
+        value = _isinstance_or_raise(value, str, none_allowed=False)
+        self._properties["serializationLibrary"] = value
+
+    @property
+    def name(self) -> Any:
+        """Optional. Name of the SerDe. The maximum length is 256 characters."""
```
Linchin, line 606: nit: similarly, we can remove `Optional.` here.

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+
+        return self._properties.get("serializationLibrary")
+
+    @serialization_library.setter
+    def serialization_library(self, value: str):
+        value = _isinstance_or_raise(value, str, none_allowed=False)
+        self._properties["serializationLibrary"] = value
+
+    @property
+    def name(self) -> Any:
```
Linchin, line 605:
```suggestion
    def name(self) -> Optional[str]:
```

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+        name: Optional[str] = None,
+        parameters: Optional[dict[str, str]] = None,
+    ):
+        self._properties: Dict[str, Any] = {}
+        self.serialization_library = serialization_library
+        self.name = name
+        self.parameters = parameters
+
+    @property
+    def serialization_library(self) -> Any:
```
Linchin, line 591:
```suggestion
    def serialization_library(self) -> str:
```

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+
+        return self._properties.get("name")
+
+    @name.setter
+    def name(self, value: Optional[str] = None):
+        value = _isinstance_or_raise(value, str, none_allowed=True)
+        self._properties["name"] = value
+
+    @property
+    def parameters(self) -> Any:
```
Linchin, line 616:
```suggestion
    def parameters(self) -> Optional[dict[str, str]]:
```
For consistent type annotations.

google/cloud/bigquery/schema.py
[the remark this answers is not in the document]
chalmerlowe: Resolved.

google/cloud/bigquery/schema.py
chalmerlowe: This is great feedback. I had not realized that.

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+        self._properties["serializationLibrary"] = value
+
+    @property
+    def name(self) -> Any:
+        """Optional. Name of the SerDe. The maximum length is 256 characters."""
+
+        return self._properties.get("name")
+
+    @name.setter
+    def name(self, value: Optional[str] = None):
```
Linchin, line 611: It feels a bit weird to let the parameter of a property setter to be optional, because we will use it as `some_serializer.name = "a_name"`, it shouldn't happen that `value` is no provided.

google/cloud/bigquery/schema.py
chalmerlowe: Resolved.

google/cloud/bigquery/schema.py
chalmerlowe: Resolved.

google/cloud/bigquery/schema.py
```diff
@@ -556,3 +560,88 @@ def to_api_repr(self) -> dict:
...
+
+    @property
+    def parameters(self) -> Any:
+        """Optional. Key-value pairs that define the initialization parameters
+        for the serialization library. Maximum size 10 Kib."""
+
+        return self._properties.get("parameters")
+
+    @parameters.setter
+    def parameters(self, value: Optional[dict[str, str]] = None):
```
Linchin, line 623: Same as https://github.com/googleapis/python-bigquery/pull/2108/files#r1909362061

tests/unit/test_schema.py
chalmerlowe: Hmm. I attempted to correct this for both `PolicyTagList` and `SerDeInfo`.
Unfortunately, there is something about how PolicyTagList is handled in our code that did not like the change and it produced several failing tests. I did not want to clutter this PR with unrelated changes, nor did I want to delay the issuance of this PR.

So I chose to do the following:
* I corrected `SerDeInfo` and throughout the code we reference `schema.SerDeInfo`.
* I left the PolicyTagList code exactly as it was found.

google/cloud/bigquery/schema.py
chalmerlowe: `SerDeInfo` is the name in the proto/REST API definitions.

google/cloud/bigquery/schema.py
chalmerlowe: As mentioned in chat, I prefer to leave this as is.

google/cloud/bigquery/schema.py
chalmerlowe: As mentioned in chat, I prefer to leave this as is.

google/cloud/bigquery/schema.py
chalmerlowe: To type checkers `Optional` has a very specific meaning.
It means: `value` could be one of the types found in the brackets OR it could be `None`.

From the [Python Docs on the typing module](https://docs.python.org/3/library/typing.html#typing.Optional):

`typing.Optional`
`Optional[X]` is equivalent to `X | None` (or `Union[X, None]`).

Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the `Optional` qualifier on its type annotation just because it is optional. For example:

```
def foo(arg: int = 0) -> None:
    ...
```

On the other hand, if an explicit value of `None` is allowed, the use of `Optional` is appropriate, whether the argument is optional or not. For example:

```
def foo(arg: Optional[int] = None) -> None:
    ...
```

google/cloud/bigquery/schema.py
chalmerlowe: Resolved.

google/cloud/bigquery/schema.py
chalmerlowe: See reply above from Python Docs about the concept of **optional** versus `typing.Optional`.

google/cloud/bigquery/schema.py
chalmerlowe: Resolved.

google/cloud/bigquery/schema.py
tswast: B.T.W. Whenever we drop Python 3.9 and have Python 3.10 as a minimum, the recommendation is to use `X | None` to prevent exactly this sort of confusion.

https://google.github.io/styleguide/pyguide.html#3195-nonetype

Linchin: Thank you for the explanation!

Linchin (approved):

PR #2111, opened 2025-01-15
feat: adds ExternalCatalogDatasetOptions and tests

chalmerlowe:
This PR adds the `ExternalCatalogDatasetOptions` class and the associated tests, plus minor tweaks to support both of those changes.

### DIFF
--- a/google/cloud/bigquery/dataset.py
+++ b/google/cloud/bigquery/dataset.py
@@ -911,7 +911,7 @@ class Dataset(object):
         )
 
         if prop is not None:
-            prop = external_config.ExternalCatalogDatasetOptions().from_api_repr(prop)
+            prop = external_config.ExternalCatalogDatasetOptions.from_api_repr(prop)
         return prop
 
     @external_catalog_dataset_options.setter
--- a/google/cloud/bigquery/dataset.py
+++ b/google/cloud/bigquery/dataset.py
@@ -921,7 +921,7 @@ class Dataset(object):
         )
         self._properties[
             self._PROPERTY_TO_API_FIELD["external_catalog_dataset_options"]
-        ] = value.to_api_repr()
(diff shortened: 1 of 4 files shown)


google/cloud/bigquery/dataset.py
```diff
@@ -898,6 +900,29 @@ def storage_billing_model(self, value):
...
+        """Options defining open source compatible datasets living in the
+        BigQuery catalog. Contains metadata of open source database, schema
+        or namespace represented by the current dataset."""
+
+        prop = _helpers._get_sub_prop(
+            self._properties, ["externalCatalogDatasetOptions"]
+        )
+
+        if prop is not None:
+            prop = external_config.ExternalCatalogDatasetOptions().from_api_repr(prop)
```
tswast, line 914: `from_api_repr` should be a classmethod, so no instance is needed.

```suggestion
            prop = external_config.ExternalCatalogDatasetOptions.from_api_repr(prop)
```

google/cloud/bigquery/dataset.py
```diff
@@ -898,6 +900,29 @@ def storage_billing_model(self, value):
...
+        return prop
+
+    @external_catalog_dataset_options.setter
+    def external_catalog_dataset_options(self, value):
+        value = _helpers._isinstance_or_raise(
+            value, external_config.ExternalCatalogDatasetOptions, none_allowed=True
+        )
+        self._properties[
+            self._PROPERTY_TO_API_FIELD["external_catalog_dataset_options"]
+        ] = value.to_api_repr()
```
tswast, line 924: If the user passes in `None`, `to_api_repr()` will fail with `AttributeError`. Try something like this:

```suggestion
        ] = value.to_api_repr() if value is not None else None
```

tswast (approved):

PR #2097, opened 2024-12-27
feat: preserve unknown fields from the REST API representation in `SchemaField`

tswast:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Towards simplifying https://github.com/googleapis/python-bigquery/pull/2081
Fixes #6 even more than it already was by avoiding even more copy operations.
🦕

PR #2115, opened 2025-01-16
fix: add default value in SchemaField.from_api_repr()

Linchin:
Fixing https://github.com/googleapis/python-bigquery-pandas/issues/854, do not merge until yanking 3.28.0 is complete.

In https://github.com/googleapis/python-bigquery/pull/2097/files#diff-dc8e320384c6a5d2b65e1e05c4853b853575618ff4706a6f2c74b9363d49ea14 in method `SchemaField.from_api_repr()`, we no longer set a default value for `mode` if it's not provided in the API representation, causing tests to fail in pandas-gbq. This PR adds it back.

### DIFF
--- a/google/cloud/bigquery/schema.py
+++ b/google/cloud/bigquery/schema.py
@@ -228,6 +228,12 @@ class SchemaField(object):
         # fields. See https://github.com/googleapis/python-bigquery/issues/6
         placeholder._properties = api_repr
 
+        # Add the field `mode` with default value if it does not exist. Fixes
+        # an incompatibility issue with pandas-gbq:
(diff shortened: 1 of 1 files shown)


Linchin:
After our discussion, I don't think we should add `description`, because before the change that was to be reverted, field `description` wouldn't be created if we just provide _DEFAULT_VALUE. See https://github.com/googleapis/python-bigquery/blob/9c504186f03ffb8b86836c683912b310db2435e9/google/cloud/bigquery/schema.py#L191-L192

tswast (approved):

PR #2126, opened 2025-02-04
Feat: Adds foreign_type_info attribute to table class and adds unit tests.

chalmerlowe:
This PR adds support for the `foreign_type_info` attribute on the Table class and the associated tests.


google/cloud/bigquery/table.py
```diff
@@ -451,6 +453,15 @@ def schema(self):
...
                 instance or a compatible mapping representation of the field.
+
+        NOTE: If you are referencing a schema for an external catalog table such
+        as a Hive table, it will also be necessary to populate the foreign_type_info
+        attribute. This is not necessary if defining the schema for a BigQuery table.
+
+        For details, see:
+        https://cloud.google.com/bigquery/docs/external-tables
+        https://cloud.google.com/bigquery/docs/datasets-intro#external_datasets
+
```
tswast, line 464: I think we'll need some logic in the setter for schema to avoid overwriting the `schema` property entirely. Instead, it'll need to be responsible for just `schema.fields`.

google/cloud/bigquery/table.py
```diff
@@ -1075,6 +1086,41 @@ def external_catalog_table_options(
...
+                Foreign type information, or :data:`None` if not set.
+
+        NOTE: foreign_type_info is only required if you are referencing an
+        external catalog such as a Hive table.
+        For details, see:
+        https://cloud.google.com/bigquery/docs/external-tables
+        https://cloud.google.com/bigquery/docs/datasets-intro#external_datasets
+        """
+
+        prop = self._properties.get(self._PROPERTY_TO_API_FIELD["foreign_type_info"])
```
tswast, line 1105: Even if we are exposing this at the table level, it needs to be fetched from schema, still, right?

This is exactly the sort of thing the `_get_sub_prop` (https://github.com/googleapis/python-bigquery/blob/54c8d07f06a8ae460c9e0fb1614e1fbc21efb5df/google/cloud/bigquery/_helpers.py#L781) and `_set_sub_prop` (https://github.com/googleapis/python-bigquery/blob/54c8d07f06a8ae460c9e0fb1614e1fbc21efb5df/google/cloud/bigquery/_helpers.py#L830) helpers are intended to be used for.

We even use it in other `Table` properties, such as `project`: https://github.com/googleapis/python-bigquery/blob/54c8d07f06a8ae460c9e0fb1614e1fbc21efb5df/google/cloud/bigquery/table.py#L175-L177


```suggestion
        prop = _helpers._get_sub_prop(self._properties, self._PROPERTY_TO_API_FIELD["foreign_type_info"])
```

google/cloud/bigquery/table.py
```diff
@@ -411,6 +412,7 @@ class Table(_TableBase):
         "max_staleness": "maxStaleness",
         "resource_tags": "resourceTags",
         "external_catalog_table_options": "externalCatalogTableOptions",
+        "foreign_type_info": "foreignTypeInfo",
```
tswast, line 415: We should adjust this to the format used for `_helpers._get_sub_prop` and `_helpers._set_sub_prop`. Likewise, let's replace `schema` with something compatible with that.

```suggestion
        "foreign_type_info": ["schema", "foreignTypeInfo"],
        # TODO: remove "schema" from above (between time_partitioning and "snapshot_definition"
        "schema": ["schema", "fields"],
```

tests/unit/test_table.py
```diff
@@ -5993,6 +5994,71 @@ def test_external_catalog_table_options_from_api_repr(self):
         assert result == expected
 
 
+class TestForeignTypeInfo:
```
tswast, line 5997: I'd also like to see a test where we do `Table.from_api_repr` and `Table.to_api_repr` so that we can visually compare that the correct `schema.foreignTypeInfo` field of the REST API object is set.

tswast (changes requested):

google/cloud/bigquery/table.py
Linchin: Also, I'm not sure if this format will render well in the docs. We might just move all the contents under `NOTE:` to after `Table's schema.`

tests/unit/test_table.py
```diff
@@ -5993,6 +5994,71 @@ def test_external_catalog_table_options_from_api_repr(self):
...
+        return Table
+
+    def _make_one(self, *args, **kw):
+        return self._get_target_class(self)(*args, **kw)
+
+    def test_foreign_type_info_default_initialization(self):
+        table = self._make_one(self.TABLEREF)
+        assert table.foreign_type_info is None
+
+    def test_foreign_type_info_valid_inputs(self):
```
Linchin, line 6022: Could we also add test cases for the setter for other supported types, i.e., dict and `None`.

google/cloud/bigquery/table.py
chalmerlowe: Done.

google/cloud/bigquery/table.py
chalmerlowe: Done.

google/cloud/bigquery/table.py
chalmerlowe: I have not yet addressed Tim's comment here.
I spoke with Linchin about the note and with the revision I added, Sphinx should be able to handle the note with no problem.

tests/unit/test_table.py
chalmerlowe: Not yet addressed.

tests/unit/test_table.py
chalmerlowe: Not yet addressed.

google/cloud/bigquery/table.py
chalmerlowe: Partially complete. Added a new value for `schema`, but not yet done with `foreign_type_info`.

chalmerlowe: Complete.

google/cloud/bigquery/table.py
chalmerlowe: Complete.

tests/unit/test_table.py
chalmerlowe: Done.

tests/unit/test_table.py
chalmerlowe: Complete.
This test is now parametrized so that it tests under three input conditions:

* dict 
* None
* using a ForeignTypeInfo object

google/cloud/bigquery/_helpers.py
```diff
@@ -978,10 +978,12 @@ def _build_resource_from_properties(obj, filter_fields):
...
     partial = {}
     for filter_field in filter_fields:
-        api_field = obj._PROPERTY_TO_API_FIELD.get(filter_field)
+        api_field = _get_sub_prop(obj._PROPERTY_TO_API_FIELD, filter_field)
         if api_field is None and filter_field not in obj._properties:
             raise ValueError("No property %s" % filter_field)
         elif api_field is not None:
+            if isinstance(api_field, list):
+                api_field = api_field[0]
             partial[api_field] = obj._properties.get(api_field)
```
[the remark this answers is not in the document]
chalmerlowe: Done.

tswast (approved):
Thanks!

Linchin (approved):
LGTM. Thank you!

PR #2127, opened 2025-02-04
test: add unit test covering the case where worker streams are stopped early

### LINKED ISSUE #2032: Deadlock due to race condition in download thread when using Bigquery Storage API

tswast:
This PR is a follow-up to https://github.com/googleapis/python-bigquery/pull/2034 with the purpose of testing the error condition.

Note: I have confirmed that this test fails without the changes from https://github.com/googleapis/python-bigquery/pull/2034. Actually, it hangs forever because the child threads aren't shutdown but the assertions do fail when I step through with a debugger.

Fixes #2032 🦕

### DIFF
--- a/tests/unit/test__pandas_helpers.py
+++ b/tests/unit/test__pandas_helpers.py
@@ -1874,10 +1874,10 @@ def test__download_table_bqstorage_shuts_down_workers(
         ]
     )
     arrow_rows = pyarrow.record_batch(
-        {
-            "int_col": [0, 1, 2],
-            "str_col": ["a", "b", "c"],
-        },
+        [
+            pyarrow.array([0, 1, 2], type=pyarrow.int64()),
+            pyarrow.array(["a", "b", "c"], type=pyarrow.string()),
+        ],
         schema=arrow_schema,
     )
     session = google.cloud.bigquery_storage_v1.types.ReadSession()
--- a/tests/unit/test__pandas_helpers.py
+++ b/tests/unit/test__pandas_helpers.py
@@ -16,6 +16,7 @@ import collections
 import datetime
 import decimal
 import functools
+import gc
 import operator
 import queue
 from typing import Union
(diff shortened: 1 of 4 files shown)


tswast:
Looking at the test failures, there's a chance that

```
FAILED samples/tests/test_download_public_data.py::test_download_public_data
FAILED samples/tests/test_download_public_data_sandbox.py::test_download_public_data_sandbox

File: tests/unit/test__pandas_helpers.py [hunk 1]
```diff
@@ -1846,6 +1846,94 @@ def fake_download_stream(
...
+    )
+
+    result_gen_iter = iter(result_gen)
+    next(result_gen_iter)
+    assert download_state.started_workers == 3
+    assert download_state.finished_workers == 0
+
+    del result_gen, result_gen_iter
+    assert download_state.started_workers == 3
+    assert download_state.finished_workers == 3
```
tswast, line 1934: Thought: Is this going to be flaky? What if it takes a while for the worker threads to resume? Is `del` synchronous?

From: https://www.reddit.com/r/Python/comments/18vgrc9/comment/kfrlg5z/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button it sounds like a `gc.collect()` call might make this a bit safer with regards to confidence that the background threads should really have stopped. I'll add that in.

Linchin (approved):
Thanks a lot for reviewing the PR and adding the tests! Since we are deleting the part verifying the log, I wonder if it would be useful to add something to check that a storage read stream was indeed created? (maybe it's not necessary because the sample already set `create_bqstorage_client=True`)

Issue #2132, opened 2025-02-14
Cleanup tech debt from removing Python 3.7 and 3.8


chalmerlowe:
While removing Python 3.7 and 3.8 as dependencies, a number of minor blockers or bits of tech debt were discovered. Resolve or disposition the following items:

- [ ] Add a test regarding range_timestamp_dtype. See TODO @ ~#L340. [python-bigquery/google/cloud/bigquery/_pandas_helpers.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/_pandas_helpers.py)
- [ ] Add a test regarding dataframe_column_dtypes. See TODO @ ~#L648. [python-bigquery/google/cloud/bigquery/test_query_pandas.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/test_query_pandas.py)
- [ ] Add a test regarding dataframe_nullable_scalars. See TODO @ ~#L37. [python-bigquery/google/cloud/bigquery/test_table_pandas.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/test_table_pandas.py)
- [x] Identify the correct type hinting for the code associated with the TODO statements:
  - [x] [python-bigquery/google/cloud/bigquery/external_config.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/external_config.py) @ ~#L838.
  - [x] [python-bigquery/google/cloud/bigquery/job/base.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/job/base.py) @ ~#L418
  - [x] [python-bigquery/google/cloud/bigquery/routine.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/routine.py) @ ~#L521, ~#L528, and ~#L535
  - [x] [python-bigquery/google/cloud/bigquery/schema.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/schema.py) @ ~#L236
  - [x] [python-bigquery/google/cloud/bigquery/table.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/table.py) @ ~#L138
  - [x] [python-bigquery/google/cloud/bigquery/schema.py](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/schema.py) @ ~#L236

PR #2133, opened 2025-02-14
deps: Remove Python 3.7 and 3.8 as supported runtimes

chalmerlowe:
Fix: Update Python versions by dropping support for versions 3.7, 3.8.

Also includes minor fixes and tweaks needed to support the primary changes related to dropping runtimes.

~This is a work in progress. Do not merge.~


tswast:
Nit: I think `deps: ` is the most appropriate category for the changelog. See: https://github.com/googleapis/release-please/blob/90058f543cd82858539fdc7d1b34eabd348edbee/src/strategies/python.ts#L32-L43 for all possible options.

google/cloud/bigquery/external_config.py
```diff
@@ -835,8 +835,10 @@ def schema(self):
         See
         https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#ExternalDataConfiguration.FIELDS.schema
         """
-        prop = self._properties.get("schema", {})
-        return [SchemaField.from_api_repr(field) for field in prop.get("fields", [])]
+        # TODO: The typehinting for this needs work. Setting this pragma to temporarily
```
Linchin, line 838: Just a gentle reminder to add this to #2132

google/cloud/bigquery/job/base.py
```diff
@@ -415,7 +415,9 @@ def __init__(self, job_id, client):
     @property
     def configuration(self) -> _JobConfig:
         """Job-type specific configurtion."""
-        configuration = self._CONFIG_CLASS()
+        # TODO: The typehinting for this needs work. Setting this pragma to temporarily
```
Linchin, line 418: similar tohttps://github.com/googleapis/python-bigquery/pull/2133/files#r1962595719

tests/unit/test_table.py
```diff
@@ -3621,11 +3617,10 @@ def test_to_dataframe_no_tqdm_no_progress_bar(self):
...
         ]
-        # With Python 3.7 and 3.8, len(user_warnings) = 3. With pandas < 1.5,
-        # pandas.ArrowDtype is not supported. We raise warnings because
-        # range columns have to be converted to object.
-        # With higher Python versions and noextra tests, len(user_warnings) = 0
-        self.assertIn(len(user_warnings), [0, 3])
+        # With pandas < 1.5 pandas.ArrowDtype is not supported. We raise warnings
+        # because range columns have to be converted to object.
+        # With Python versions > 1.5 and noextra tests, len(user_warnings) = 0
+        self.assertEqual(len(user_warnings), 0)
```
Linchin, line 3623: What is the expected number of warnings with pandas < 1.5? Not sure if we still need the comments.

Linchin:
Thanks so much for removing 3.7 and 3.8, it turned out to be much for convoluted than we thought! It looks good, just need to pass lint. Also, it would be helpful to add the TODO trackings in #2132.

google/cloud/bigquery/__init__.py
```diff
@@ -126,12 +126,12 @@
 
 if sys_major == 3 and sys_minor in (7, 8):
     warnings.warn(
-        "The python-bigquery library will stop supporting Python 3.7 "
-        "and Python 3.8 in a future major release expected in Q4 2024. "
+        "The python-bigquery library no longer supports Python 3.7 "
```
Linchin, line 129: Should we raise an exception here instead of warning? WDYT?

chalmerlowe: `FutureWarning` is a step up from a `PendingDeprecationWarning` (and unlike a `PendingDeprecationWarning`, it can be seen by end users.)

Since for customers running on 3.7 or 3.8, next time they go to install, an Exception would be a breaking change for them ... I think it makes sense to stick with the `FutureWarning` and change to an `Exception` during the next major release.

google/cloud/bigquery/external_config.py
chalmerlowe: All TODO items in this PR have been added to Issue #2132

google/cloud/bigquery/job/base.py
chalmerlowe: All TODO items in this PR have been added to Issue #2132

chalmerlowe:
> Thanks so much for removing 3.7 and 3.8, it turned out to be much for convoluted than we thought! It looks good, just need to pass lint. I also wanted to double check on the coverage as we have several sections skipped. Are they safe to skip for now or we had better add coverage? For the ones ok to be skipped, let's make sure we track them in #2132.

I provided some context in chat for your consideration.

google/cloud/bigquery/__init__.py
chalmerlowe: I can see about changing it to an Exception.

tests/unit/test__pandas_helpers.py
```diff
@@ -521,43 +518,6 @@ def test_bq_to_arrow_array_w_nullable_scalars(module_under_test, bq_type, rows):
     assert rows == roundtrip
 
 
-@pytest.mark.parametrize(
```
Linchin, line 524: Is it safe to remove this test completely? It still seems possible to use pandas 1.x

.kokoro/test-samples-impl.sh
```diff
@@ -33,7 +33,6 @@ export PYTHONUNBUFFERED=1
 env | grep KOKORO
 
 # Install nox
-# `virtualenv==20.26.6` is added for Python 3.7 compatibility
```
Linchin, line 36: should we remove `virtualenv==20.26.6` because 3.7 is removed too?

chalmerlowe: We want virtualenv.
I am testing locally to see if we can simply install virtualenv without specifying a version number.

chalmerlowe: Somewhere Owlbot.py did some post processing and reverted that bit of code to add that change back in because it pulled in the template for certain files and overwrote my changes.
I edited owlbot.py so that it will not touch certain templates AND so that for certain templates that it will protect the edits that we need to have go through.

Linchin (approved):
Thank you so much for working on deprecating 3.7 and 3.8, it's been great work! LGTM. 🎉

PR #2136, opened 2025-02-21
deps: updates required checks list in github

chalmerlowe:
Removes Required status for github checks against Python versions 3.7 and 3.8. These checks have been removed upstream in google3 and will no longer run. Thus these github checks will never finish, blocking the ability to merge PRs.

PR #2133 will further update this file to require checks of 3.9, as part of a larger effort.


Linchin (approved):
LGTM, thank you!

PR #1876, opened 2024-03-27
fix: avoid "Unable to determine type" warning with JSON columns in `to_dataframe`

tswast:
Based on https://github.com/googleapis/python-bigquery/pull/2144, which should merge first.

TODO:

- [x] tests
- [x] maybe we don't want string columns for JSON? Update: keeping string because that's consistent with (1) current behavior and (2) the BQ Storage Read API.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes https://github.com/googleapis/python-bigquery/issues/1580
🦕

PR #2144, opened 2025-03-11
chore: refactor cell data parsing to use classes for easier overrides

tswast:

Towards https://github.com/googleapis/python-bigquery/pull/1876
 🦕


GarrettWu (approved):

google/cloud/bigquery/_helpers.py
```diff
@@ -133,262 +133,279 @@ def _not_null(value, field):
...
+    ) -> Optional[relativedelta.relativedelta]:
+        """Coerce 'value' to an interval, if set or not nullable."""
+        if not _not_null(value, field):
+            return None
+        if value is None:
+            raise TypeError(f"got {value} for REQUIRED field: {repr(field)}")
+
+        parsed = _INTERVAL_PATTERN.match(value)
+        if parsed is None:
+            raise ValueError(f"got interval: '{value}' with unexpected format")
```
chalmerlowe, line 190: If there is a URL we can send folks to so that they can get more information on what an acceptable format looks like, please include it in the error message to make this more actionable.

tswast: That's a good point. https://github.com/googleapis/python-bigquery/pull/2144/commits/3ea3abbc43aa7943564a4a8dd4673311c230cc27 adds URLs for each of the ValueErrors I could find in this PR.

tests/unit/_helpers/test_cell_data_parser.py
```diff
@@ -0,0 +1,467 @@
+# Copyright 2021 Google LLC
```
chalmerlowe, line 1: Somebody remind me: what is the policy on updating the Copyright boilerplate, especially as this is effectively a new document?

chalmerlowe (approved):
I have one question about boilerplate copyright text.

That is not a show stopper. 

It looks like coverage is grumbly about one of the lines in `_helpers.py`. Once that gets cleaned up, we should be good to go.

Gonna approve this.

tests/unit/_helpers/test_cell_data_parser.py
tswast: Official policy:

> Note: Use the year the file was created as the date in your copyright header. For more detailed instructions, see go/copyright.

Public docs here: https://opensource.google/documentation/reference/releasing/preparing#license-headers

But in go/copyright it also says:

> There is no need to update the year if the file is just being moved from one place to another.

In this case, it's a bit ambiguous as much of these tests are similar to the ones from 2021, but also I've modified just about every line of code in this file, so it probably does count as "new".

```suggestion
# Copyright 2025 Google LLC
```

PR #2153, opened 2025-04-02
test: adds pytest-xdist to speed up processing of CI/CD checks

chalmerlowe:
DO NOT MERGE.
WIP experiments with pytest-xdist.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -98,6 +98,7 @@ def default(session, install_extras=True):
         "pytest",
         "google-cloud-testutils",
         "pytest-cov",
+        "pytest-xdist",
         "freezegun",
         "-c",
         constraints_path,
@@ -129,6 +130,7 @@ def default(session, install_extras=True):
     # Run py.test against the unit tests.
     session.run(
         "py.test",
+        "-n=auto",
         "--quiet",
         "-W default::PendingDeprecationWarning",
         "--cov=google/cloud/bigquery",
--- a/noxfile.py
+++ b/noxfile.py
@@ -226,7 +226,12 @@ def system(session):
 
     # Install all test dependencies, then install local packages in place.
     session.install(
-        "pytest", "psutil", "google-cloud-testutils", "-c", constraints_path
+        "pytest",
+        "psutil",
+        "pytest-xdist",
+        "google-cloud-testutils",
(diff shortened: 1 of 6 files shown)


samples/notebooks/requirements-test.txt
```diff
@@ -1,3 +1,4 @@
 google-cloud-testutils==1.6.0
 pytest==8.3.5
 mock==5.2.0
+pytest-==3.6.1
```
chalmerlowe, line 4:
```suggestion
pytest==3.6.1
```

chalmerlowe:
@Linchin 

Using `pytest-xdist` **for tests that run under pytest** shortens the time for a number of our CI/CD checks. In this image, darker green means more time savings. For some checks we cut total time by five minutes.

Our two worst offenders do not benefit, because: 
* PRESUBMIT: has `nox` sessions internally and though each is less than 2 minutes long, each goes through a virtualenv setup, package install, etc and does so sequentially thus by the time we finish ~10 sessions, we have aggregated 13 minutes). This should be broken into individual CI/CD checks (out of scope). 
* LINTING-TYPING: does not use `pytest`, instead these tests run using `mypy`, `pytype`, etc. `pytype` is the long pole in tent at 10 minutes. Despite this, these should be broken into individual CI/CD checks to allow more of them to run in parallel (out of scope). It may be possible to provide a wrapper to run `pytype` against multiple files in parallel using `multiprocessing` or `subprocess` modules (out of scope). 

Screenshot 2025-04-11 at 6 37 57 AM

noxfile.py
```diff
@@ -129,6 +130,7 @@ def default(session, install_extras=True):
     # Run py.test against the unit tests.
     session.run(
         "py.test",
+        "-n=auto",
```
tswast, line 133: I think the default is the number of CPUs on the instance, right? I recall in bigframes we set this a bit higher than that and it seemed to help.

tswast (approved):

noxfile.py
chalmerlowe: I believe you are correct, that the default is the number of CPUs on the instance.

Linchin (approved):
Thanks for making the tests a lot faster!

PR #2154, opened 2025-04-02
fix: table iterator should not use bqstorage when page_size is not None

chelsea-lin:

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1873,6 +1873,11 @@ class RowIterator(HTTPIterator):
         """total bytes processed from job statistics, if present."""
         return self._total_bytes_processed
 
+    @property
+    def page_size(self) -> Optional[int]:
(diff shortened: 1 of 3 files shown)


tswast (approved):
Thanks!

tswast:
Looks like two tests are failing:

```
FAILED tests/unit/test_dbapi_cursor.py::TestCursor::test_fetchall_w_bqstorage_client_fetch_error_no_fallback
FAILED tests/unit/test_dbapi_cursor.py::TestCursor::test_fetchall_w_bqstorage_client_no_arrow_compression
```

I think we need to remove this logic:

https://github.com/googleapis/python-bigquery/blob/77d71736fcc006d3ab8f8ba17955ad5f06e21876/google/cloud/bigquery/dbapi/cursor.py#L380-L384

Page size of 1 is going to be super slow, so I don't think we should be setting that if people explicitly request "None" as the page size.

Issue #2157, opened 2025-04-04
expose biglakeConfiguration properties on Table


tswast:
**Is your feature request related to a problem? Please describe.**

I'd like to lookup some info about my [biglake tables](https://cloud.google.com/bigquery/docs/biglake-intro) with the Python client. Specifically, I'd like to check if the `tableFormat` is `ICEBERG` and if it is, then try to read the parquet files from the `storageUri`.

 **Describe the solution you'd like**

 A `Table.biglake_configuration` property that returns some `BigLakeConfiguration` object would make the most sense to me and align with the [REST API for BigLakeConfiguration](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#biglakeconfiguration).

 **Describe alternatives you've considered**

My workaround:

```python
table_resource = table.to_api_repr()
if (data_config := table_resource.get("biglakeConfiguration", None)) is not None:
    if (
        (data_config.get("tableFormat", None) == "ICEBERG")
        and (data_config.get("fileFormat", None) == "PARQUET")
        and ((uri := data_config.get("storageUri")) is not None)
    ):
        print(uri)
```

 **Additional context**

Release note: https://cloud.google.com/bigquery/docs/release-notes#July_25_2022

> July 25, 2022
>
> [BigLake](https://cloud.google.com/bigquery/docs/biglake-intro) is now [generally available](https://cloud.google.com/products/#product-launch-stages) (GA). You can now create [BigQuery ML](https://cloud.google.com/bigquery-ml/docs/introduction) models using data in Cloud Storage by using BigLake and publish BigLake tables as [Analytics Hub listings](https://cloud.google.com/bigquery/docs/analytics-hub-introduction).

shollyman:
Still pending a release I believe, but this should now be exposed.

PR #2162, opened 2025-04-10
feat: support BigLakeConfiguration (managed Iceberg tables)

shollyman:
This PR adds the BigLakeConfiguration class to tables, and the necessary property mappings from Table.  It also adds some utility enums (BigLakeFileFormat, BigLakeTableFormat) to more easily communicate available values for configuraiton.

This PR adds unit testing, but eschews integration testing at this time given the multiple cross-service dependencies involved.

Fixes:https://github.com/googleapis/python-bigquery/issues/2157

### DIFF
--- a/tests/unit/test_table.py
+++ b/tests/unit/test_table.py
@@ -2308,7 +2308,7 @@ class TestBigLakeConfiguration(unittest.TestCase):
         }
 
         first = self._make_one(_properties=resource)
-        second = self._make_one(_properties=resource)
+        second = self._make_one(_properties=copy.deepcopy(resource))
         self.assertEqual(first, second)
         self.assertNotEqual(first, resource)
 
--- a/tests/unit/test_table.py
+++ b/tests/unit/test_table.py
@@ -2288,8 +2288,6 @@ class TestBigLakeConfiguration(unittest.TestCase):
         self.assertEqual(instance.to_api_repr(), resource)
 
     def test_from_api_repr_partial(self):
-        from google.cloud.bigquery.enums import BigLakeTableFormat
-
         klass = self._get_target_class()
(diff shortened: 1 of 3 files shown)


tswast (approved):
Thanks!

shollyman:
snippet failure unrelated to PR, issue with auth in multiple repos today
```
ERROR: (gcloud.auth.activate-service-account) There was a problem refreshing your current auth tokens: invalid_grant: Invalid JWT Signature.
```

tests/unit/test_table.py
```diff
@@ -2166,6 +2235,93 @@ def test_ctor_full_resource(self):
...
+            "fileFormat": "FILE",
+            "tableFormat": "TABLE",
+        }
+
+        first = self._make_one(_properties=resource)
+        second = self._make_one(_properties=copy.deepcopy(resource))
+        self.assertNotEqual(first, resource)
+        self.assertEqual(first, second)
+        self.assertEqual(hash(first), hash(second))
+        self.assertNotEqual(first, resource)
```
chalmerlowe, line 2315: Overall, LGTM. Approved.

QUESTION/NIT:
Is it necessary to have this test twice? We seem to have the same test twice within three lines of each other. However, I don't see any intermediate step that is likely to change either of these two inputs (`first` or `resource`).  Below, when we test `first` and `second` we purposely update the value of `first.foo` to match `second.foo` so repeating the test makes sense in that case.
```
self.assertNotEqual(first, resource)
```

chalmerlowe (approved):

tests/unit/test_table.py
shollyman: cleaned up the dupe, and added comments to make intent clearer.

PR #2167, opened 2025-04-21
fix: `query()` now warns when `job_id` is set and the default `job_retry` is ignored

tswast:

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -3388,7 +3388,7 @@ class Client(ClientWithProject):
         project: Optional[str] = None,
         retry: retries.Retry = DEFAULT_RETRY,
         timeout: TimeoutType = DEFAULT_TIMEOUT,
-        job_retry: retries.Retry = DEFAULT_JOB_RETRY,
+        job_retry: Optional[retries.Retry] = DEFAULT_JOB_RETRY,
         api_method: Union[str, enums.QueryApiMethod] = enums.QueryApiMethod.INSERT,
     ) -> job.QueryJob:
         """Run a SQL query.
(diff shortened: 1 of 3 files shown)


google/cloud/bigquery/client.py
```diff
@@ -3524,7 +3515,7 @@ def query_and_wait(
         api_timeout: TimeoutType = DEFAULT_TIMEOUT,
         wait_timeout: Union[Optional[float], object] = POLLING_DEFAULT_VALUE,
         retry: retries.Retry = DEFAULT_RETRY,
-        job_retry: retries.Retry = DEFAULT_JOB_RETRY,
+        job_retry: Optional[retries.Retry] = DEFAULT_JOB_RETRY,
```
tswast, line 3518: The warning only applies to `query()` since we don't take a `job_id` in `query_and_wait`. We can revert this change.

```suggestion
        job_retry: retries.Retry = DEFAULT_JOB_RETRY,
```

tests/unit/test_job_retry.py
```diff
@@ -511,26 +511,34 @@ def api_request(method, path, query_params=None, data=None, **kw):
...
         TypeError,
-        match=re.escape(
+        match=(
             "`job_retry` was provided, but this job is"
             " not retryable, because a custom `job_id` was"
             " provided to the query that created this job."
-        ),
+        ).replace(" ", r"\s"),
     ):
+        # Expliclty providing a job_retry is an error.
```
shollyman, line 541: the nittiest nit in the world: s/Expliclty/Explicitly/

shollyman (approved):

tests/unit/test_job_retry.py
tswast:
```suggestion
        # Explicitly providing a job_retry is an error.
```

shollyman:
rerunning kokoro, infra issues should be resolved.

PR #2175, opened 2025-05-12
fix(table): fix table constraints equal method

AnasBalk:


chalmerlowe:
The code in this PR is mangled. As such, it is currently unusable. Will see if an updated PR is issued with usable code, otherwise will close as will not fix.

One example:

```
"""str: The file format the table data is stored in. See 
BigLakeFileFormat for available    @file_format.setter
        values."""f, value: Optional[str]):
```

chalmerlowe (changes requested):
This PR cannot be merged as is. The code is mangled. Please issue an updated PR.

chalmerlowe:
The code in this PR is severally mangled. Closing as "will not use".

PR #2176, opened 2025-05-12
feat: Add dtype parameters to to_geodataframe  functions

### LINKED ISSUE #1902: Support string_dtype, etc. in to_geodataframe
Way back in #954 @tswast and I were discussing support for pyarrow-backed strings, that was added by @chelsea-lin in #1529 by way of the `string_dtype` etc arguments.

Would it make sense to add the same thing to the `to_geodataframe` API? Especially now that pandas is moving towards using the arrow backend by default, I think it would be nice to support these types everywhere.

chalmerlowe:
This change adds support for `bool_dtype`, `int_dtype`, `float_dtype`, and `string_dtype` parameters to the `to_geodataframe` method in `RowIterator` and `QueryJob`.

These parameters allow you to specify the desired pandas dtypes for boolean, integer, float, and string columns when converting BigQuery results to GeoDataFrames.

The changes include:
- Updating `RowIterator.to_geodataframe` to accept and pass these dtype parameters to the underlying `to_dataframe` method.
- Updating `QueryJob.to_geodataframe` to accept and pass these dtype parameters to the underlying `RowIterator.to_geodataframe` method.
- Adding unit tests to verify the correct handling of these parameters.

Similar to #1529 
Fixes #1902  🦕

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,7 +61,7 @@ extensions = [
 
 # autodoc/autosummary flags
 autoclass_content = "both"
-autodoc_default_options = {"members": True, "inherited-members": True}
+autodoc_default_options = {"members": True}
 autosummary_generate = True
 
 
@@ -109,7 +109,6 @@ language = None
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 exclude_patterns = [
-    "google/cloud/bigquery_v2/**",  # Legacy proto-based types.
     "_build",
     "**/.nox/**/*",
     "samples/AUTHORING_GUIDE.md",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ ipywidgets = ["ipywidgets >= 7.7.1", "ipykernel >= 6.2.0"]
 geopandas = ["geopandas >= 0.9.0, < 2.0.0", "Shapely >= 1.8.4, < 3.0.0"]
 ipython = ["ipython >= 7.23.1", "bigquery-magics >= 0.6.0"]
 matplotlib = [
-  "matplotlib == 3.9.2; python_version == '3.9'",
+  "matplotlib >= 3.7.1, <= 3.9.2; python_version == '3.9'",
   "matplotlib >= 3.10.3; python_version >= '3.10'",
 ]
 tqdm = ["tqdm >= 4.23.4, < 5.0.0"]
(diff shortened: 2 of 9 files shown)


docs/conf.py
```diff
@@ -61,7 +61,7 @@
 
 # autodoc/autosummary flags
 autoclass_content = "both"
-autodoc_default_options = {"members": True, "inherited-members": True}
```
tswast, line 64: I'm curious about this change. What inherited members were causing problems? IIRC, there's a few methods defined in the base class for jobs that we want to make sure are documented.

pyproject.toml
```diff
@@ -82,7 +82,11 @@ pandas = [
 ipywidgets = ["ipywidgets >= 7.7.1", "ipykernel >= 6.2.0"]
 geopandas = ["geopandas >= 0.9.0, < 2.0.0", "Shapely >= 1.8.4, < 3.0.0"]
 ipython = ["ipython >= 7.23.1", "bigquery-magics >= 0.6.0"]
-tqdm = ["tqdm >= 4.7.4, < 5.0.0"]
+matplotlib = [
+  "matplotlib == 3.9.2; python_version == '3.9'",
```
tswast, line 86: This is quite a restrictive pin. As an "extra" it wouldn't be enforced all the time, but I could see it causing trouble for folks that use the "all" extra. Can we pick a lower bound that's about 2 years old to not force 3.9 users on the latest suppoorted version?

```suggestion
  "matplotlib >= 3.7.1, <= 3.9.2; python_version == '3.9'",
```

testing/constraints-3.9.txt
```diff
@@ -29,4 +29,4 @@ pyarrow==4.0.0
 python-dateutil==2.8.2
 requests==2.21.0
 Shapely==1.8.4
-tqdm==4.7.4
+tqdm==4.23.4
```
tswast, line 32: Need to include the lower bound for matplotlib here.

```suggestion
matplotlib==3.7.1
```

pyproject.toml
```diff
@@ -82,7 +82,11 @@ pandas = [
...
+  "matplotlib >= 3.10.3; python_version >= '3.10'",
+]
+tqdm = ["tqdm >= 4.23.4, < 5.0.0"]
```
tswast, line 89: [No action required] I'm curious. What forced the tqdm upgrade? 4.23.4 is still quite old, so I'm OK with this. I don't think we need to support folks who are stuck in 2016 for 4.7.4.

tswast (approved):

docs/conf.py
antdking: Are there plans to restore the missing docs, such as `reservation` and `job_timeout_ms` on the `*JobConfig` classes?

PR #2181, opened 2025-05-13
fix: fix rows returned when both start_index and page_size are provided

### LINKED ISSUE #2173: QueryJob.result() - pages don't respect the start index
The `start_index` option is ignored when the `page_size` option is used (or possible just multiple pages):

```
start_index = 10
query_job = bigquery.Client().query(
    "SELECT word FROM `bigquery-public-data.samples.shakespeare` LIMIT 50;",
    job_id_prefix="test_query_w_start_index_",
)

# check the normal row iterator
result = query_job.result(start_index=start_index, page_size=10)
print(result.total_rows)        
print(len(list(result)))

# iterate over pages
result = query_job.result(start_index=start_index, page_size=10)
print(result.total_rows)        
print(sum(len(list(page)) for page in result.pages))
```

bigqquery version 3.30.0

Linchin:
TODO:
- [x] move deep copy import
- [x] double check total rows - done, total rows should not be affected by start_index
- [x] unit and system tests

Fixes #2173 🦕

### DIFF
--- a/tests/unit/job/test_query.py
+++ b/tests/unit/job/test_query.py
@@ -1696,6 +1696,9 @@ class TestQueryJob(_Base):
             "schema": {"fields": [{"name": "col1", "type": "STRING"}]},
             "totalRows": "7",
         }
+
+        # Although the result has 7 rows, the response only returns 6, because
+        # start_index is 1.
         tabledata_resource_1 = {
             "totalRows": "7",
             "pageToken": "page_token_1",
--- a/tests/unit/job/test_query.py
+++ b/tests/unit/job/test_query.py
@@ -1682,6 +1682,73 @@ class TestQueryJob(_Base):
             tabledata_list_request[1]["query_params"]["maxResults"], page_size
         )
 
+    def test_result_with_start_index_multi_page(self):
+        # When there are multiple pages of response and the user has set
+        # start_index, we should supply start_index to the server in the first
+        # request. However, in the subsequent requests, we will pass only
+        # page_token but not start_index, because the server only allows one
+        # of them.
+        from google.cloud.bigquery.table import RowIterator
+
+        query_resource = {
+            "jobComplete": True,
+            "jobReference": {"projectId": self.PROJECT, "jobId": self.JOB_ID},
+            "schema": {"fields": [{"name": "col1", "type": "STRING"}]},
+            "totalRows": "7",
+        }
+        tabledata_resource_1 = {
+            "totalRows": "7",
+            "pageToken": "page_token_1",
+            "rows": [
+                {"f": [{"v": "abc"}]},
+                {"f": [{"v": "def"}]},
+                {"f": [{"v": "ghi"}]},
+            ],
+        }
+        tabledata_resource_2 = {
+            "totalRows": "7",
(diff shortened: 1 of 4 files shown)


tests/unit/job/test_query.py
```diff
@@ -1682,6 +1682,75 @@ def test_result_with_start_index(self):
...
+        tabledata_resource_2 = {
+            "totalRows": "7",
+            "pageToken": None,
+            "rows": [
+                {"f": [{"v": "jkl"}]},
+                {"f": [{"v": "mno"}]},
+                {"f": [{"v": "pqe"}]},
+            ],
+        }
+
```
chalmerlowe, line 1717: Please go through the rest of this test and provide some additional comments.

Clarify why we are using `total_rows` of `7`.
Clarify why setting `total_rows` to `7` despite only have six rows of tabledata is ok/reasonable
Clarify what is at the indexed positions:

* `tabledata_list_request_1[1]`
* `tabledata_list_request_2[1]`

tests/unit/job/test_query.py
```diff
@@ -1682,6 +1682,75 @@ def test_result_with_start_index(self):
             tabledata_list_request[1]["query_params"]["maxResults"], page_size
         )
 
+    def test_result_with_start_index_multi_page(self):
+        # When there are multiple pages of response and the user has set
+        # start_index, we should supply start_index to the server in the first
+        # request. However, in the subsequent requests, we will pass only
+        # page_token but not start_index, because the server only allows one
```
chalmerlowe, line 1689: Here we mention sending `page_token`, but nowhere in the test do we confirm that `page_token` is set or that it has been sent, etc.

In fact, while it is correct, with this language in place, it seems weird that in `tabledata_resource_2`, `pagetoken` is set to `None`. The above language implies that we might send something besides `None`. 

Should we revise the language slightly?
Do we mean some other argument/attribute besides `page_token`?

chalmerlowe (changes requested):
Added some comments about the docstrings and comments in the test.
The test feels complicated especially upon first read so it might benefit from a bit more commentary and potentially some rewording.

tests/unit/job/test_query.py
Linchin: In the last message, the page token will be empty because there is no following page.

tests/unit/job/test_query.py
Linchin: > Clarify why we are using total_rows of 7.
Clarify why setting total_rows to 7 despite only have six rows of tabledata is ok/reasonable

Done.

> Clarify what is at the indexed positions:
tabledata_list_request_1[1]
tabledata_list_request_2[1]

There are comments above the code block explaining that these are the first and the second requests that the client sends to the server.

tests/unit/job/test_query.py
chalmerlowe: I am gonna drop this comment.

tests/unit/job/test_query.py
chalmerlowe: I am gonna drop this comment.

chalmerlowe (approved):

PR #2183, opened 2025-05-14
chore: Fix two types of warnings in unit tests

chalmerlowe:
This commit reduces the number of times two warnings appear when running unit tests:

1.  `PytestRemovedIn9Warning` in `tests/unit/test_opentelemetry_tracing.py`: Removed a `@pytest.mark.skipif` decorator from a fixture. The skip condition is already present on the test methods that use the fixture.

#### Example warning:
```
/tmpfs/src/github/python-bigquery/tests/unit/test_opentelemetry_tracing.py:45:
  PytestRemovedIn9Warning: Marks applied to fixtures have no effect
  See docs: https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function
  @pytest.mark.skipif(opentelemetry is None, reason="Require `opentelemetry`")
```    

2.  `FutureWarning` in `tests/unit/test_client.py`: Updated calls to `client.query()` to include `job_retry=None` when `job_id` is also specified. This is to avoid ambiguity as BigQuery cannot retry a failed job with the exact same ID.

#### Example warning:
```
tests/unit/test_client.py::TestClient::test_query_preserving_explicit_default_job_config
  /tmpfs/src/github/python-bigquery/tests/unit/test_client.py:4951:
  FutureWarning: job_retry must be explicitly set to None if job_id is set.
  BigQuery cannot retry a failed job by using the exact
  same ID. Setting job_id without explicitly disabling
  job_retry will raise an error in the future. To avoid this
  warning, either use job_id_prefix instead (preferred) or
  set job_retry=None.
    client.query(query, job_id=job_id, location=self.LOCATION, job_config=None)
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -4719,7 +4719,7 @@ class TestClient(unittest.TestCase):
         client._connection = make_connection({})
 
         with self.assertRaises(TypeError) as exc:
-            client.query(query, job_id="abcd", api_method="QUERY")
+            client.query(query, job_id="abcd", api_method="QUERY", job_retry=None)
         self.assertIn(
             "`job_id` was provided, but the 'QUERY' `api_method` was requested",
             exc.exception.args[0],
@@ -4774,7 +4774,7 @@ class TestClient(unittest.TestCase):
         conn = client._connection = make_connection(resource)
 
         client.query(
-            query, job_id=job_id, project="other-project", location=self.LOCATION
+            query, job_id=job_id, project="other-project", location=self.LOCATION, job_retry=None
         )
 
         # Check that query actually starts the job.
@@ -4833,7 +4833,7 @@ class TestClient(unittest.TestCase):
         original_config_copy = copy.deepcopy(job_config)
 
         client.query(
-            query, job_id=job_id, location=self.LOCATION, job_config=job_config
+            query, job_id=job_id, location=self.LOCATION, job_config=job_config, job_retry=None
         )
 
         # Check that query actually starts the job.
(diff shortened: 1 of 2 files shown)


tests/unit/test_client.py
```diff
@@ -4719,7 +4719,7 @@ def test_query_w_api_method_query_and_job_id_fails(self):
         client._connection = make_connection({})
 
         with self.assertRaises(TypeError) as exc:
-            client.query(query, job_id="abcd", api_method="QUERY")
+            client.query(query, job_id="abcd", api_method="QUERY", job_retry=None)
         self.assertIn(
             "`job_id` was provided, but the 'QUERY' `api_method` was requested",
```
chalmerlowe, line 4724:
```suggestion
            "`job_id` was provided, but the 'QUERY' `api_method` was requested.",
```

tests/unit/test_client.py
```diff
@@ -4719,9 +4719,9 @@ def test_query_w_api_method_query_and_job_id_fails(self):
         client._connection = make_connection({})
 
         with self.assertRaises(TypeError) as exc:
-            client.query(query, job_id="abcd", api_method="QUERY")
+            client.query(query, job_id="abcd", api_method="QUERY", job_retry=None)
         self.assertIn(
-            "`job_id` was provided, but the 'QUERY' `api_method` was requested",
+            "`job_id` was provided, but the 'QUERY' `api_method` was requested.",
```
chalmerlowe, line 4724:
```suggestion
            "`job_id` was provided, but the 'QUERY' `api_method` was requested",
```

Linchin (approved):

PR #2187, opened 2025-05-15
ci: Import numpy before pyarrow in tests to resolve import warning

chalmerlowe:
A `PytestDeprecationWarning` was occurring in several test files because `pyarrow`, when imported by `pytest.importorskip`, would fail to import `numpy.core.multiarray`.

This change addresses the warning by explicitly importing `numpy` before `pytest.importorskip("pyarrow", ...)` in the affected test files. This ensures that numpy is fully initialized before pyarrow attempts to use it, resolving the underlying import error.

```
tests/unit/test_table.py::TestRowIterator::test_to_arrow_w_empty_table
  /repo/python-bigquery/tests/unit/test_table.py:3270: PytestDeprecationWarning: 
  Module 'pyarrow' was found, but when imported by pytest it raised:

      **ImportError('numpy.core.multiarray failed to import')**

  In pytest 9.1 this warning will become an error by default.
  You can fix the underlying problem, or alternatively overwrite this behavior and silence this warning by passing exc_type=ImportError explicitly.
  See https://docs.pytest.org/en/stable/deprecations.html#pytest-importorskip-default-behavior-regarding-importerror
    pyarrow = pytest.importorskip(
```

### DIFF
--- a/tests/unit/test__pandas_helpers.py
+++ b/tests/unit/test__pandas_helpers.py
@@ -16,7 +16,8 @@ import collections
 import datetime
 import decimal
 import functools
-import gc
+
+# import gc
 import operator
 import queue
 from typing import Union
@@ -1887,6 +1888,7 @@ def test__download_table_bqstorage(
     assert queue_used.maxsize == expected_maxsize
 
 
+'''
 @pytest.mark.skipif(isinstance(pyarrow, mock.Mock), reason="Requires `pyarrow`")
 def test__download_table_bqstorage_shuts_down_workers(
     monkeypatch,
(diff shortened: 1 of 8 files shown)


tests/unit/test__pyarrow_helpers.py
```diff
@@ -14,7 +14,7 @@
 
 import pytest
 
-
+import numpy
```
chalmerlowe, line 17:
```suggestion
numpy = pytest.importorskip("numpy")
```

tests/unit/test__pandas_helpers.py
```diff
@@ -16,7 +16,8 @@
 import datetime
 import decimal
 import functools
-import gc
+
+# import gc
```
chalmerlowe, line 20:
```suggestion
```

tswast (approved):

PR #2188, opened 2025-05-15
fix: ensure SchemaField.field_dtype returns a string

chelsea-lin:


google/cloud/bigquery/_pandas_helpers.py
```diff
@@ -122,6 +123,7 @@ def _to_wkb(v):
     "uint16": "INTEGER",
     "uint32": "INTEGER",
     "geometry": "GEOGRAPHY",
+    "string": "STRING",
```
tswast, line 126: This dictionary should only get used if pandas-gbq is not installed. Let's make sure to make the same change here:

https://github.com/googleapis/python-bigquery-pandas/blob/7d354f1ca90eb2647c81d8d8422d0146d56f802a/pandas_gbq/schema/pandas_to_bigquery.py#L29

tswast (approved):

google/cloud/bigquery/_pandas_helpers.py
chelsea-lin: I undid the changes after discovering the root cause: the `dataframe_to_bq_schema` method incorrectly relies on `None` for `SchemaField.field_dtype`, which actually required a string value. Hence, this change refactors the `dataframe_to_bq_schema` a little bit.

tswast (approved):

PR #2189, opened 2025-05-16
ci: Update to the CI/CD pipeline via github workflow to help cut turn-around time

chalmerlowe:
Update to the CI/CD pipeline via github workflow to help cut turn-around time.

* added github workflow
* changed the number of pytest-xdist workers from "auto" to "8" (based on local tests and discussion with Tim, choosing auto sometimes takes longer to run than choosing a smaller number. I suspect this is partly because for small or short tests the overhead needed to setup a worker exceeds the time savings of having extra workers).
* modified numerous tests to explicitly include a project path to avoid an attempt to find the project by making an external call via the pydata-google-auth workflow (which opens an input and waits for response from the user that never comes).


.github/workflows/unittest.yml
```diff
@@ -0,0 +1,89 @@
...
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install nox
+      run: |
+        python -m pip install --upgrade setuptools pip wheel
+        python -m pip install nox
+    - name: Run unit_noextras tests
+      env:
+        COVERAGE_FILE: .coverage-${{ matrix.python }}
```
chalmerlowe, line 55:
```suggestion
        COVERAGE_FILE: .coverage-unit-noextras-${{ matrix.python }}
```

noxfile.py
```diff
@@ -128,8 +128,9 @@ def default(session, install_extras=True):
     # Run py.test against the unit tests.
     session.run(
         "py.test",
-        "-n=auto",
-        "--quiet",
+        # "-n=8",
+        # "--quiet",
+        "-vv",
```
chalmerlowe, line 133:
```suggestion
        "-n=8",
        "--quiet",
```

Linchin (approved):
LGTM, thanks.

PR #2190, opened 2025-05-19
feat: job creation mode GA

shollyman:
This PR makes the underlying functionality related to how queries can optionally avoid job creation a GA feature.  It does the following:

* no longer uses the preview QUERY_PREVIEW_ENABLED environment variable to control job creation
* adds a new argument to Client instantiation to control job creation mode
* adds a property/setter to Client to control job creation mode

This PR also updates/renames the sample demonstrating how to leverage job creation mode with Client.query_and_wait.

internal: b/396766171

### DIFF
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -282,7 +282,7 @@ class Client(ClientWithProject):
         self._connection = Connection(self, **kw_args)
         self._location = location
         self._default_load_job_config = copy.deepcopy(default_load_job_config)
-        self._default_job_creation_mode = default_job_creation_mode
+        self.default_job_creation_mode = default_job_creation_mode
 
         # Use property setter so validation can run.
         self.default_query_job_config = default_query_job_config
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -4142,7 +4142,7 @@ class Client(ClientWithProject):
                 rows that were affected.
             query (Optional[str]):
                 The query text used.
-            total_bytes_processed (Optinal[int]):
+            total_bytes_processed (Optional[int]):
                 total bytes processed from job statistics, if present.
 
         Returns:
--- a/google/cloud/bigquery/client.py
+++ b/google/cloud/bigquery/client.py
@@ -221,6 +221,10 @@ class Client(ClientWithProject):
         client_options (Optional[Union[google.api_core.client_options.ClientOptions, Dict]]):
             Client options used to set user options on the client. API Endpoint
             should be set through client_options.
+        default_job_creation_mode (Optional[str]):
+            Sets the default job creation mode used by query methods such as
+            query_and_wait().  For lightweight queries, JOB_CREATION_OPTIONAL is
+            generally recommended.
 
     Raises:
         google.auth.exceptions.DefaultCredentialsError:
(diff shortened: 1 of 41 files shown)


shollyman:
re: sample tag change, will do the freeze and rename dance on the docs side when we're ready.

google/cloud/bigquery/client.py
```diff
@@ -286,6 +292,15 @@ def location(self):
         """Default location for jobs / datasets / tables."""
         return self._location
 
+    @property
+    def default_job_creation_mode(self):
+        """Default job creation mode used for query execution."""
+        return self._default_job_creation_mode
```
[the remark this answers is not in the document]
tswast: Chatted offline. I'll aim to get bigframes 1.x and 2.x releases out to support this change before the client library release to minimize disruption.

samples/client_query_job_optional.py
```diff
@@ -44,10 +46,12 @@ def client_query_shortmode() -> None:
     if rows.job_id is not None:
         print("Query was run with job state.  Job ID: {}".format(rows.job_id))
     else:
-        print("Query was run in short mode.  Query ID: {}".format(rows.query_id))
+        print(
+            "Query was run without creating a job..  Query ID: {}".format(rows.query_id)
```
Linchin, line 50:
```suggestion
            "Query was run without creating a job.  Query ID: {}".format(rows.query_id)
```

tests/unit/test__job_helpers.py
```diff
@@ -605,6 +601,7 @@ def test_query_and_wait_sets_job_creation_mode(monkeypatch: pytest.MonkeyPatch):
 
 def test_query_and_wait_sets_location():
     client = mock.create_autospec(Client)
+    client.default_job_creation_mode = None
```
Linchin, line 604: Just for my education, why do we need to set `client.default_job_creation_mode = None` explicitly? It seems using default value would be enough.

google/cloud/bigquery/client.py
```diff
@@ -286,6 +292,15 @@ def location(self):
...
+
+    @default_job_creation_mode.setter
```
Linchin, line 300: nit: add unit test for the property

tswast (approved):

google/cloud/bigquery/client.py
shollyman: done

tests/unit/test__job_helpers.py
shollyman: Originally I tried to keep the asserted request unchanged, but I've instead made this less confusing and used mock.ANY for the jobCreationMode aspect of the request.

samples/tests/test_client_query_job_optional.py
```diff
@@ -14,13 +14,13 @@
...
-from .. import client_query_shortmode
+from .. import client_query_job_optional
 
 if typing.TYPE_CHECKING:
     import pytest
 
 
 def test_client_query_shortmode(capsys: "pytest.CaptureFixture[str]") -> None:
-    client_query_shortmode.client_query_shortmode()
+    client_query_job_optional.client_query_job_optional()
```
Linchin, line 24: Should we add some verification that the query was indeed run in jobless mode?

Linchin (approved):

samples/tests/test_client_query_job_optional.py
shollyman: Control over execution is still the responsible of the service, and it can choose to fallback to using jobs regardless of the requested mode.

PR #2191, opened 2025-05-19
Revert "fix: table iterator should not use bqstorage when page_size is not None (#2154)"

chelsea-lin:
Fixes internal issue 418814848🦕

reverts changes introduced by https://github.com/googleapis/python-bigquery/pull/2154

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -1897,11 +1897,6 @@ class RowIterator(HTTPIterator):
         """total bytes processed from job statistics, if present."""
         return self._total_bytes_processed
 
-    @property
-    def page_size(self) -> Optional[int]:
-        """The maximum number of rows in each page of results from this request, if present."""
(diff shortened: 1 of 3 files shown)


tswast (approved):

shollyman (approved):

Linchin:
Will merge after we confirm the action plan

google/cloud/bigquery/table.py
```diff
@@ -1897,11 +1897,6 @@ def total_bytes_processed(self) -> Optional[int]:
         """total bytes processed from job statistics, if present."""
         return self._total_bytes_processed
 
-    @property
-    def page_size(self) -> Optional[int]:
-        """The maximum number of rows in each page of results from this request, if present."""
-        return self._page_size
-
```
tswast, line 1904: This would technically be a breaking change, but maybe it's OK since this property was only present in one version?

PR #2199, opened 2025-05-23
ci: Configure Renovate to keep Python at 3.10 for docs workflow

chalmerlowe:
When Renovate Bot checks for potential updates to dependency versions, it is currently updating a file that it should not be looking at: `docs.yml` and it is updating a `matplotlib` dependency in `pyproject.toml` that we need to have locked down tight when using Python 3.9.

This PR puts `docs.yml` on an excludes list so that Owlbot will ignore it.
It adds a regex line to `renovate.json` to ensure that the `matplotlib` dependency does not change.

### DIFF
--- a/owlbot.py
+++ b/owlbot.py
@@ -76,7 +76,7 @@ s.move(
         ".kokoro/continuous/prerelease-deps.cfg",
         ".kokoro/samples/python3.7/**",
         ".kokoro/samples/python3.8/**",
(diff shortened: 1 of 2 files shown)


chalmerlowe:
The failing tests in `unittest / unit (3.13)` associated with `test_magics.py` are a known issue and are being addressed via a different PR.

leahecole (approved):
✅  approving in the name of science and experimenting to get these configs just right! (once tests pass)

PR #2229, opened 2025-07-02
feat: adds time_zone to external config and load job

chalmerlowe:
This commit introduces new configuration options for BigQuery load jobs and external table definitions, aligning with recent updates to the underlying protos.

New options added:

`time_zone`: Time zone used when parsing timestamp values that do not have specific time zone information. (Applies to `LoadJobConfig`, `LoadJob`, and `ExternalConfig`)

Changes include:

Added corresponding properties (getters/setters) to `LoadJobConfig`, `LoadJob`, and `ExternalConfig`.
Updated docstrings and type hints for all new attributes.
Updated unit tests to cover the new options, ensuring they are correctly handled during object initialization, serialization to API representation, and deserialization from API responses.

### DIFF
--- a/tests/unit/job/test_load.py
+++ b/tests/unit/job/test_load.py
@@ -450,7 +450,7 @@ class TestLoadJob(_Base):
         job = klass.from_api_repr(RESOURCE, client)
         api_repr = job.to_api_repr()
 
-        # as per the documentation in load.py > LoadJob.to_api_repr(),
+        # as per the documentation in load.py -> LoadJob.to_api_repr(),
         # the return value from to_api_repr should not include statistics
         expected = {
             "jobReference": RESOURCE["jobReference"],
--- a/tests/unit/job/test_load.py
+++ b/tests/unit/job/test_load.py
@@ -441,6 +441,24 @@ class TestLoadJob(_Base):
         self.assertIs(job._client, client)
         self._verifyResourceProperties(job, RESOURCE)
 
+    def test_to_api_repr(self):
+        self._setUpConstants()
+        client = _make_client(project=self.PROJECT)
+        RESOURCE = self._make_resource(ended=False)
+
+        klass = self._get_target_class()
+        job = klass.from_api_repr(RESOURCE, client)
+        api_repr = job.to_api_repr()
+
+        # as per the documentation in load.py > LoadJob.to_api_repr(),
+        # the return value from to_api_repr should not include statistics
+        expected = {
+            "jobReference": RESOURCE["jobReference"],
+            "configuration": RESOURCE["configuration"],
+        }
+
+        self.assertEqual(api_repr, expected)
+
     def test_begin_w_already_running(self):
         conn = make_connection()
         client = _make_client(project=self.PROJECT, connection=conn)
--- a/tests/unit/job/test_load.py
+++ b/tests/unit/job/test_load.py
@@ -38,10 +38,14 @@ class TestLoadJob(_Base):
         self.OUTPUT_ROWS = 345
         self.REFERENCE_FILE_SCHEMA_URI = "gs://path/to/reference"
 
+        self.TIME_ZONE = "UTC"
+
     def _make_resource(self, started=False, ended=False):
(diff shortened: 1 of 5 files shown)


tests/unit/test_external_config.py
Linchin, line 1: In addition to testing to/from_api_repr, I think we also need to test against the class property itself.

tests/unit/job/test_load_config.py
Linchin, line 1: Similar to https://github.com/googleapis/python-bigquery/pull/2229/files#r2191183829, I think in general we need to test against to/from_api_repr and the property itself. The same goes for `LoadJobConfig` and `LoadJob`. Could you add these changes? Otherwise the PR looks good in general.

chalmerlowe: The pattern/philosophy for `test_load_config.py` is different than the approach used in `test_external_config.py`

Here, the author chose to test each attribute in more of a stand alone fashion using a trifecta of tests:

* `x_missing()`
* `x_hit()`
* `x_setter()`

But:
the author did not test `to_api_repr()` or `from_api_repr()` directly (I can add such a pair of tests).

Examples:

* `def test_write_disposition_missing()`
* `def test_write_disposition_hit()`
* `def test_write_disposition_setter()`
* `def test_time_zone_missing()`
* `def test_time_zone_hit()`
* `def test_time_zone_setter()`

#### missing
Confirms that when the target class is created, the attribute is not populated
```
self.assertIsNone(config.write_disposition)`
```

#### hit
Sets (behind the scenes using `._properties`) the value of the attribute
Confirms that the **getter** functions correctly and that the expected/result values match

```
config._properties["load"]["writeDisposition"] = write_disposition
self.assertEqual(config.write_disposition, write_disposition)
```

#### setter
Sets the value of the attribute directly using the setter function
Confirms that the **setter** functions correctly by examining the `_properties` object directly

```
config.write_disposition = write_disposition
self.assertEqual(config._properties["load"]["writeDisposition"], write_disposition)
```

tests/unit/test_external_config.py
chalmerlowe: Regarding adding a `to_api_repr` and `from_api_repr` pair, there may be some merit to creating a single test that includes all the attributes as part of a `BASE_RESOURCE` much as we see in `test_external_config.py`

tests/unit/job/test_load_config.py
```diff
@@ -10,8 +10,7 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
-
+# limitations under the License.n
```
chalmerlowe, line 13:
```suggestion
# limitations under the License.

```

tests/unit/job/test_load.py
```diff
@@ -38,10 +38,14 @@ def _setUpConstants(self):
         self.OUTPUT_ROWS = 345
```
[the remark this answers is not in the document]
chalmerlowe: I added a `test_to_api_repr()` test because the original file did not include one.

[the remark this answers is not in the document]
Linchin: Thank you for adding the tests!

Linchin (dismissed):

tests/unit/job/test_load.py
```diff
@@ -431,6 +441,24 @@ def test_from_api_repr_w_properties(self):
...
+    def test_to_api_repr(self):
+        self._setUpConstants()
+        client = _make_client(project=self.PROJECT)
+        RESOURCE = self._make_resource(ended=False)
+
+        klass = self._get_target_class()
+        job = klass.from_api_repr(RESOURCE, client)
+        api_repr = job.to_api_repr()
+
+        # as per the documentation in load.py > LoadJob.to_api_repr(),
```
chalmerlowe, line 453:
```suggestion
        # as per the documentation in load.py -> LoadJob.to_api_repr(),
```

Linchin (approved):

PR #2260, opened 2025-08-12
feat: add created/started/ended properties to RowIterator.

### LINKED ISSUE #2261: Expose Job Lifecycle Timestamps (created, started, ended) on RowIterator
#Motivation
The BigFrames library uses the client.query_and_wait() method, which returns a google.cloud.bigquery.table.RowIterator object. This RowIterator object does not currently expose the lifecycle timestamps (created, started, ended) from its underlying QueryJob.

These timestamps are required for performance evaluation and optimization during the development of BigFrames, as they are used to measure the end-to-end latency of operations.

#Proposed Solution
Add three read-only properties to the RowIterator class that delegate to the underlying job object:

- created: The job's creation time (job.created).
- 
- started: The job's start time (job.started).
- 
- ended: The job's end time (job.ended).

Genesis929:

Fixes #2261🦕

### DIFF
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -857,30 +857,6 @@ class QueryJob(_AsyncJob):
         return typing.cast(QueryJobConfig, super().configuration)
 
     @property
-    def created(self):
-        """Datetime at which the job was created."""
-        millis = _helpers._int_or_none(self._job_statistics().get("creationTime"))
-        if millis is not None:
-            return _helpers._datetime_from_microseconds(millis * 1000)
-        return super().created
-
-    @property
-    def started(self):
-        """Datetime at which the job was started."""
-        millis = _helpers._int_or_none(self._job_statistics().get("startTime"))
-        if millis is not None:
-            return _helpers._datetime_from_microseconds(millis * 1000)
-        return super().started
-
-    @property
-    def ended(self):
-        """Datetime at which the job finished."""
-        millis = _helpers._int_or_none(self._job_statistics().get("endTime"))
-        if millis is not None:
-            return _helpers._datetime_from_microseconds(millis * 1000)
-        return super().ended
-
-    @property
     def connection_properties(self) -> List[ConnectionProperty]:
         """See
         :attr:`google.cloud.bigquery.job.QueryJobConfig.connection_properties`.
--- a/google/cloud/bigquery/job/query.py
+++ b/google/cloud/bigquery/job/query.py
@@ -857,6 +857,30 @@ class QueryJob(_AsyncJob):
         return typing.cast(QueryJobConfig, super().configuration)
 
     @property
+    def created(self):
+        """Datetime at which the job was created."""
+        millis = _helpers._int_or_none(self._job_statistics().get("creationTime"))
+        if millis is not None:
+            return _helpers._datetime_from_microseconds(millis * 1000)
+        return super().created
+
+    @property
+    def started(self):
+        """Datetime at which the job was started."""
+        millis = _helpers._int_or_none(self._job_statistics().get("startTime"))
+        if millis is not None:
+            return _helpers._datetime_from_microseconds(millis * 1000)
+        return super().started
+
+    @property
+    def ended(self):
+        """Datetime at which the job finished."""
+        millis = _helpers._int_or_none(self._job_statistics().get("endTime"))
+        if millis is not None:
+            return _helpers._datetime_from_microseconds(millis * 1000)
+        return super().ended
+
+    @property
     def connection_properties(self) -> List[ConnectionProperty]:
         """See
         :attr:`google.cloud.bigquery.job.QueryJobConfig.connection_properties`.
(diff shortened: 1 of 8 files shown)


google/cloud/bigquery/query.py
```diff
@@ -1310,6 +1310,50 @@ def num_dml_affected_rows(self):
         if num_dml_affected_rows is not None:
             return int(num_dml_affected_rows)
 
+    @property
+    def created(self) -> Optional[datetime.datetime]:
+        """Datetime at which the job was created.
+
+        See:
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobStatistics.FIELDS.creation_time
```
tswast, line 1318: I don't think this is available on the Query response object, is it?

tswast: Please update the link to the correct object: http://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#body.QueryResponse.FIELDS.creation_time

google/cloud/bigquery/job/query.py
```diff
@@ -856,6 +856,30 @@ def configuration(self) -> QueryJobConfig:
...
+        return super().started
+
+    @property
+    def ended(self):
+        """Datetime at which the job finished."""
+        millis = _helpers._int_or_none(self._job_statistics().get("endTime"))
+        if millis is not None:
+            return _helpers._datetime_from_microseconds(millis * 1000)
+        return super().ended
+
```
tswast, line 882: I'm pretty sure these are already defined in the base class.

See: https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJob#google_cloud_bigquery_job_QueryJob_created

and

https://github.com/googleapis/python-bigquery/blob/3deff1d963980800e8b79fa3aaf5b712d4fd5062/google/cloud/bigquery/job/base.py#L587-L623

```suggestion
```

google/cloud/bigquery/query.py
```diff
@@ -1310,6 +1310,51 @@ def num_dml_affected_rows(self):
         if num_dml_affected_rows is not None:
             return int(num_dml_affected_rows)
 
+    @property
+    def created(self):
+        """Datetime at which the job was created.
```
tswast, line 1315: There might not actually be a "job" associated with the query. Let's use the same description as the REST API docs use.

```suggestion
        """Creation time of this query.
```

google/cloud/bigquery/query.py
```diff
@@ -1310,6 +1310,51 @@ def num_dml_affected_rows(self):
...
+            Optional[datetime.datetime]:
+                the creation time (None until set from the server).
+        """
+        millis = self._properties.get("creationTime")
+        if millis is not None:
+            return _helpers._datetime_from_microseconds(int(millis) * 1000.0)
+
+    @property
+    def started(self):
+        """Datetime at which the job was started.
```
tswast, line 1330: Same here:

```suggestion
        """Start time of this query.

        This field will be present when the query transitions from the
        PENDING state to either RUNNING or DONE.
```

google/cloud/bigquery/query.py
```diff
@@ -1310,6 +1310,51 @@ def num_dml_affected_rows(self):
...
+            Optional[datetime.datetime]:
+                the start time (None until set from the server).
+        """
+        millis = self._properties.get("startTime")
+        if millis is not None:
+            return _helpers._datetime_from_microseconds(int(millis) * 1000.0)
+
+    @property
+    def ended(self):
+        """Datetime at which the job finished.
```
tswast, line 1345: Same here:

```suggestion
        """End time of this query.

        This field will be present whenever a query is in the DONE state.
```

google/cloud/bigquery/table.py
```diff
@@ -1789,6 +1789,14 @@ class RowIterator(HTTPIterator):
...
         total_bytes_processed (Optional[int]):
             total bytes processed from job statistics, if present.
+        slot_millis (Optional[int]):
+            Number of slot ms the user is actually billed for.
+        created (Optional[datetime.datetime]):
+            Datetime at which the job was created.
+        started (Optional[datetime.datetime]):
+            Datetime at which the job was started.
+        ended (Optional[datetime.datetime]):
+            Datetime at which the job finished.
```
tswast, line 1799: RowIterator is also used for `list_rows(table)`, so let's clarify that this is for the associated query, if present.

```suggestion
        total_bytes_processed (Optional[int]):
            If representing query results, the total bytes processed by the associated query.
        slot_millis (Optional[int]):
            If representing query results, the number of slot ms billed for the associated query.
        created (Optional[datetime.datetime]):
            If representing query results, the creation time of the associated query.
        started (Optional[datetime.datetime]):
            If representing query results, the start time of the associated query.
        ended (Optional[datetime.datetime]):
            If representing query results, the end time of the associated query.
```

google/cloud/bigquery/table.py
```diff
@@ -1905,6 +1919,21 @@ def slot_millis(self) -> Optional[int]:
...
+        return self._job_created
+
+    @property
+    def started(self) -> Optional[datetime.datetime]:
+        """Datetime at which the job was started."""
+        return self._job_started
+
+    @property
+    def ended(self) -> Optional[datetime.datetime]:
+        """Datetime at which the job finished."""
```
tswast, line 1934: Same here.

```suggestion
        """If representing query results, the creation time of the associated query."""
        return self._job_created

    @property
    def started(self) -> Optional[datetime.datetime]:
        """If representing query results, the start time of the associated query."""
        return self._job_started

    @property
    def ended(self) -> Optional[datetime.datetime]:
        """If representing query results, the end time of the associated query."""
```

tswast (approved):
Thanks!

PR #2268, opened 2025-08-25
feat: updates to fastpath query execution

shollyman:
This PR updates query handling to allow base config properties like job timeout, reservation, and a preview max slots field to leverage the faster path (e.g. using jobs.query rather than jobs.insert).

### DIFF
--- a/google/cloud/bigquery/job/base.py
+++ b/google/cloud/bigquery/job/base.py
@@ -225,6 +225,37 @@ class _JobConfig(object):
             self._properties.pop("jobTimeoutMs", None)
 
     @property
+    def max_slots(self) -> Optional[int]:
+        """The maximum rate of slot consumption to allow for this job.
+
+        If set, the number of slots used to execute the job will be throttled
+        to try and keep its slot consumption below the requested rate.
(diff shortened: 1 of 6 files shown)


google/cloud/bigquery/job/base.py
```diff
@@ -224,6 +224,37 @@ def job_timeout_ms(self, value):
...
+            value = _int_or_none(value)
+        except ValueError as err:
+            raise ValueError("Pass an int for max slots, e.g. 100").with_traceback(
+                err.__traceback__
+            )
+
+        if value is not None:
+            self._properties["maxSlots"] = str(value)
+        else:
+            self._properties.pop("maxSlots", None)
```
chalmerlowe, line 256: #QUESTION 

What is the impetus for `.pop()`?

If `value` is `None`, delete the `key: value` pair entirely?
I know we also did it with `jobTimeoutMs` and I don't get it there either.

chalmerlowe (dismissed):
Just one question to cure my curiosity. 

Approved.

chalmerlowe (approved):
LGTM

PR #2270, opened 2025-08-26
feat: add additional query stats

shollyman:
This PR adds support for incremental query stats.


chalmerlowe (dismissed):
LGTM, with one question. See inline comment.

If you have to make a change, ping me, I will come back and reapprove.

mrfaizal (approved):

PR #2273, opened 2025-08-28
feat: update BigQueryClient methods

chalmerlowe:
This PR:

* Adds several helper functions that enable processing of convenience strings (e.g. `project_id.dataset_id`
* Updates the setters and getters for the `*SERVICECLIENT` attributes
* Updates docstrings and comments across the board
* Updates TWO methods associated with the `*SERVICECLIENTs` to allow discussion about the approach. The methods take in either a request object OR a convenience string as an argument and transmogrify the input so that the underlying service client attribute can be called directly. e.g.:
  * `get_dataset()`
  * `list_datasets()`

### NOTES 

1. Getting the approach to these service clients and methods is important because the microgenerator will be updated to create seven clients and 30+ methods.
1. The **GitHub Actions** are being used to help ensure that unit tests pass.
Screenshot 2025-08-20 at 9 14 08 AM<br>
1. The **KOKORO tests** are failing. This is a **known problem** and will be dealt with in a separate PR. It should not affect merging into the `autogen` dev branch.
Screenshot 2025-08-20 at 9 13 45 AM<br>


tests/unit/gapic/bigquery_v2/test_centralized_service.py
```diff
@@ -162,122 +138,148 @@ def test_client_initialization_arguments(
...
-        expected_model = model.Model(
-            etag=DEFAULT_ETAG,
-            model_reference={
-                "project_id": PROJECT_ID,
-                "dataset_id": DATASET_ID,
-                "model_id": MODEL_ID,
-            },
-        )
-        mock_model_service_client.get_model.return_value = expected_model
+        mock_instance = mock_model_service_client_class.return_value
```
chalmerlowe, line 250:
```suggestion
        expected_model = model.Model(
            etag=DEFAULT_ETAG,
            model_reference={
                "project_id": PROJECT_ID,
                "dataset_id": DATASET_ID,
                "model_id": MODEL_ID,
            },
        )
        mock_instance = mock_model_service_client_class.return_value
```

google/cloud/bigquery_v2/services/centralized_service/_helpers.py
```diff
@@ -14,11 +14,87 @@
...
+            identifier_value=dataset_id,
+            identifier_name="dataset_id",
+            parser=self._parse_dataset_id_to_dict,
+            identifier_required=True,
+        )
+    """
+    if user_request is not None and identifier_value is not None:
+        raise ValueError(
+            f"Provide either a request object or '{identifier_name}', not both."
+        )
```
tswast, line 86: IIRC, there are cases where we merge these two in the existing hand-written client. For example, load jobs can take a string as a destination but merge in the job config object to the final request:

https://github.com/googleapis/python-bigquery/blob/ef2740a158199633b5543a7b6eb19587580792cd/google/cloud/bigquery/client.py#L2577-L2590

I actually haven't thought too much about how the non-query jobs fit into this design, though. I suppose the user needs to specify more than just an identifier for all of the job types, so this method wouldn't apply?

Note: In addition to query jobs, load jobs using jobs.insert REST API will need a bit of handwritten magic to support load from local data via "resumable media uploads" (https://cloud.google.com/bigquery/docs/reference/api-uploads). I imagine we're planning on providing a separate hand-written helper for this, similar to queries? Actually do we know if the GAPICs even support the resumable upload API? AFAIK, it's only used in BigQuery, Cloud Storage, and Google Drive APIs. CC @parthea

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
     OptionalRetry = Union[retries.Retry, gapic_v1.method._MethodDefault, None]
 except AttributeError:  # pragma: NO COVER
     OptionalRetry = Union[retries.Retry, object, None]  # type: ignore
 
-# TODO: This line is here to simplify prototyping, etc.
-PROJECT_ID = os.environ.get("GOOGLE_CLOUD_PROJECT")
+DatasetIdentifier = Union[str, dataset_reference.DatasetReference]
 
+# TODO: This variable is here to simplify prototyping, etc.
+PROJECT_ID = os.environ.get("GOOGLE_CLOUD_PROJECT")
```
tswast, line 57: Note: When using `google.auth.default()`, it will read this variable for you.

https://github.com/googleapis/google-auth-library-python/blob/c30a6a781d3e385598a0ac28a370a7f4800010cc/google/auth/environment_vars.py#L18

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
+                client will attempt to use the default credentials.
+            client_options:
+                A dictionary of client options to pass to the underlying
+                service clients.
+        """
+
+        self._clients: Dict[str, object] = {}
         self._credentials = credentials
         self._client_options = client_options
+        self.project = PROJECT_ID
```
tswast, line 87: Might want to upgrade this to a `@property` so that it can be included in public documentation (and also be suggested to be read-only).

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
+            credentials:
+                The credentials to use for authentication. If not provided, the
+                client will attempt to use the default credentials.
+            client_options:
+                A dictionary of client options to pass to the underlying
+                service clients.
+        """
+
+        self._clients: Dict[str, object] = {}
         self._credentials = credentials
```
tswast, line 85: We'll want to call `credentials, default_project_id = google.auth.default()` (https://googleapis.dev/python/google-auth/latest/reference/google.auth.html#google.auth.default) if the credentials aren't set so that we don't repeat the auth flow for each sub-client construction.

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
+        """
+        Initializes the BigQueryClient.
+
+        Args:
+            credentials:
+                The credentials to use for authentication. If not provided, the
+                client will attempt to use the default credentials.
+            client_options:
+                A dictionary of client options to pass to the underlying
+                service clients.
```
tswast, line 81: The type above also allows for a single `client_options`.

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
 
-
 # Create Centralized Client
 class BigQueryClient:
+    """A centralized client for BigQuery API."""
+
     def __init__(
         self,
         *,
         credentials: Optional[auth_credentials.Credentials] = None,
```
tswast, line 69: In the past, `project` is a pretty common input to the client constructor. This is especially important in environments like https://colab.research.google.com that don't have a project associated with the environment.

Likewise, `client_info` is another important one. That's how applications can amend the user-agent, which is very important for BQ to track usage attributable to open source "connectors" as well as partner company usage. I suppose we could guide such users to use the raw clients then, but if so maybe `client_options` falls in the same bucket of "advanced" features and should be excluded?

Speaking of commonly used arguments: Fun fact, the Kaggle team was (is?) using `_http` to provide a custom proxy to BigQuery for their special Kaggle free tier that included BQML, unlike the BQ Sandbox. Might be worth reaching out to them to come up with an alternative if they're still doing that. See: https://www.kaggle.com/discussions/general/50159 Note that 5 TB is 5x more than the BQ Sandbox 1 TB. That post was from 8 years ago, though, so I don't know if that still applies.

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
+
+        Args:
+            dataset_path: A string in the format 'project_id.dataset_id' or
+                'dataset_id'.
+
+        Returns:
+            A tuple of (project_id, dataset_id).
+        """
+        if "." in dataset_path:
+            project_id, dataset_id = dataset_path.split(".", 1)
```
tswast, line 102: Neat use of the `maxsplit` argument! If we do this, I would actually recommend using `rsplit` because there are some legacy project IDs like `google.com:my-project`, so to specify the dataset as a string it could look like `google.com:my-project.my_dataset`, so we'd want to split on the right-most `.`.

```suggestion
            # Use rsplit to handle legacy paths like `google.com:my-project.my_dataset`.
            project_id, dataset_id = dataset_path.rsplit(".", 1)
```

Note: I'd love to see a unit test covering some of these legacy-format IDs, too.

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
 
+        Args:
+            dataset_id: A string or DatasetReference.
+
+        Returns:
+            A dict of {"project_id": project_id, "dataset_id": dataset_id_str }.
+        """
+        if isinstance(dataset_id, str):
+            project_id, dataset_id_str = self._parse_dataset_path(dataset_id)
+            return {"project_id": project_id, "dataset_id": dataset_id_str}
```
tswast, line 118: Curious to see `snake_case` here. IIRC the REST API mostly uses camel-case, but per https://protobuf.dev/programming-guides/json/ both should be acceptable.

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
+        elif isinstance(dataset_id, dataset_reference.DatasetReference):
```
tswast, line 119: In this case, can't we use the https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.MessageToDict or the proto-plus equivalent?

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,198 @@
...
+            A tuple of (project_id, dataset_id).
+        """
+        if "." in dataset_path:
+            project_id, dataset_id = dataset_path.split(".", 1)
+            return project_id, dataset_id
+        return self.project, dataset_path
+
+    def _parse_dataset_id_to_dict(self, dataset_id: DatasetIdentifier) -> dict:
+        """
+        Helper to create a request dictionary from a project_id and dataset_id.
```
tswast, line 108: By request dictionary, do we mean the canonical JSON representation of the protobuf? https://protobuf.dev/programming-guides/json/

google/cloud/bigquery_v2/services/centralized_service/client.py
chalmerlowe: Thank you for these insights into what customers commonly pass into the client constructor.

I can make a note of these for later inclusion in the beta or release candidates.

Due to schedules and work load, I don't think I can include these in the alpha.

chalmerlowe: Including `project` is handled in this [separate PR](https://github.com/googleapis/python-bigquery/pull/2276).

For the moment, I have not made any attempts to also process `client_info`.

google/cloud/bigquery_v2/services/centralized_service/client.py
```diff
@@ -43,139 +45,200 @@
...
+
+        Args:
+            dataset_path: A string in the format 'project_id.dataset_id' or
+                'dataset_id'.
+
+        Returns:
+            A tuple of (project_id, dataset_id).
+        """
+        if "." in dataset_path:
+            # Use rsplit to handle legacy paths like `google.com:my-project.my_dataset`.
```
Linchin, line 102: For my education, is `google.com:my-project.my_dataset` supposed to yield `['google.com:my-project', 'my_dataset']` or `['my-project', 'my_dataset']`?

PR #2274, opened 2025-09-03
bug: updates `__eq__` comparison on TableConstraint

### LINKED ISSUE #2263: TableConstraints __eq__ is not comprehensive
The TableConstraints class lists `primary_key` and `foreign_keys` arguments as optional. However, the __eq__ method does not account for instances in which primary key and/or foreign_keys is None.
https://github.com/googleapis/python-bigquery/blob/3deff1d963980800e8b79fa3aaf5b712d4fd5062/google/cloud/bigquery/table.py#L3549-L3551.

The eq method should account for instances in which a key for self AND other are None.

#### Reprex

``` python
pk_name = "my_pk"
pk = bigquery.table.PrimaryKey(columns = [pk_name])

fk_name = "my_fk"
column_references = []
table_ref = bigquery.TableReference.from_string("my-project.mydataset.mytable")
fk = bigquery.table.ForeignKey(name = fk_name, referenced_table = table_ref, column_references = column_references)

pk_and_fk_constraint = bigquery.table.TableConstraints(
        primary_key = pk,
        foreign_keys = fk
    )

print(pk_and_fk_constraint == pk_and_fk_constraint) # True

only_pk_constraint = bigquery.table.TableConstraints(
        primary_key = pk,
        foreign_keys = None
    )

print(only_pk_constraint == only_pk_constraint) # None

only_fk_constraint = bigquery.table.TableConstraints(
        primary_key = None,
        foreign_keys = fk
    )

print(only_fk_constraint == only_fk_constraint) # None

empty_constraint = bigquery.table.TableConstraints(
        primary_key = None,
        foreign_keys = None
    )

print(empty_constraint == empty_constraint) # None

```

I'm unable to find any other __eq__ methods in the package than include conditional logic for None, perhaps this was intended to solve for optional parameters specified as None? If so, the paratheses need to be moved to only evaluate the rhs.

### Fix
``` python 
       return self.primary_key == (other.primary_key if other.primary_key else None) and self.foreign_keys == (other.foreign_keys if other.foreign_keys else None)
````

chalmerlowe:
Fixes the `__eq__()` dunder method on `TableConstraints`.
Previously it would give incorrect results (e.g. a return value of `None` instead of `False`)

* Shifts the parentheses to ensure that the correct comparisons are being done.

Fixes #2263 🦕

### DIFF
--- a/google/cloud/bigquery/table.py
+++ b/google/cloud/bigquery/table.py
@@ -3574,10 +3574,9 @@ class TableConstraints:
     def __eq__(self, other):
         if not isinstance(other, TableConstraints) and other is not None:
             raise TypeError("The value provided is not a BigQuery TableConstraints.")
-        return (
-            self.primary_key == (other.primary_key if other.primary_key else None)
-            and self.foreign_keys == (other.foreign_keys if other.foreign_keys else None)
-        )
+        return self.primary_key == (
+            other.primary_key if other.primary_key else None
+        ) and self.foreign_keys == (other.foreign_keys if other.foreign_keys else None)
 
     @classmethod
     def from_api_repr(cls, resource: Dict[str, Any]) -> "TableConstraints":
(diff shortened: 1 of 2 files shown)


iennae (approved):
LTGM; subtle logic issue and the improved tests would catch this problem and fully test edge cases. Thanks!

tswast (approved):

PR #2280, opened 2025-09-05
fix: remove deepcopy while setting properties for _QueryResults

### LINKED ISSUE #2279: Memory spike caused by deepcopy in _QueryResult
See internal issue 442629997.

While calling `Client.query_and_wait()`, we could see a roughly 200MB memory spike, and it is caused by the [helper function that processes API response](https://github.com/googleapis/python-bigquery/blob/8a13c12905ffcb3dbb6086a61df37556f0c2cd31/google/cloud/bigquery/_job_helpers.py#L346-L351) triggering a [deepcopy() while setting values in `_QueryResult`](https://github.com/googleapis/python-bigquery/blob/8a13c12905ffcb3dbb6086a61df37556f0c2cd31/google/cloud/bigquery/query.py#L1403).

Image

Linchin:
See internal issue 442629997

This deepcopy() is causing memory spike when people call `Client.query_and_wait()`.

Fixes #2279 🦕

### DIFF
--- a/google/cloud/bigquery/query.py
+++ b/google/cloud/bigquery/query.py
@@ -1400,7 +1400,7 @@ class _QueryResults(object):
             api_response (Dict): Response returned from an API call
         """
         self._properties.clear()
-        self._properties.update(copy.deepcopy(api_response))
+        self._properties.update(api_response)
 
 
 def _query_param_from_api_repr(resource):


tswast (approved):
Thanks!

PR #2284, opened 2025-09-11
feat: microgen - adds config file

chalmerlowe:
Adds a configuration file to guide the microgenerator and identify templates and output locations.

- [ ] Follows: PR #2283 (should be merged after that PR is merged.)

PR #2285, opened 2025-09-11
chore: microgen - add noxfile for the microgenerator

chalmerlowe:
Adds a noxfile.py to assist with testing the microgenerator and performing tasks such as blacken and lint.

- [ ] Follows: PR #2284 (should be merged after that PR is merged.)


leahecole (commented):
Is this noxfile derived from other noxfiles we use? If so, is there anything we need to do to make sure this is set up to be kept up to date by our other automation? If not, I would just ask we open an internal tracking bug to get to that at some point

leahecole (approved):

PR #2294, opened 2025-09-15
feat: microgen - adds code generation logic

chalmerlowe:
- [x] Follows: PR #2293 (should be merged after that PR is merged.)

Adds code generation logic
   
Implements functions for generating Centralized Client code, including:

- Generating import statements.
- Rendering code from Jinja2 templates using data from the source file we previously analyzed.


scripts/microgenerator/generate.py
```diff
@@ -492,3 +492,121 @@ def analyze_source_files(
 
     return parsed_data, all_imports, all_types, request_arg_schema
 
+
+# =============================================================================
+# Section 3: Code Generation
+# =============================================================================
+
+def _generate_import_statement(
+    context: List[Dict[str, Any]], key: str, path: str
```
tswast, line 501: nit: if we want to use the python-y names for this, `package` would be more accurate than `path`.

With path, I think "file path" usually.

scripts/microgenerator/generate.py
```diff
@@ -492,3 +492,121 @@ def analyze_source_files(
...
+    Generates source code files using Jinja2 templates.
+    """
+    data, all_imports, all_types, request_arg_schema = analysis_results
+    project_root = config["project_root"]
+    config_dir = config["config_dir"]
+
+    templates_config = config.get("templates", [])
+    for item in templates_config:
+        template_path = os.path.join(config_dir, item["template"])
+        output_path = os.path.join(project_root, item["output"])
```
tswast, line 529: [nit, optional] The `pathlib.Path`'s `/` operator is a little less verbose and seems to be the preferred for new code.

scripts/microgenerator/generate.py
```diff
@@ -492,3 +492,121 @@ def analyze_source_files(
...
+
+    Returns:
+        A formatted, multi-line import statement string.
+    """
+    names = sorted(list(set([item[key] for item in context])))
+    names_str = ",\n    ".join(names)
+    return f"from {path} import (\n    {names_str}\n)"
+
+
+def generate_code(config: Dict[str, Any], analysis_results: tuple) -> None:
```
tswast, line 518: A tuple input is a bit difficult to review to determine if the order of the fields is correct. Have you considered using a frozen data class? Or if positional access is required a named tuple?

scripts/microgenerator/generate.py
```diff
@@ -492,3 +492,121 @@ def analyze_source_files(
...
+                    if key.endswith(f".{inferred_request_name}"):
+                        fq_request_name = key
+                        break
+
+                # If found, augment the method context.
+                if fq_request_name:
+                    context["request_class_full_name"] = fq_request_name
+                    context["request_id_args"] = request_arg_schema[fq_request_name]
+
+                methods_context.append(context)
```
tswast, line 566: With several nested loops and if statements, I'm having some trouble following along today. Maybe worth adding some private helper methods.

scripts/microgenerator/generate.py
```diff
@@ -492,3 +492,121 @@ def analyze_source_files(
...
+
+        template = utils.load_template(template_path)
+        methods_context = []
+        for class_name, methods in data.items():
+            for method_name, method_info in methods.items():
+                context = {
+                    "name": method_name,
+                    "class_name": class_name,
+                    "return_type": method_info["return_type"],
+                }
```
tswast, line 539: Thoughts on using a data class for this instead of a dictionary?

scripts/microgenerator/generate.py
chalmerlowe: When the code first started, we were only passing one item, which became two in a tuple and then three and is now four items.

I agree, **it is time to move it to a more robust solution**. Not all the parts that will end up being affected by this move are in this PR, so I would much prefer to merge all the outstanding PRs before doing too many changes to logic, etc.

I will defer this to the [TODO list hosted internally at b/445158219](https://b.corp.google.com/issues/445158219) for now.

scripts/microgenerator/generate.py
chalmerlowe: I will look this over and consider whether it should be modified in a future PR. Right now, for an alpha release to see what works and what doesn't, a very small dict is probably a reasonable conveyance in a microgenerator. Also added this to the TODO list for tracking.

tswast (approved):

PR #2296, opened 2025-09-16
feat: microgen - adds `testing` directory

chalmerlowe:
Adds a testing directory for the microgenerator with a new file: `constraints-3.13.txt`

Used by the `nox.session()` in the microgenerator `noxfile.py`. Added to support alpha release. Other constraints files will likely follow as the project evolves.

PR #2297, opened 2025-09-16
feat: microgen - upgrade `_to_snake_case()` function

chalmerlowe:
- [ ] Follows: PR #2296  (should be merged after that PR is merged.)

This PR upgrades the `_to_snake_case()` function to better handle names that have acronyms or initialisms.
Also adds a `raise` statement if method name is empty.


tswast (approved):

shollyman (approved):
Do you need to worry about special case acronym be

PR #2302, opened 2025-09-19
fix: honor custom `retry` in `job.result()`

### LINKED ISSUE #2210: `LoadJob.result()` ignores its default `retry`
#### Environment details

  - OS type and version: Fedora Linux 42
  - Python version: 3.13.3
  - pip version: 24.3.1
  - `google-cloud-bigquery` version: 3.34.0

#### Steps to reproduce

Here is a failing unit test, to add in "test_job_retry.py". 
The test fails with the `DEFAULT_RETRY`, and passes with a copy of the object.

```python
@pytest.mark.parametrize(
    "result_retry",
    [
        pytest.param(
            {},
            id="default retry",
        ),
        pytest.param(
            {"retry": google.cloud.bigquery.retry.DEFAULT_RETRY.with_timeout(timeout=10.0)},
            id="custom retry object",
        ),
    ],
)
def test_retry_load_job_result(result_retry, PROJECT, DS_ID):
    from google.cloud.bigquery.dataset import DatasetReference
    from google.cloud.bigquery.job.load import LoadJob

    client = make_client()
    conn = client._connection = make_connection(
        dict(
            status=dict(state="RUNNING"),
            jobReference={"jobId": "id_1"}, 
        ),
        google.api_core.exceptions.ServiceUnavailable("retry me"),
        dict(
            status=dict(state="DONE"),
            jobReference={"jobId": "id_1"},
            statistics={"load": {"outputRows": 1}},
        ),
    )

    table_ref = DatasetReference(project=PROJECT, dataset_id=DS_ID).table("new_table")
    job = LoadJob("id_1", source_uris=None, destination=table_ref, client=client)
    result = job.result(**result_retry)

    assert job.state == "DONE"
    assert result.output_rows == 1

    # We made all the calls we expected to.
    assert conn.api_request.call_count == 3
```

The culprit appears to be this line:
https://github.com/googleapis/python-bigquery/blob/eb9c2aff242c5107f968bbd8b6a9d30cecc877f6/google/cloud/bigquery/job/base.py#L1000 introduced in https://github.com/googleapis/python-bigquery/pull/41
and whose expected behavior was broken after the refactoring of Retry in https://github.com/googleapis/python-api-core/pull/462

chalmerlowe:
The `_AsyncJob.result()` method was not correctly passing the right arguments to the superclass's `result()` method when the `retry` object was expected to be the same as the `DEFAULT_RETRY` object. Passing an empty `dict` was a way of simply letting the `google.api_core` handle things. 

```
{} if retry is DEFAULT_RETRY else {"retry": retry}
```

When the `retry` object is instantiated and when the `DEFAULT_RETRY` object is instantiated, these two elements are separate objects. This means that because the `is` keyword checks for **identity** NOT **equality** that code path never evaluates to `True` and is not used as one would expect.  

This leads to subtle bugs that this PR corrects by simply passing the `retry` object and relying on the `google.api_core` to understand what a `DEFAULT_RETRY` value is and managing it appropriately.

A new test case is added to verify that `job.result()` correctly handles both the default retry use case and a custom retry object use case.

Fixes #2210 🦕


tests/unit/test_job_retry.py
```diff
@@ -615,3 +615,49 @@ def test_query_and_wait_retries_job_for_DDL_queries(global_time_lock):
...
+
+    table_ref = DatasetReference(project=PROJECT, dataset_id=DS_ID).table("new_table")
+    job = LoadJob("id_1", source_uris=None, destination=table_ref, client=client)
+    result = job.result(**result_retry)
+
+    assert job.state == "DONE"
+    assert result.output_rows == 1
+
+    # We made all the calls we expected to.
+    assert conn.api_request.call_count == 3
```
daniel-sanche, line 663: This test seems to ensure that the rpc is being retried, but not that it's using the retry configuration that was actually passed in. 

If a regression made it fall back to DEFAULT_RETRY in the handwritten layer, or even the default retry configured in the gapic layer like before, wouldn't the tests still pass? Or is that tested elsewhere?

chalmerlowe: Spent some time thinking about this PR.

Began to doubt whether checking the `client._connection.api_request()` (e.g. `conn.api_request`) method was the right way to go. 

Turns out, the `client._call_api()` method wraps the call to `client._connection.api_request()`. The `retry` object is used by `_call_api` to manage any needed retries, but the `retry` object itself is NOT passed as a direct argument to `client._connection.api_request`. Nor is it expected to be.      

Here's what we did:

   1. Mock `client._call_api`: we wrapped the original `client._call_api` using `mock.patch.object` with `wraps=client._call_api`. This lets the real `_call_api` (including the `retry` decorator) execute, but we can still inspect the calls.
   2. Inspect `retry` argument: we added a loop through `wrapped_call_api.mock_calls` to check the `retry` object passed as the first argument to `_call_api`.
   3. Conditional Assertions: Inside the loop, we added assertions to verify that the properties of the `called_retry` object match the `expected_retry` (which is either `DEFAULT_RETRY` or the "custom retry" object). we've included a specific check for the `_timeout == 10.0` in the "custom retry" case.
   4. Kept Existing Assertions: The original assertions on `job.state`, `result.output_rows`, and `conn.api_request.call_count` are still in place.

daniel-sanche (approved):

PR #2308, opened 2025-09-23
fix: remove ambiguous error codes from query retries

tswast:
Context: internal issue b/445984807 comment 10.
🦕


chalmerlowe (approved):
LGTM

PR #2309, opened 2025-09-24
build: update `pyproject.toml` to follow PEP 639

walid83-0:
Hi, I updated pyproject.toml following PEP 639.
Please review and merge if everything is fine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 [build-system]
-requires = ["setuptools >= 77.0.3"]
+requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "google-cloud-bigquery"
 authors = [{ name = "Google LLC", email = "googleapis-packages@google.com" }]
-license = { text = "Apache-2.0" }
+license = "Apache-2.0"
 license-files = ["LICENSE"]
 requires-python = ">=3.9"
 description = "Google BigQuery API client library"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,13 +13,14 @@
 # limitations under the License.
 
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools >= 77.0.3"]
 build-backend = "setuptools.build_meta"
 
 [project]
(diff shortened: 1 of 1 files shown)


chalmerlowe:
@walid83-0 

I am curious about the decision to provide a minimum version number for setuptools.  What is the rationale for this decision? And why such a recent version of setuptools (i.e. "77.0.3"), which, if memory serves, came out in the past six months.

As a general rule, Google tends to avoid putting too many limits on dependency versions as our library users are a global audience with many different environments. We want to be able to enable customers to use this library with as little friction as possible.

For projects in the Googleapis organization that have `pyproject.toml` files, I can only find three that include a version number with setuptools and those versions are pretty old `{"42", "61.0"}`

ACTION: unless we have a really good reason, I request that the version bump be stripped out entirely.

walid83-0:
Thanks for pointing this out. I mistakenly pinned setuptools to 77.0.3 simply because it was the most recent release. I understand now that this creates unnecessary friction for users. Since we don’t depend on any features from that specific version, I’ll update the requirement to be more flexible (>=61) or remove it entirely.
I’m still learning and just wanted to contribute, so thanks for the guidance.

chalmerlowe:
Let's just remove the version number completely, please. Thanks!

walid83-0:
/

pyproject.toml
```diff
@@ -19,7 +19,8 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "google-cloud-bigquery"
 authors = [{ name = "Google LLC", email = "googleapis-packages@google.com" }]
-license = { text = "Apache 2.0" }
+license = { text = "Apache-2.0" }
```
chalmerlowe, line 22: Dicts appear to be deprecated. It appears that the correct format is a simple string:
Fixing this should clear up the failing tests. <fingers crossed>.

```suggestion
license = "Apache-2.0"
```

chalmerlowe: @walid83-0 Take a look. The tests are failing. Here is an example Error:

```
× Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [44 lines of output]
      configuration error: `project.license` must be string
      GIVEN VALUE:
          {
              "text": "Apache-2.0"
          }
      
      OFFENDING RULE: 'type'
      
      DEFINITION:
          {
              "type": "string"
          }
```

daniel-sanche (approved):

chalmerlowe (approved):

PR #2331, opened 2025-11-18
fix: Handle KeyboardInterrupt in _wait_or_cancel

### LINKED ISSUE #15694: Query/job is not cancelled when interrupted by KeyboardInterrupt
Looking at current code, in `_wait_or_cancel` we cancel query if any of the exceptions happens. However, since we're intercepting `Exception`, this means that `KeyboardInterrupt` is not handled (as it inherits from `BaseException`, not `Exception`) and bubbles up (usually resulting in program termination). If query was already submitted at that moment, it will continue running, which is not desirable behavior (e.g. I realized query wasn't right and pressed Ctrl+C to terminate the program)

#### Environment details

Not relevant

#### Steps to reproduce

  1. Submit any long-running query using `python-bigquery` and wait for ~10s to make sure query was submitted
  2. Press Ctrl+C to raise `KeyboardInterrupt`
  3. Program exits, but query is still running on the BigQuery

#### Code example

```python

client = bigquery.Client()

query = """
WITH a AS (
  SELECT x AS i
  FROM UNNEST(GENERATE_ARRAY(1, 10000)) AS x
),
b AS (
  SELECT y AS j
  FROM UNNEST(GENERATE_ARRAY(1, 10000)) AS y
),
pairs AS (
  SELECT
    i,
    j,
    SIN(i * j) AS s,
    COS(i + j) AS c
  FROM a
  CROSS JOIN b
),
heavy AS (
  SELECT
    i,
    j,
    -- intentionally nasty CPU-heavy expression
    EXP(
      SIN(s) * COS(c)
      + LOG(ABS(s) + 1.000001)
      + LOG(ABS(c) + 1.000001)
    )
    + EXP(
      SIN(i) * COS(j)
      + LOG(i + 1.000001)
      + LOG(j + 1.000001)
    ) AS heavy_val
  FROM pairs
)
SELECT
  COUNT(*) AS row_count,
  SUM(heavy_val) AS total_heavy_val
FROM heavy;
"""

rows = client.query_and_wait(query)

for row in rows:
    print(f"row_count={row['row_count']}, total_heavy_val={row['total_heavy_val']}")
```

OlegWock:

Fixes googleapis/google-cloud-python#15694 🦕

### DIFF
--- a/google/cloud/bigquery/_job_helpers.py
+++ b/google/cloud/bigquery/_job_helpers.py
@@ -720,11 +720,11 @@ def _wait_or_cancel(
                 )
             )
         return query_results
-    except Exception:
+    except (KeyboardInterrupt, Exception):
         # Attempt to cancel the job since we can't return the results.
         try:
             job.cancel(retry=retry, timeout=api_timeout)
-        except Exception:
+        except (KeyboardInterrupt, Exception):
             # Don't eat the original exception if cancel fails.
             pass
         raise
--- a/tests/unit/test__job_helpers.py
+++ b/tests/unit/test__job_helpers.py
@@ -1162,3 +1162,30 @@ def test_wait_or_cancel_exception_raises_original_exception():
         timeout=123,
         retry=retry,
     )
+
+
+def test_wait_or_cancel_keyboard_interrupt_cancels_job():
+    job = mock.create_autospec(job_query.QueryJob, instance=True)
+    job.result.side_effect = KeyboardInterrupt()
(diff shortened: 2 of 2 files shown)


chalmerlowe:
**Migration Notice:** This library is moving to the [google-cloud-python](https://github.com/googleapis/google-cloud-python) monorepo soon.

We closed this PR due to inactivity to ensure a clean migration. Please re-open this work in the new monorepo once the migration is complete!

PR #2333, opened 2025-11-20
feat: support timestamp_precision in table schema

Linchin:


google/cloud/bigquery/schema.py
```diff
@@ -374,6 +387,19 @@ def policy_tags(self):
...
         return PolicyTagList.from_api_repr(resource) if resource is not None else None
 
+    @property
+    def timestamp_precision(self):
+        """Optional[int]: Precision (maximum number of total digits in base 10)
+        for seconds of TIMESTAMP type.
+
+        Possible values include:
+
+        - 6 (Default, for TIMESTAMP type with microsecond precision)
```
daniel-sanche: it seems confusing to say 6 is the default here, since this is a getter

google/cloud/bigquery/schema.py
```diff
@@ -196,6 +196,16 @@ class SchemaField(object):
...
+
+        timestamp_precision: Optional[int]
+            Precision (maximum number of total digits in base 10) for seconds
+            of TIMESTAMP type.
+
+            Possible values include:
+
+            - 6 (Default, for TIMESTAMP type with microsecond precision)
+
+            - 12 (For TIMESTAMP type with picosecond precision)
```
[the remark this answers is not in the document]
Linchin: This is an option, I did not choose it because the backend defined it to be an integer, and I think we can let the backend handle value validation. What do you think?

google/cloud/bigquery/schema.py
```diff
@@ -417,6 +435,7 @@ def _key(self):
             self.description,
             self.fields,
             policy_tags,
+            self.timestamp_precision,
```
daniel-sanche: I think this would be a problem with the current code, because None, 0, and enums.TimestampPrecision.MICROSECOND will be treated as different values, even though they represent the same state

If you store the value in a consistent way like I suggest in other comments, this won't be a problem

daniel-sanche (changes requested):

google/cloud/bigquery/enums.py
```diff
@@ -480,3 +480,18 @@ class SourceColumnMatch(str, enum.Enum):
     NAME = "NAME"
     """Matches by name. This reads the header row as column names and reorders
     columns to match the field names in the schema."""
+
+
+class TimestampPrecision(object):
```
daniel-sanche: Looking at the code again after our talk, I still think this should be turned into a true enum.

With the current code, TimestampPrecision is just a shortcut to creating int or None values. [The argument type can't be TimeStampPrecision](https://github.com/googleapis/python-bigquery/pull/2333/files#r2604634140). This means all the docstrings have to describe all the values as `Union[int | None]`, and then do more work to describe how to build one or what those special values mean, and we lose a lot of the value of encapsulating it in an enum to begin with

If we change this to actually be an enum, we will have to convert it back to it's raw value before sending it to the backend. But it should make typing a lot cleaner

daniel-sanche: But if you want to leave it as-is, that's fine. We'd just need to update the typing to be accurate

google/cloud/bigquery/schema.py
daniel-sanche: if you decide to do an Enum subclass, this could look like: self.timestamp_precision.value

google/cloud/bigquery/schema.py
```diff
@@ -263,6 +279,15 @@ def from_api_repr(cls, api_repr: dict) -> "SchemaField":
...
 
+        # The API would return a string despite we send an integer. To ensure
+        # success of resending received schema, we convert string to integer
+        # to ensure consistency.
+        if (
+            isinstance(api_repr, dict)
+            and type(api_repr.get("timestampPrecision")) is str
+        ):
+            api_repr["timestampPrecision"] = int(api_repr["timestampPrecision"])
+
```
daniel-sanche: suggestion: this might be cleaner with a try/catch:

```
try:
  # ensure timestamp is in int format
  api_repr["timestampPrecision"] = int(api_repr["timestampPrecision"])
except TypeError, KeyError:
  # ignore unset timestamps
  pass
```

daniel-sanche (dismissed):
LGTM, with some questions/comments

google/cloud/bigquery/schema.py
```diff
@@ -263,6 +279,15 @@ def from_api_repr(cls, api_repr: dict) -> "SchemaField":
         """
         placeholder = cls("this_will_be_replaced", "PLACEHOLDER")
 
+        # The API would return a string despite we send an integer. To ensure
+        # success of resending received schema, we convert string to integer
+        # to ensure consistency.
```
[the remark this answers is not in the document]
Linchin: Upon a second thought, this has been a persistent issue with other integer fields as well, hence the helper function would force a type conversion here: https://github.com/googleapis/python-bigquery/blob/bfd9d8da91d0cb1a1e5d87007fa8a5462486dca9/google/cloud/bigquery/_helpers.py#L983

daniel-sanche (dismissed):

daniel-sanche (approved):

PR #2338, opened 2025-11-24
chore(python): removes a filter put in place due to a dependency issue with pyarrow

chalmerlowe:
Due to an issue with `pyarrow`, a significant dependency for certain python-bigquery use cases, not being compatible with Python 3.14, we temporarily skipped the failing CI/CD check for 3.14 while awaiting the update to pyarrow. Pyarrow is now fully compatible, so that filter is being removed.

**KNOWN ISSUES**: this will show that unittests for 3.14 are failing. This has nothing to do with this PR/these changes. It is being addressed in an alternate mod. It is due to a missing dependency related to handling IO for `geopandas` (namely it is missing `libgdal-dev`, etc which are normally installed with `pyogrio` + `geopandas`). Because `pyogrio` is currently not compatible with Python 3.14 the tests in 3.14 cannot complete.

This should not prevent **this PR from being merged** to help solve the current issue, which is a blocker for getting our continuous tests to green.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -112,9 +112,6 @@ def default(session, install_extras=True):
     if install_extras and session.python == UNIT_TEST_PYTHON_VERSIONS[0]:
         install_target = ".[bqstorage,pandas,ipywidgets,geopandas,matplotlib,tqdm,opentelemetry,bigquery_v2]"
     elif install_extras:  # run against all other UNIT_TEST_PYTHON_VERSIONS
-        install_target = (
-            ".[bqstorage,pandas,ipywidgets,matplotlib,tqdm,opentelemetry,bigquery_v2]"
-        )
     else:
         install_target = "."
     session.install("-e", install_target, "-c", constraints_path)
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -112,7 +112,9 @@ def default(session, install_extras=True):
     if install_extras and session.python == UNIT_TEST_PYTHON_VERSIONS[0]:
         install_target = ".[bqstorage,pandas,ipywidgets,geopandas,matplotlib,tqdm,opentelemetry,bigquery_v2]"
     elif install_extras:  # run against all other UNIT_TEST_PYTHON_VERSIONS
-        install_target = ".[all]"
+        install_target = (
+            ".[bqstorage,pandas,ipywidgets,matplotlib,tqdm,opentelemetry,bigquery_v2]"
+        )
```
chalmerlowe:
```suggestion
```

gkevinzheng (approved):

